PR #182, opened 2015-01-11
Improve the user signup experience

creature:
This changeset introduces a new flow for users who sign up to Codebar. 

The first couple of steps are the same. Click 'Students' from the front page, then 'Sign in with Github': 
Front page of website
Second step

Once the user returns from Github, though, there's a new form with fewer fields. It should be friendlier, and have less chance the user gets lost. There's different copy for students and coaches on this page. 

First step of the new wizard

After clicking 'Next', the user can pick which groups they'd like to subscribe to. If they're a student, then the coach groups are hidden by default and listed second. If they're a coach, then the student groups are hidden by default and listed second. 

Second step allowing users to select lists to join
Second step, with one list joined

There's feature tests for this flow, and also for the old edit page too. That one's kept around to let people edit things after they've signed up.


despo:
thanks @creature

PR #185, opened 2015-01-21
Reminder emails for people on the waiting list

creature:
At the moment, if you're attending a Codebar meeting, you get a nice email reminder the night before. That's not the case if you're on the waiting list. People on the waiting list forget they're on it, or don't bring their laptop with them, or make other plans for Wednesday nights. This makes the waiting list less effective than it could be. Wouldn't it be nice if they got a reminder email too? 

screen shot 2015-01-21 at 01 27 48

There are tests for this feature, though I'm not happy with them; there's a *lot* of setup that (potentially) can be moved into the fabricators. I haven't attempted that for this, though. I'd also appreciate someone else going over the changes carefully; code that sends emails always makes me nervous.


despo:
:+1:

PR #187, opened 2015-01-22
CONTRIBUTING.md typo

passy:
Ironically, I don't really have a meaningful description for this change. :)


creature:
Thanks for this!

YAHSON0:
> Ironically, I don't really have a meaningful description for this change. :)

The lady across the street is some kind of major IT tech with IT tech team system 1111 and came here for investigation they found out that the person that they were helping was lying the accounts wasn't hers so they just decided to steal it

PR #199, opened 2015-02-05
Multilingual

laszpio:
Split content into translation strings


despo:
@laszpio moved it on #200

PR #209, opened 2015-02-20
Fixes sessions’s merge of date and time so it passes in all timezones.

laszpio:
Sessions' spec for merging date and time (#combine_date_and_time) on save was failing when run in other time zones.


despo:
thanks @laszpio :+1: 

holding off until Travis is done.

PR #217, opened 2015-03-04
Fix payment answer in FAQ

legoscia:
The answer to "Do I need to pay to attend Codebar?" used to repeat itself, instead of displaying the second half of the answer.


despo:
@legoscia thanks for fixing this :+1:

PR #218, opened 2015-03-14
Remove unused code

gregmolnar:
It looks like the role finder never returns anything, also `user.roles`
expects a `Permission` not a `Role` so even if I try to set the role with a request it throws an exception(otherwise it might be a security issue I guess)


despo:
@gregmolnar thanks. You are right, when I first plugged this in I needed the roles so I could do the migration and never removed it afterwards. :+1:

PR #219, opened 2015-03-18
Update ruby 2.1.5. Rails 4.1.9. Plus other gem updates.

jlsync:
because security, speed, my laptop, etc.


despo:
@jlsync thank you - good idea and much needed. :+1:

PR #230, opened 2015-04-19
Revamp

despo:
<3


knuxus:
looks lovely, but maybe the copy on the main hero needs a little more bumping up? it looks a little too small for the importance that it is.

Issue #229, opened 2015-04-16
add contacts to a sponsor


despo:
Add contact email and ability to assign members as contacts for a sponsor.

PR #236, opened 2015-05-09
Add contact details to sponsor

margOnline:
[WIP] Addresses issue #229

The view needs work and has not been completed as I would like clarification on a couple of points:
* what is meant by 'ability to assign members as contacts for a sponsor'. 
  * Can it be any member?
  * I've assumed a one-to-many association between member and sponsor
* Any thoughts on the ui for selecting a member to be the contact point?
* Any requirements for validation on the new fields?


despo:
Hey @margOnline 

Yes, any member sounds right. I don't think any validations are necessary since they should already have an email setup - and we don't really need a phone number.

A dropdown box I think, with multiple select enabled similar to here: 
https://github.com/codebar/planner/blob/master/app/views/admin/workshops/_edit_form.html.haml#L20

No validation on the new fields either as we definitely have existing sponsors whose contact details we dont necessarily have.

I hope this answers your questions. Let me know if there is anything else.

app/models/sponsor.rb
```diff
@@ -3,9 +3,10 @@ class Sponsor < ActiveRecord::Base
   has_one :address
   has_many :sponsor_sessions
   has_many :sessions, through: :sponsor_sessions
+  belongs_to :member
```
despo: This should also be a many to many association, as a sponsor could have more than one points of contact.

despo:
So, a sponsor should have a contact email, as we currently dont store that in the app which means we need to look around at times to find details.

But, a sponsor should also have many members as points of contacts. This should in the future enable us to prioritize coaches, or add them to the workshop whitelist if the event is hosted at their venue.

app/views/admin/sponsors/new.html.haml
```diff
@@ -17,6 +17,11 @@
                 = f.input :website, as: :url
             .row
               .large-3.columns
+                = f.input :contact_first_name
+              .large-3.columns
+                = f.input :contact_surname
+            .row
+              .large-3.columns
```
despo: This should also include an email address. (Our sponsors dont need to be codebar members)

deniseyu:
@margOnline happy to review this for merge, please pull latest changes from upstream and resolve conflicts. Thanks for your work :smile:

PR #239, opened 2015-05-13
Add missing 'rake' command to error message

pedromaltez:
Advice given on error was missing `rake`


despo:
@pedromaltez thanks :+1:

PR #244, opened 2015-05-31
Restrict Mailchimp API calls to production

### LINKED ISSUE #243: Do not call mailchimp API on development
>Mailchimp is not really required for development - it's called when subscribing and unsubscribing to different groups to update our mailing lists. I missed this because I was using my own account when initially developing to verify.

>https://github.com/codebar/planner/blob/5271b16cbc9b56b7eeab62d2dc9a1860e2b89ce8/lib/services/mailing_list.rb#L9 - this should probably change to reflect that, I should add an issue.

— @despo, [#242]

I took the liberty of creating the issue myself. Hope you don't mind!

[#242]: https://github.com/codebar/planner/issues/242#issuecomment-102443714

pedromaltez:
This should fix #243. Can we even test this change? I'm kind of confused as to what's going on with the tests for this. What's "xit"? It skips the tests.

Also the style the parameters are passed into `client.lists.unsubscribe` is different to the style in `client.lists.subscribe`. Checked gibbon and the readme examples are written the same way. Oddly inconsistent though, for the methods to work differently. I look at it and feel like they should both be hashes.

### DIFF
--- a/lib/services/mailing_list.rb
+++ b/lib/services/mailing_list.rb
@@ -6,7 +6,7 @@ class MailingList
   end
 
   def subscribe(email, first_name, last_name)
-    return true if Rails.env.test?
+    return true unless Rails.env.production?
     client.lists.subscribe({id: list_id,
                             email: {email: email},
                             merge_vars: {FNAME: first_name, LNAME: last_name},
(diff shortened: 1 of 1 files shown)


gcapizzi:
Thanks! Wouldn't it be better if we used the [Null Object pattern](http://refactoring.com/catalog/introduceNullObject.html) here though?

pedromaltez:
@gcapizzi Possibly. I don't see how that relates to this situation exactly but please expand. :)

gcapizzi:
I was thinking about having a `MalingList.from_env` method that returned a `NullMailingList` unless `Rails.env.production?`. This would allow us to keep the `MailingList` class unchanged (and unskip its tests :)

PR #263, opened 2015-07-15
Change return to next in InvitationManager#send_session_emails

samsworldofno:
As soon as the loop gets to a banned student, it will exit the method and allow following students (and all coaches) won't get invitations. Replace `return` with `next`.


deniseyu:
I suspected this to be the error. I'll merge it and let's see what happens this week... can always roll back if necessary.

Issue #285, opened 2015-07-21
Add UI to display contact details for sponsor to admins and organisers


deniseyu:
Contact details are now being saved on sponsor tables, so the details should be displayed to organiser and admin members somewhere.

@margOnline since you are fixing the other bug, do you want to take this one as well? May be possible to add a new column to the table on /admin/sponsors with the contact details.

margOnline:
Addressed by https://github.com/codebar/planner/pull/297

Issue #284, opened 2015-07-21
Contact details for sponsor -- existing member not properly being associated


deniseyu:
UI to add existing member to a sponsor as a point of contact has been temporarily disabled, because the association was not being saved in the DB. Needs some investigation!

Currently being worked on by @margOnline

PR #297, opened 2015-07-26
Add members as contacts for sponsors

margOnline:
Addresses issues #284 and #285

### DIFF
--- a/app/controllers/admin/sponsors_controller.rb
+++ b/app/controllers/admin/sponsors_controller.rb
@@ -9,6 +9,7 @@ class Admin::SponsorsController < Admin::ApplicationController
   def new
     @sponsor = Sponsor.new
     @sponsor.build_address
+    @sponsor.contacts.build
     authorize @sponsor
   end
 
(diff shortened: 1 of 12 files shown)


app/controllers/admin/sponsors_controller.rb
```diff
@@ -16,6 +17,7 @@ def create
     @sponsor = Sponsor.new(sponsor_params)
     authorize @sponsor
     @sponsor.build_address unless @sponsor.address.present?
+    @sponsor.contacts.build
```
deniseyu: do you not need the same `unless` condition here?

app/decorators/sponsor_decorator.rb
```diff
@@ -0,0 +1,19 @@
+class SponsorDecorator < Draper::Decorator
+  delegate_all
+
+  def contact_info
+    [ member_contact_details, contact_full_name, email ].flatten.delete_if(&:empty?).join("<br/>").html_safe
```
deniseyu: right before or after the delete_if method, please also add a `compact` - there have been issues with methods being called on nil object

app/decorators/sponsor_decorator.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def member_contact_details
+    contacts.map do |member| 
+      h.link_to(member.full_name, member)
```
deniseyu: what is `h` referencing?

margOnline: h allows you to access helper methods inside a decorator

margOnline:
I think this is good to go now

deniseyu:
:+1: thanks @margOnline !

Issue #295, opened 2015-07-25
Add a nicer 404 page


deniseyu:

knuxus:
https://better-error-pages.statuspage.io/ ?

deniseyu:
@knuxus that might be a bit of a sledgehammer in this case :smile: 

Should be possible from within the app. http://easyactiverecord.com/blog/2014/08/19/redirecting-to-custom-404-and-500-pages-in-rails/

We actually only need to rescue 500 errors, because 404s are automatically being redirected to the homepage at the moment.

alexanderkustov:
Feel free to use the https://github.com/codebar/static_pages as a template.

PR #302, opened 2015-08-01
add a custom 404

lislis:
Hey there!
@Rosa-Fox and myself did a custom 404 as mentioned in #295. We are not quite sure how to remove the redirect to the home page. Help welcome!


deniseyu:
@lislis @Rosa-Fox thanks! one small thing - in config/routes could you also add a route for 500 errors. I'm not sure either why the 404s are currently being bounced to the home path, but the 500 errors are the ones that point to the default Rails error at the moment.

PR #318, opened 2015-08-23
Custom error pages

margOnline:
Builds on pr #302 which addresses #295.

Probably not the best way to handle internal errors but its better than the rails page.


deniseyu:
:+1: thanks @margOnline !

Issue #272, opened 2015-07-20
Gimme a favicon!


deniseyu:
Like the title, the codebar app could use a favicon. Maybe a cropped version of the current logo. Any designers want to take a try?

denmch:
I'd be happy to do that :)

fredericmarx:
I just saw this on https://twitter.com/yourfirstpr and got an idea for a quick sketch. I basically did a super minimalistic  version of the logo you already have, which also looks like a "C" for codebar if you flip it 90 degrees.

Let me know what y'all think and if you like it :)

![bildschirmfoto 2015-09-06 um 17 42 27](https://cloud.githubusercontent.com/assets/954294/9705246/0d86c4f2-54bf-11e5-8f35-f7adaa9ddc69.png)

kaichanvong:
@deniseyu @despo – when are initial entries to be submitted?
How many iterations may this go through?
How many minutes, seconds...? Sorry. Harsh joke.
How many hours are we hoping students or submissions to spend?

Here's my initial workings out and thinkings:
![sketch expanding own thoughts on own favicon ideas](https://farm6.staticflickr.com/5717/21241112316_7eb0e118f9_m_d.jpg)

deniseyu:
@fredericmarx thanks for picking this up :) and sorry for the delayed response. I'm a little concerned that the level of detail on this may not render when sized down, and the white foreground on black background is a bit inconsistent with the branding throughout the rest of the site. What do you think instead of just a white background with a purple 'c' ? maybe inside a set of `< >` ?

@kaichanvong this is not a competition.

kaichanvong:
@deniseyu huh? competition? I just asked some questions. Sorry if offended.

kaichanvong:
@deniseyu – this was also found by me the account @fredericmarx mentioned. One suggestion then is to set "Assignee" as "No one assigned" at the moment? Sorry again.

fredericmarx:
@deniseyu I agree, the Logo should probably not be black :)

I'd suggest to get the shape down before thinking about colors though.

Do you know what font was used for the current logo? That might be helpful if you wanna just use a letter.

deniseyu:
@fredericmarx I'm not sure what font that is... @matyikriszta do you have any contact info for the designer?

matyikriszta:
@deniseyu that would be me :) the font is Monda, you can download it from here https://github.com/vernnobile/mondaFont

matyikriszta:
[...]

PR #323, opened 2015-09-22
adding custom favicon to planner

matyikriszta:
Had to change the bg color to pass contrast specification, but I think it still looks purdy
Solves #272


despo:
Already implemented. Closing.

Issue #317, opened 2015-08-15
Coach expertise tags


deniseyu:
Feature request:

Coaches should be able to add tags to their own profiles for technologies that they are comfortable teaching. This should be editable from the /profiles route.

The tags should be a many-to-many relationship, so that students and other coaches can click on the "Rails" tag on Denise's profile and see a list of all coaches that have listed "Rails" under their areas of expertise.

A migration should also be written to pre-seed the tags table with common things, like HTML, CSS, JavaScript, jQuery, etc. It would be great if the UX for adding tags included some type of auto-complete/prompt feature that scans the database for existing tags, to try to encourage re-use. Similar to tumblr tags.

margOnline:
@deniseyu I can have a look at this.

deniseyu:
@margOnline cool - happy to pair on it with you as well :)

margOnline:
@deniseyu great! [here's](https://github.com/margOnline/planner/tree/coach_tags) what I've done so far. I've used the ActsAsTaggableOn gem. 

Haven't looked at it in a couple of weeks, any questions slack me.

PR #324, opened 2015-09-25
Coach expertise tags

margOnline:
addresses #317


app/controllers/skills_controller.rb
```diff
@@ -0,0 +1,6 @@
+class SkillsController < ApplicationController
+  def show
+    @skill = params[:id]
```
despo: Shouldn't this be set to the relevant skill?

despo:
@margOnline looks good, hadn't realised the skill param is the name.

despo:
:ship: :+1:

PR #336, opened 2015-11-12
View multiple upcoming events from dashboard

deniseyu:


app/models/event.rb
```diff
@@ -29,6 +31,10 @@ def to_s
     self.name
   end
 
+  def next_upcoming(n)
+    Event.order('date_and_time').where('date_and_time > ?', Date.today).take(3)
```
boddhisattva: @deniseyu I'd like to discuss a couple of points(mentioned below) wrt this PR:

* Do we need both a scope and an instance method here ? I see that the controller is only using the scope, hence the question.

* Also, in case we need both, what do you think about using a class method instead of an instance method over here, since we're getting the next upcoming n events and the method isn't specific to a particular event instance . 

* Lastly, I see that you're trying to pass `n` events through the function prototype, but the method is instead having a hard coded value `3` in order to fetch 3 events at the most, this wouldn't work as expected if this method is called with `n` events in the future .  

Please feel free to correct me if I'm missing something here. Also, do let me know if you'd want me to send a PR for the same. Thanks.

app/controllers/dashboard_controller.rb
```diff
@@ -64,7 +64,7 @@ def upcoming_events_for_user
   def all_events(workshops)
     course = Course.next
     meeting = Meeting.next
-    event = Event.next
+    event = Event.upcoming(3)
```
boddhisattva: @deniseyu What do you think about using a constant here ? Something like `NUMBER_OF_EVENTS = 3`.

app/models/event.rb
deniseyu: omg... this is why i should not code after going to the pub. thank you

deniseyu: you're completely right, I was attempting to write a class method here (failed) and failed to remove this after writing the scope

boddhisattva: No worries :+1:

Issue #322, opened 2015-09-22
Investigate memory leaks


deniseyu:
The codebar app has some memory leak problems. Right now they are not bad enough to make the app fall over but it is important that we understand what methods are causing them. If you have experience debugging and resolving memory leaks in Ruby/Rails apps, please get in touch.

despo:
Relevant article
https://blog.codeship.com/debugging-a-memory-leak-on-heroku/

asuffield:
Let's start by upgrading redcarpet to 3.3.3, because of https://github.com/vmg/redcarpet/pull/516 (leaks ~300 bytes on every markdown render).

If it's still leaking after that, I'll take a closer look.

PR #344, opened 2015-11-22
Update redcarpet to 3.3.3

asuffield:
Picking up https://github.com/vmg/redcarpet/pull/516 (leaks ~300 bytes
on every markdown render), which might help with #322.

(I believe this pushing this change is sufficient for heroku to
update, but I'm not intimately familiar with heroku ruby deployments.)


despo:
@asuffield Thanks :+1:

PR #349, opened 2015-11-25
Configure tests to format coverage results locally

asuffield:
This should keep coveralls working, but also emit the output in
coverage/index.html so that you can see the effect of your change when
writing it.

Also use the rails profile so results are neatly grouped.


despo:
@asuffield  :+1:

PR #350, opened 2015-11-25
Spec auth infrastructure

asuffield:
Currently, none of the tests really cover signed-in behaviour. This gives us enough test mocking for specs to click on the "Sign in" button and have it work as expected (the github auth is mocked out).


asuffield:
Ah, found some partial existing support buried in specs. Will be back with this all merged together.

PR #353, opened 2015-11-27
Random spot allocation for workshops

### LINKED ISSUE #254: Random spot allocation for workshops
When coach and student invitations go out for each workshop, instead of RSVPing directly to the event, they should be invited to "Request a spot". They can also be invited to indicate whether they would like to join the waitlist in the event they are not allocated a spot.

At 9PM on that evening, invitations will be allocated. If (Number of People Requesting Slots) > Capacity, then slots will be randomly assigned. Otherwise, everyone will receive a slot.

asuffield:
Random spot allocation, first cut. An allocation time can be set in the admin page. Before this time, all invitation responses go onto the waiting list. When the allocation job runs, it picks people randomly from the waiting list. After this, spot allocation works in the traditional way (preferring earlier signups).

This isn't as tested as I'd like it to be yet - the code is all at least superficially covered by specs, but I haven't written a full integration test for it or set up enough of a working dev system to test this properly. However, I'm fairly confident that it doesn't break any existing functionality if the feature isn't turned on, so it should be safe to merge, even if it needs more changes after it.

This pull request closes #254.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2013 Despo Pentara
+Copyright (c) 2013 Despo Pentara and Google Inc.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ ruby '2.2.2'
 
 gem 'rails', '4.1.12'
 
+gem 'activejob_backport'
 gem 'acts-as-taggable-on', '~> 3.4'
 gem 'haml'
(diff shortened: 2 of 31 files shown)


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2013 Despo Pentara
```
despo: Google has no copyrights to our app. Can this be removed please.

LICENSE
```diff
@@ -1,4 +1,4 @@
...
+Copyright (c) 2013 Despo Pentara and Google Inc.
```
despo: Note above referring to this addition.

LICENSE
asuffield: Any code that I've contributed here is owned by my employer, which requires attribution as part of our open-source contributions policy. I have to list the company somewhere.

despo: Ok. I need something to think and discuss this with the organisers so we can figure out what to do. We (codebar) should also change our license so maybe we can figure out a way that is more suitable and enables attribution.

In the meantime, any tips/suggestions are welcome.

despo:
@gcapizzi :v:

PR #354, opened 2015-11-27
Fix #347: Fix spelling in en.yml

### LINKED ISSUE #347: Spelling error: "You must be logged in to acess this page."
Can be seen on any page that requires logging in, when not logged in. http://codebar.io/dashboard is a good one.

zamakkat:
Ref Issue #347


zamakkat:
Actually I think I will create a new PR once 24PR starts (1 Dec) :smile:

PR #361, opened 2015-12-05
Open emails in browser while in development

Arkham:
This makes developing emails easier :)

@codebar


deniseyu:
@Arkham could you try creating a Sessions (in Rails console, `Member.find(your_id).add_role(:admin)`create new workshop) and sending out invitations - my belief is that this was previously commented out because it could potentially open up hundreds of new browser windows when an invite is sent to each member of the chapter.

despo:
Thanks @Arkham but what @deniseyu said. Closing.

PR #362, opened 2015-12-05
Validation for coach and student number added

emilyzheng828:
The previous event page accepts 0 and negative coach and student number.


deniseyu:
:+1: thanks @emilyzheng828

PR #366, opened 2015-12-05
Resolved https errors for google fonts and gravatar images.

### LINKED ISSUE #356: Mixed Content Warnings when loading codebar.io over HTTPS
https://codebar.io attempts to load some resources over http, which results in mixed content warnings. Some of those cause content to be blocked. while others are mere warnings.

Blocked content:
* Google fonts

Warning content:
* gravatar images

To fix this:
* Find a URL that works for google fonts, and use that instead of the existing hard-coded http link in app/views/layouts/application.html.haml 
* Find a URL that works for gravatar images and use that image instead of the existing hard-coded http link in https://github.com/codebar/planner/blob/0d9f0478bc64dce54dfcd9b4d17f11ed02aa8486/app/models/member.rb#L68
* Audit the codebase for any other instances of hard-coded http links.
* If needed, update the tests so they expect new URLs
Hopefully there are [protocol-relative URLs](http://www.paulirish.com/2010/the-protocol-relative-url/) for each of those URLs. If not then we should prefer the HTTPS style rather than having to do any logic around working out what url to show.

---

To reproduce:

Open https://codebar.io/ with the browser console open. Note a number of error/warnings that start with "Mixed Content"

Expected output: None of these errors are present.

ma3east:
Fixes #356 
Created during @codebar and @ladieswhocode event.

Had a look at resolving assets.codebar.io.... errors as well, but not able to resolve this, as the link will not work over https.

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -65,7 +65,7 @@ class Member < ActiveRecord::Base
   end
 
   def avatar size=100
-    "http://gravatar.com/avatar/#{md5_email}?s=#{size}"
+    "https://secure.gravatar.com/avatar/#{md5_email}?s=#{size}"
   end
 
   def attended_sessions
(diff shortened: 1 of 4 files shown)


despo:
@ma3east thank you!

Issue #370, opened 2016-01-06
Minor issue with workshop waitlist reminder email 


harshisuresh:
I received a "WORKSHOP WAITING LIST REMINDER" email for worshop scheduled to happen on the same day. But the content of the email mentioned  <b>tomorrow</b> instead of <b>today</b>

"This is a quick email to remind you that you're on the waiting list for <b>tomorrow's</b> workshop.

If an attendee drops out, the next person on the waiting list will automatically get their place. We'll email you if this happens. Attendees often drop out at short notice, so you should keep your laptop with you and check your email <b>tomorrow</b> afternoon."

I guess this could be improved

despo:
Closed as fixed in PR mentioned above.

PR #373, opened 2016-01-07
Fix waitlist reminder email

harshisuresh:
Fix for issue

https://github.com/codebar/planner/issues/370

Fixed email content to contain date of workshop instead of hardcoded string tomorrow as email is being sent between 4 and 30 hours... fyi : @RabeaGleissner


despo:
@harshisuresh thank you! :+1:

Issue #371, opened 2016-01-06
Delete a workshop


KimberleyCook:
Admin needs to be able to delete a workshop in the interface. This can currently only be done in the console.

boddhisattva:
@KimberleyCook I would like to give this a shot. Please suggest how could I assign this issue to myself on Github issues and on the waffle board. Thank you.

KimberleyCook:
@boddhisattva We don't really assign as such but if you want a shot at it people will read this thread and know someone is working on it so please go ahead. Thank you very much :D

boddhisattva:
Thanks @KimberleyCook :)

despo:
This should be constrained to workshops where no invitations have been emailed out and that have been created within X time. Maybe a couple of hours.

boddhisattva:
@despo Thanks for letting me know about this. Just wanted to better understand, how would it help to check if a workshop is created within X hours from the current time before deciding to delete it ?

despo:
@boddhisattva we need this functionality to be plugged in in case someone accidentally creates multiple workshops and constraining it would ensure that we dont remove information that has been on our website for a while. We dont promote deleting workshops. If a workshop is cancelled it should be updated so that people can find that information out.

Does this make sense?

boddhisattva:
@despo This definitely makes it more clear, thank you for the explanation :).

boddhisattva:
@despo Based on your point [here](https://github.com/codebar/planner/issues/371#issuecomment-169356782) . I've assumed the value of 'X' to be 4 hours. Please let me know if this is okay or request you to suggest an alternate value of 'X'. Thanks.

despo:
https://github.com/codebar/planner/pull/374

PR #374, opened 2016-01-10
Adds the ability to delete a workshop

boddhisattva:
Changes as part of this PR address [this issue](https://github.com/codebar/planner/issues/371).

Based on the discussion [here](https://github.com/codebar/planner/issues/371#issuecomment-169356782) I've assumed that a workshop is eligible for deletion if the current time is within 4 hours from the time a workshop is created.


app/controllers/admin/workshops_controller.rb
```diff
@@ -77,6 +79,19 @@ def invite
...
   end
 
+  def destroy
+    if workshop_has_no_invitees? && workshop_created_within_specific_time_frame?
+      @workshop.destroy
+
+      redirect_to admin_root_path, notice: "Workshop deleted succesfully"
+    else
+      redirect_to admin_workshop_path(@workshop),
+                  notice: "Workshop cannot be deleted. This is because it has invitees" \
```
despo: Can we extract the text to the locale and use i18n to retrieve it.

despo:
@boddhisattva looks great. It would be great if we had the text extracted to the locale file as well.
The configuration looks good!

Happy to merge as is as well as we could always pick up the small changes afterwards. Just let me know!

boddhisattva:
@despo Good point :+1: . I have extracted it out into the locale file. Thanks for your feedback on the PR :)

despo:
@boddhisattva thank you :+1:

PR #377, opened 2016-01-13
Update LICENSE

Francososa:


despo:
@Francososa thank you so much but I'm afraid we had a PR tackling this issue just before you opened yours.

PR #391, opened 2016-02-06
Working on issue 390 regarding invites to.  Threee instances of invit…

jemagee:
Found three instances of invites to that were changed to invitations to as it seemed appropriate.  Altered the text of the initial example to We *will* send you *invitations*.  Double checked all tests as passing just to be safe.
(This is my first official submission, apologies if I missed something)


despo:
@jemagee hi John. Thanks for your PR. 
Could you please remove the database configuration and bin/rspec files as they are not needed (and I believe the change in the configuration is breaking the built)

jemagee:
Removed them local and did a new push - the rspec binstubs is just my default - didn’t know it would cause a problem

despo:
@jemagee  I meant revert the database configuration [to the original](https://github.com/codebar/planner/blob/master/config/database.yml) as we still need it.

jemagee:
Sigh - sorry for that misunderstanding i’m hoping i fixed it now - reverted it to the original

despo:
@jemagee no problem! Thanks for your PR :+1:

jemagee:
Glad I could help - try to be a bit more efficient next time

Issue #395, opened 2016-02-12
"Updating Profile" mangles skills list


thecodeite:
If you go to update your skills list at http://codebar.io/member/edit then when it loads it loads your skills as a space separated list rather than a comma separated list. If you don't notice and click Save then all your skills are combined into one mega skill.

deniseyu:
what, you haven't heard of the new Javascript framework called rubyjavascriptrailsnodecsshtml?

margOnline:
I'll have a look at this.

PR #396, opened 2016-02-16
Separate skills with a comma in member update form

margOnline:
Addresses #395


despo:
@margOnline thank you :+1:

PR #398, opened 2016-02-19
Making css consistent (.scss files all across the board)

matyikriszta:
converting sass file to scss and cleaning up stylesheets folder: moving event.scss and star-rating.scss to partials folder


despo:
@matyikriszta :+1:  thank you, that's great

PR #402, opened 2016-02-21
Move pronoun field below first/last name field

### LINKED ISSUE #400: Move preferred pronoun further down on profile page
Currently, people are getting confused by the preferred gender box field during the new member setup process and on the edit profile view. It is the topmost field on the page, in the place where you might normally expect a field for title or nickname. As result, people are mistakenly putting "Miss", "Mrs", "Mr", etc. instead of "she/her", "they/them", etc.

We should move this box down the page a bit. If it was after first and last name and contained a placeholder such as "e.g. she/her", the intended purpose would become clearer.

Charlotteis:
Fixes #400 

This PR moves the "Preferred Pronouns" field on the "New Member" form further down the page, underneath the First/Last name fields. We're doing this because people can get confused and think "Mr/Mrs/Ms/Mx" goes in there (as often this is the first field you will see on any sign up form).

codebar_preferred_pronouns_fixed

Hope I did this right! Really easy to follow 'Getting Started' guide in the README of this repo.

:sun_with_face:

### DIFF
--- a/app/views/members/_form.html.haml
+++ b/app/views/members/_form.html.haml
@@ -1,12 +1,12 @@
 = simple_form_for @member, method: :put do |f|
   .row
-    .large-4.column
-      = f.input :preferred_pronoun
-  .row
     .large-6.columns
       = f.input :name
     .large-6.columns
       = f.input :surname
+  .row
+    .large-4.column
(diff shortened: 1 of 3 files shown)


deniseyu:
:+1: thanks @Charlotteis !

PR #405, opened 2016-02-28
change “preferred_pronoun” to “pronouns”

### LINKED ISSUE #403: Changed 'preferred pronoun' field to 'What pronoun do you use?'
http://mashable.com/2015/10/18/transgender-ally-words/#mrp.FBlM0kq3
It seems that "preferred pronoun" is a harmful thing to say, and we should instead use "What pronoun do you use?".

I'll see if I can get to this tomorrow.

Charlotteis:
Fixes #403

This PR changes "preferred_pronoun" to "pronouns" in the Database.

Also changes the question “Preferred pronoun?” to “What pronouns do you use?” in the member sign up form. And “Preferred pronoun” on a profile page to “Pronouns”.

403_pronouns_on_signup

Based on: http://mashable.com/2015/10/18/transgender-ally-words/. "Preferred pronouns" could suggest that they are not someones "actual" pronouns, where in fact they are of course. So instead we can ask "What pronouns do you use?" :+1: 

---

This is the most Rails work I've ever done, so I may not have gotten it right! Help greatly appreciated.

:sun_with_face:

### DIFF
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -62,7 +62,7 @@ class MembersController < ApplicationController
 
   def member_params
     params.require(:member).permit(
-      :preferred_pronoun, :name, :surname, :email, :mobile, :twitter, :about_you, :skill_list
+      :pronouns, :name, :surname, :email, :mobile, :twitter, :about_you, :skill_list
     )
   end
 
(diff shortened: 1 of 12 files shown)


jkbits1:
I like your revised phrase on the UI

deniseyu:
Thanks!

PR #465, opened 2016-08-20
Rescue from Standard Error instead of rescuing from Exception

boddhisattva:
For details as to why rescuing from Standard Error is preferable when compared to rescuing from Exception one can read -
1. http://stackoverflow.com/a/10048406/272398
2. https://robots.thoughtbot.com/rescue-standarderror-not-exception

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -19,7 +19,7 @@ if Rails.env.development?
     end
 
     puts "..done!"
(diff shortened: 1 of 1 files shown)


deniseyu:
Yep - sorry, because of our CI checker and master being out of date, I reopened and merged this in another PR. thanks @boddhisattva

PR #475, opened 2016-10-05
added length validation on members about_you field and added spec

razasayed:


KimberleyCook:
thank you so much @razasayed

Issue #467, opened 2016-09-08
Feedback to show chapter


KimberleyCook:
In the front-end for the admin feedback view it would be great to know what chapter the feedback was from.

mptap:
I'd like to take a stab at this.

KimberleyCook:
@mptap amazing yes please do :)

KimberleyCook:
@mptap how are you getting on with this?

tgturner:
@KimberleyCook If no one is working on this, I'd love to go for it. I took a look at the code and was wondering a couple of things. 

1. Where would be the best place in the view to put the chapter?
2. What would be the best way to access the chapter the feedback came from? I could get it from `feedback.tutorial.workshop.chapter` but that seems a little round about. 

Any suggestions?

KimberleyCook:
@tgturner yes please by all means :)

1) Top right I think is a good place
2) That would be fine for now.

Thank you for taking a look

PR #486, opened 2016-11-02
addition: add chapter to admin:feedback display

tgturner:
Each piece of feedback now also renders the chapter that feedback came from. To do this, I added a bit of code to the admin:feedback index template. This is an attempt to address issue #467.

### DIFF
--- a/app/views/admin/feedback/index.html.haml
+++ b/app/views/admin/feedback/index.html.haml
@@ -17,6 +17,8 @@
             .large-2.columns
               - (0...feedback.rating).each do |rating|
                 %i.fa.fa-star
+            .large-1.columns
+              = feedback.tutorial.workshop.chapter.name
           - if feedback.request.present?
             .row
               .large-12.columns


deniseyu:
Hey @tgturner thanks for this PR.

This is going to crash the page if a tutorial isn't associated with a workshop id, and in the production DB, this scenario occurs very often. A better approach would be to wrap this call in a method inside the Feedback class, and either 

a) create a new foreign key  to link a feedback to a chapter, and write a migration that backfills this field wherever possible

or

b) code very defensively using Rails' `try` method

Let me know if you'd like to take a try at either.

deniseyu:
Hi @tgturner sorry but the project maintainer team had some miscommunication about this.

`feedback.tutorial.workshop` does not actually correspond to the workshop for which a piece of feedback is given... my best guess is that it's a legacy design decision that was never implemented.

Sorry about this and for the time you spent on it. I'm going to open a new issue for this as a new feature request and close this PR to avoid confusion.

PR #508, opened 2016-12-11
Update ruby-version and documentation to Ruby 2.3.3

hanmd82:
This follows up from #504 where Ruby was upgraded to 2.3.3 in Gemfile and Travis CI


deniseyu:
Thanks @hanmd82 - good catch.

PR #511, opened 2016-12-29
Add workshop association for feedback

boddhisattva:
**Please note**: Currently I haven't added any index wrt `workshop_id` column in the Feedback table because I don't think we are currently having any page where we display workshop specific feedback.

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -35,7 +35,7 @@ class FeedbackController < ApplicationController
   private
 
   def feedback_params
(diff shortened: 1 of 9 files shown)


deniseyu:
hey @boddhisattva - I _think_ the issue may have been a stateful Travis worker, or some weird migration vs load:schema behaviour introduced with the Ruby 2.3 upgrade. I've changed it to use rake db:migrate in the travis task instead, so I'm going to close this PR and merge your commit from a different branch. Thanks!

PR #512, opened 2017-01-04
Boddhisattva add workshop association for feedback

deniseyu:


boddhisattva:
@deniseyu Thanks for following up on this based on the discussion [here](https://github.com/codebar/planner/pull/511#issuecomment-270482317) .

Issue #442, opened 2016-05-16
Add universal redirect for people who have not finished the sign-up process


deniseyu:
Occasionally, someone starts the sign-up process and then navigates to another page or closes their window. The impact is that they never receive a welcome email which contains important information including the code of conduct.

People who have not finished the sign-up process will have a model attribute called `received_student_welcome_email` or `received_coach_welcome_email` set to false. If one of these is false, that's OK - it means they subscribed as either a student or a coach but not both. If both of these are false, any time they try to access a private page (any page after the login screen) they should be redirected to Step 1 of the new member sign-up pages -- this page is under the Member controller. A notification should also flash with the text "We noticed that you did not finish signing up. Please complete your registration."

If a user has an empty string for their name and their surname, it is another indication that they did not finish signing up. In this case, the same logic as above should kick in.

Sivan1525:
@deniseyu Hi, Can I take a look on this and implement this. :)

Indrajitnaiya09:
@notapatch is this issue still open? I'm planning to pick then.

notapatch:
I do comment on a lot of PR's don't I @Indrajitnaiya09 ?

Only the maintainers can tell you if it's still open and something they want (@despo, @matyikriszta, @KimberleyCook).

There's a good chance the bug is still there. I've not checked in the last year but it was quite easy to reproduce.  It's interesting that others have tried solutions that haven't been merged but I don't know why.

Indrajitnaiya09:
@despo @matyikriszta @KimberleyCook Could you help me with this!! I'd like to solve this issue if it is not fixed yet.

PR #515, opened 2017-01-07
Add universal redirect for people who have not finished the sign-up process

j-sm-n:
New feature for Issue #442.

As it stands, after a new member authenticates with GitHub, the user will be redirected to the Step 1 route if they never hit "next."

In addition for Step 2, it felt like a less jarring UX decision to redirect to the Step 2 route if a member doesn't subscribe to an email list. If a member hits "Done" without selecting at least one email list to subscribe to, the page will reload and display the error message `"Please subscribe to at least one group before continuing."`. 

Let me know what you think! :)


deniseyu:
Hi @j-sm-n - thanks for your work on this! I'm going to do a bit of testing on this and do my best to get back to you within the next two weeks. No fault of your code, the app just has loads of weird behaviours that aren't obvious. Really appreciate your time and your thoughtfulness towards improving the user experience.

PR #527, opened 2017-02-26
Make signing up to things more efficient

matyikriszta:
Clicking the sign-in button on a workshop, event or meeting page
shall now redirect you back to that page instead of your dashboard.

Because the flow of:

* Visit meeting page
* Press "sign in" and get redirected to the page you were just on
* Press "attend"

is loads better than:

* Visit meeting page
* Press "sign in" and get redirected to your dashboard
* Go to the events page
* Go to the single event page
* Press "attend"

This PR bought to you by @matyikriszta & @BPScott - two not-rails developers - so if anybody can review about ruby/rails idonims that'd be great.

### DIFF
--- a/app/controllers/auth_services_controller.rb
+++ b/app/controllers/auth_services_controller.rb
@@ -16,7 +16,10 @@ class AuthServicesController < ApplicationController
         session[:oauth_token]        = omnihash[:credentials][:token]
         session[:oauth_token_secret] = omnihash[:credentials][:secret]
 
-        finish_registration or redirect_to dashboard_path
+        prior_url = request.env['omniauth.origin']
+        redirect_url = is_viable_redirect_location(prior_url) ? prior_url : dashboard_path
+
+        finish_registration or redirect_to(redirect_url)
       else
         member = Member.find_by_email(omnihash[:info][:email])
         member = Member.new(email: (omnihash[:info][:email])) if member.nil?
@@ -69,5 +72,26 @@ class AuthServicesController < ApplicationController
 
   def redirect_path
     :services
+  end
+
+  def is_viable_redirect_location(url)
+    # When you visit an attendable page (e.g. a Workshop, Eventor  Meeting) you
+    # probably want to sign-up to that thing, which requires signing in. When
+    # you sign in you want to get sent back to the page you were on previously
+    # so you can click the attend button right away rather than being sent to
+    # your dashboard then having to navigate back to the Workshop/Event/Meeting
+    # page.
+    begin
+      route = Planner::Application.routes.recognize_path(url)
+
+      return true if route[:controller] == "workshops" && route[:action] == "show"
+      return true if route[:controller] == "meetings" && route[:action] == "show"
+      return true if route[:controller] == "events" && route[:action] == "show"
+    rescue ActionController::RoutingError
+      # recognize_path might throw this if you give it a url the app does
+      # not know how to turn into a route
+    end
+
+    return false
   end
 end


app/controllers/auth_services_controller.rb
```diff
@@ -70,4 +73,25 @@ def omniauth_providers
   def redirect_path
     :services
   end
+
+  def is_viable_redirect_location(url)
```
nickcharlton: A more Rails way of naming this would be to do something like `viable_redirect_location?(url)`.

app/controllers/auth_services_controller.rb
```diff
@@ -70,4 +73,25 @@ def omniauth_providers
...
+    # When you visit an attendable page (e.g. a Workshop, Eventor  Meeting) you
+    # probably want to sign-up to that thing, which requires signing in. When
+    # you sign in you want to get sent back to the page you were on previously
+    # so you can click the attend button right away rather than being sent to
+    # your dashboard then having to navigate back to the Workshop/Event/Meeting
+    # page.
+    begin
+      route = Planner::Application.routes.recognize_path(url)
```
nickcharlton: It looks like this method was deprecated in [Rails v4.1.8](http://apidock.com/rails/ActionDispatch/Routing/RouteSet/recognize_path), so I'd try and avoid using it. It's also exceptional which feels odd in Ruby.

app/controllers/auth_services_controller.rb
```diff
@@ -70,4 +73,25 @@ def omniauth_providers
...
+
+      return true if route[:controller] == "workshops" && route[:action] == "show"
+      return true if route[:controller] == "meetings" && route[:action] == "show"
+      return true if route[:controller] == "events" && route[:action] == "show"
```
nickcharlton: What do you think about avoiding the duplication by doing something like?

```ruby
%w(workshops meetings events).include?(route[:controller])
```

It'll return `true` if it's contained inside that array of strings, and in Ruby we can avoid writing explicit `return`s, too.

nickcharlton:
I thought about this further from when we paired on this at Codebar. What do you think about flipping this logic on it's head, and instead of detecting the URL when you're trying to redirect, specifying it as an argument to the login?

Something like: `<a href="/login?redirect=/workshops/1234">Login</a>` and everywhere else not doing that?

It'd avoid you having to parse the URL back and then you'd only need to provide that as an argument in the few cases where you do want the redirect.

app/controllers/auth_services_controller.rb
BPScott: That feels a bit scary as we're mapping to a "controller + action" pairing. It's a happy coincidence atm that all the actions are called "show" but if that were to change later then that'd make the logic for this more complicated.

The duplication is acceptable as it makes this easier to parse imo.

app/controllers/auth_services_controller.rb
BPScott: Do you know what API we should instead of this?

BPScott:
Doing this with a query parameter and passing this through was my first thought. However I don't think it will work as when you define github as an oauth provider you have to specify a fixed authorization callback url, you cant have that change based on a query string you pass in on a per-authorisation request basis. Thus there is no way to have the value of that `?redirect` parameter be passed back to the planner app after authentication.

Having this be done against a whitelist of routes also helps avoid the [unvalidated redirects](https://www.owasp.org/index.php/Unvalidated_Redirects_and_Forwards_Cheat_Sheet) exploit.

nickcharlton:
>  I don't think it will work as when you define github as an oauth provider you have to specify a fixed authorization callback url

BNewing:
Is the same logic going to be applied to when you sign up for an account for the first time? If so that'd be awesome! If not, is it something that's easy to add in?

Oxford is a relatively new chapter but we get new attendees each time and some of them mistake signing up for the chapter as signing up for a specific workshop when sent the link to the workshop

app/controllers/auth_services_controller.rb
KimberleyCook: @nickcharlton Hey, you mentioned this got deprecated. Do you have another solution as I really want to get this merged in?

nickcharlton:
Hi @KimberleyCook!

Sorry, I'd basically forgotten about this. :( I'd asked about a little while ago at thoughtbot to see if anyone had any other thoughts on on a nice way to implement this.

A suggestion I quite liked went along the lines of this:

1. You store the return URL in the session: `session[:last_url] = params`.
2. Then on login you can check to see if there's a value (`redirect_to url_for(session[:last_url])`).
3. If not, you can go to the dashboard.

On the security risk point: the code will only be executed when you're authenticated, and Rails is pretty sensible about redirecting with parameters to known routes. If it's really a concern, you could validate that the last URL is acceptable by doing something like: `session[:last_url] = request.path`.

What do you think?

matyikriszta:
@despo This PR is very out of date so going to close it, but we should come up with a working solution for sensible redirects

Issue #510, opened 2016-12-14
Validate Group names


deniseyu:
For every chapter, our convention is to have two different groups, one named `Students` and one named `Coaches`. Membership in one of these groups is represented by a user subscribing to one of them.

We only have used two strings for names, `Students` and `Coaches` in the last three years, so it is fairly safe to say that these are the only two group types that should exist in every chapter. This should be documented and enforced via a model validation.

In `app/models/group.rb`, we need a custom validator that verifies if the name is either 'Students' or 'Coaches'.

In `app/views/admin/group/new.html.erb` the user interface should be updated to use some sort of preset input (such as a dropdown box or radio buttons) so Students or Coaches are the only options available.

Add a quick unit test for the validation in the Group spec file as well.

ajay2507:
Hi deniseyu, Can I work on this enhancement?

deniseyu:
Sure @ajay2507, go for it.

ajay2507:
okay.. thank you @deniseyu

margOnline:
oops, I didnt see anyone assigned to this and have only just noticed that someone opted to do it. Sorry @ajay2507 maybe you could review the pr?

ajay2507:
@margOnline  I am looking at another issue.. I didn't start this enhancement.. It would take some time.. Is it okay?

margOnline:
This is addressed by #530 which is ready for review. I dont have admin access to view the changes to the form so would be good if someone could check the ui.

matyikriszta:
Addressed by #530, closing this issue

PR #530, opened 2017-02-27
Validate group names

margOnline:
This pr addresses issue #510

### DIFF
--- a/app/models/group.rb
+++ b/app/models/group.rb
@@ -1,4 +1,7 @@
 class Group < ActiveRecord::Base
+
+  NAMES = %w(Coaches Students)
+
   belongs_to :chapter
   has_many :subscriptions
   has_many :members, through: :subscriptions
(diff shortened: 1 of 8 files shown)


app/models/group.rb
```diff
@@ -9,7 +12,9 @@ class Group < ActiveRecord::Base
   scope :students, -> { where(name: 'Students') }
   scope :coaches, -> { where(name: 'Coaches') }
 
-  validates :name, :chapter_id, presence: true
+  validates :name, presence: true
+
+  validate :valid_name?
```
marten: You could save yourself a bit of code here. Rails has a built-in validator that does exactly this, it works like this:

```ruby
validates :name, presence: true, 
                 inclusion: {in: NAMES, message: "Invalid name for Group"}
```

One minor difference is that the above does add the error to the `name` field rather than the `group` field like your `valid_name?` does. But that might actually be what you wanted in the first place, since that's what the HTML form is using too (and `.group` doesn't actually seem to be a method here anyways?)

margOnline: Of course, good point. Shows how rusty my rails knowledge is, I had completely forgotten abt inclusion. Thx for reviewing!

matyikriszta:
thanks @margOnline for the PR and @marten for reviewing, I'm happy to merge it in

PR #532, opened 2017-03-06
Copy amends

KimberleyCook:
Lots of copy amends for the site where at points I think it can a little confusing.

For example using the word mentor and coach, should now always be coach. Sponsor a workshop should not be host a workshop. And clearer button text in places.


matyikriszta:
hey @KimberleyCook, this looks good, could you rebase your branch so we can merge?

matyikriszta:
thanks @KimberleyCook 👍

Issue #536, opened 2017-03-21
Events page displays every event


dondonz:
The Events page currently displays every Codebar event held in the past. I think it's good to list past events, as it's a great indicator of the quality & regularity of Codebar's workshops.

As the number of workshops is growing at an ever faster rate, I'd like to limit the number of workshops shown on the Events page and reduce the size of the DB request.

What do you think about only loading the most recent 20/30 events and then end with a bit of text saying "and 189 other workshops since 2013"? I'm open to suggestions - if you'd prefer to still see all the events, we could have pagination instead.

I'm happy to look after this ticket.

matyikriszta:
That sounds good @dondonz, I'm happy to display the last 30 events only and then a number at the end that says however many workshops in total (for workshops and maybe Monthlies?)

BPScott:
There is a headache here where events are stored in across multiple tables, so it becomes a bit fiddly to request "the last 30 of any event type". You'd probably need to request the last 30 of every type of event from the DB, munge them into one array, order that array, then throw away the everything that wasn't in the top 30.

Better long-term plan would be creating an "event" table that can store the multiple types of event, buts that's very drastic and requires a heck of a lot of DB reworking. Your plan is a really solid stopgap to speed up the the events page which is now getting pretty unmanageable.

dondonz:
Thanks @BPScott - I wasn't aware of that! Well this got quite a bit more interesting...!

Ahh I agree it would be much better to have 1 table for all events. For now, I'll rework the page to display the last 30 of any event type. Let's see how fast it is... at the moment the load time is about 5 seconds.

matyikriszta:
This has been fixed by #551. Closing this issue.

PR #538, opened 2017-03-28
Limit Events Index page to 30 most recent

dondonz:
Hello,

I've updated the Events Index Page to only display the most recent 30 events. This is addressing issue #536. 

I have created the scope 'past_display' to only access the latest 30 events. As we are storing different event types in separate tables, a separate DB call has to be made for each of the event types (Workshop, Course, Meeting and Event).

I have also added the line "Codebar has hosting [X] weekly workshops, Codebar monthlies and all-day special events since 2013." I'm open to suggestions on the wording of this line.

I retained the original 'past' scope as other methods depend on this.

The N+1 problem is present in the DB call. I'm happy to address this in a separate pull request.

I'm happy to also hear your feedback regarding how I've named methods and variables - please don't be shy!

D


app/models/concerns/listable.rb
```diff
@@ -1,12 +1,14 @@
 module Listable
+  NUMBER_OF_PAST_EVENTS_FOR_INDEX = 30
```
BPScott: Hardcoding limits feels like it limits flexibility. For instance consider if we wanted to put a little box elsewhere on the site that shows the last three events, we'd need a new scope rather than being able to have
`Course.past_display(30)` on one  page and `Course.past_display(3)` on another. What do you think to having this as parameter that you pass into the scope rather than something hardcoded? 30 could still remain the default so `Course.past_display()` would still return 30 results

app/views/events/_events.html.haml
```diff
@@ -4,4 +4,4 @@
       .row.grouped_events
         .medium-12.columns
           %h4= date
-        = render workshops
+        = render workshops
```
BPScott: Leave a blank line at the end of these files (as per the [POSIX standard](http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline)). It's a relic of history but still good practice.

Unrelated we should add an [.editorconfig](http://editorconfig.org/) file to enforce that, but that's a different PR.

app/views/events/index.html.haml
```diff
@@ -9,6 +9,12 @@
...
     .large-12.columns
-      - if @past_events.any?
+      - if @past_events_display.any?
         .past#past
-          = render partial: 'events', locals: { grouped_events: @past_events }
+          = render partial: 'events', locals: { grouped_events: @past_events_display }
+          Codebar has hosted 
+          = @all_past_events_count
+          weekly workshops, Codebar monthlies and all-day special events since 2013.
+          %br
```
BPScott: `<br>` as a mechanism of applying spacing is a bit hacky, could this spacing be done using CSS margins/padding instead?

app/models/concerns/listable.rb
dondonz: Thanks, that's a better idea! I've popped your changes in the latest push.

app/views/events/_events.html.haml
dondonz: Thanks for clearing that up - I was puzzled by the newline symbol in GitHub.

app/views/events/index.html.haml
dondonz: I've changed the scss file to add margin to #past instead - agreed that `<br>`s have no place here!

app/controllers/events_controller.rb
```diff
@@ -4,12 +4,14 @@ class EventsController < ApplicationController
   before_action :is_logged_in?, only: [:student, :coach]
 
   def index
-    events = [ Workshop.past.all ]
-    events << Course.past.all
-    events << Meeting.past.all
-    events << Event.past.all
-    events = events.compact.flatten.sort_by(&:date_and_time).reverse.group_by(&:date)
-    @past_events = events.map.inject({}) { |hash, (key, value)| hash[key] = EventPresenter.decorate_collection(value); hash}
+    record_limit = 30
```
skattyadz: If we're going with a default limit in the scope then we might not need to set it here any more

skattyadz:
I just left a little comment about duplicating the limit of 30 items. Personally I don't mind it at the controller level rather than in the model, but in this case we don't need to set the value to 30 twice

I agree that having multiple models makes this action pretty hairy. We could potentially be loading 120 records only to show 30 of them. Fun!

app/models/concerns/listable.rb
```diff
@@ -4,9 +4,10 @@ module Listable
   extend ActiveSupport::Concern
 
   included do
-
     scope :upcoming, -> { where("date_and_time >= ?", Time.zone.now).order(:date_and_time) }
     scope :past, -> { where("date_and_time < ?", Time.zone.now).order(:date_and_time) }
+    scope :past_display, ->(record_limit=30) { where("date_and_time < ?", Time.zone.now).order(date_and_time: :desc).limit(record_limit) }
```
dansteele: We already have the recent scope that pretty much does just this

app/models/concerns/listable.rb
```diff
@@ -4,9 +4,10 @@ module Listable
...
+    scope :past_count, -> { where("date_and_time < ?", Time.zone.now).count}
     scope :recent, -> { where("date_and_time < ?", Time.zone.now).order(date_and_time: :desc).limit(NUMBER_OF_RECENT_WORKSHOPS_TO_RETRIEVE) }
```
dansteele: We instead should adjust this constant

app/controllers/events_controller.rb
```diff
@@ -4,12 +4,14 @@ class EventsController < ApplicationController
...
+    events = [ Workshop.past_display(record_limit) ]
+    events << Course.past_display(record_limit)
+    events << Meeting.past_display(record_limit)
+    events << Event.past_display(record_limit)
+    events = events.compact.flatten.sort_by(&:date_and_time).reverse[0...record_limit].group_by(&:date)
```
dansteele: If we return the correct number of total events from the AR queries then we don't need to slice the array. By not slicing it, we would
- Get faster queries as less data needs to be fetched originally
- Get faster compacting/flattening/sorting as there is less data in the array

skattyadz: There's no real way to know how many events/meetings/courses there are in the most recent 30 once they're joined. I think this is *better*, but we need an activerecord-level solution on top soon

dansteele: I 100% agree - to rephrase -
Let's say we want to return 40 (for easier maths)

I think it's better that we return 10 of each, and then sort everything, and then output everything
rather than
returning 40 of each, sorting everything, and then outputting the first 40 (which is what this currently does)

skattyadz:
We can agree to disagree but I'd much rather have a slow query than
misleading information on the site. It might look like there aren't events
on dates where there are

dansteele:
@skattyadz think there's still some miscommunication :)

- This is only for events in the past
- My suggestion doesn't change the logic, just offers a performance/clarity improvement

dondonz:
Folks, thanks for the comments. Unfortunately at the moment I'm so busy, I can't contribute to the app. I understand there was another PR and I'm happy to go for that one instead of merging mine.

matyikriszta:
thanks for letting us know @dondonz and for your work. Closing this PR and moving the convo to #551

Issue #539, opened 2017-03-28
GoSquared logo image


dondonz:
Hello,

The GoSquared logo image in the footer is unfortunately not loading. The original image address was https://dl.dropboxusercontent.com/u/108831317/gosq.png.

Does anyone have the original image? (It was blue with text 'Analytics provided by GoSquared'). I can't seem to find this on the company's website.

PR #545, opened 2017-04-10
fixing the go squared image in the footer

markdessain:
Issue: #539 

Fixes the missing GoSqaured image.

screen shot 2017-04-10 at 21 23 10

I found the original image we used on the Wayback Machine and placed it in the assets rather than relying on an extra dropbox account.


BPScott:
I'd be kinda tempted to delete that image reference (and the whole row that contains it) entirely - it seems to be noise.

matyikriszta:
hey @markdessain thanks for taking a look at this. I agree with @BPScott, we don't really use this service anymore, could we just remove the image + row?

markdessain:
@matyikriszta @BPScott - done

matyikriszta:
@markdessain 👍 I'll merge once all the check have passed

matyikriszta:
awesome, thanks @markdessain

PR #551, opened 2017-05-24
Limit data on events index

dansteele:
The /events page is currently loading unreasonably slowly.

Using an amount of data that closely reflected production data (~550 events), the time to first byte went from (average of 3) 9.8s to 0.6s

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -5,14 +5,12 @@ class EventsController < ApplicationController
 
   def index
     events = [ Workshop.recent.all ]
-    events << Course.recent.all
     events << Meeting.recent.all
     events << Event.recent.all
     events = events.compact.flatten.sort_by(&:date_and_time).reverse.group_by(&:date)
     @past_events = events.map.inject({}) { |hash, (key, value)| hash[key] = EventPresenter.decorate_collection(value); hash}
 
     events = [ Workshop.upcoming.all ]
-    events << Course.upcoming.all
     events << Meeting.upcoming.all
     events << Event.upcoming.all
     events = events.compact.flatten.sort_by(&:date_and_time).group_by(&:date)
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -4,10 +4,10 @@ class EventsController < ApplicationController
   before_action :is_logged_in?, only: [:student, :coach]
 
   def index
-    events = [ Workshop.past.all ]
-    events << Course.past.all
-    events << Meeting.past.all
-    events << Event.past.all
+    events = [ Workshop.recent.all ]
+    events << Course.recent.all
+    events << Meeting.recent.all
+    events << Event.recent.all
     events = events.compact.flatten.sort_by(&:date_and_time).reverse.group_by(&:date)
     @past_events = events.map.inject({}) { |hash, (key, value)| hash[key] = EventPresenter.decorate_collection(value); hash}
 
@@ -25,7 +25,7 @@ class EventsController < ApplicationController
     @event = EventPresenter.new(event)
     @host_address = AddressDecorator.new(@event.venue.address)
 
-    if logged_in?
+    if logged_in? 
       invitation = Invitation.where(member: current_user, event: event, attending: true).try(:first)
       if invitation
         redirect_to event_invitation_path(@event, invitation) and return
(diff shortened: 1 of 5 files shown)


BPScott:
Duplicate of #538

dansteele:
@BPScott Aha I just saw, though I would argue this is a neater implementation :sweat_smile:

matyikriszta:
@dansteele would you be able to close this PR and advise the auteur of the original PR on how they could make their implementation neater? neither of us are ruby/rails devs but we would love to get this merged soon-ish

dansteele:
Sure @matyikriszta

matyikriszta:
thanks, super helpful!

dansteele:
https://github.com/codebar/planner/pull/538#issuecomment-304538246
Re-opening by request @matyikriszta

ChrisOdeon:
👍

app/controllers/events_controller.rb
```diff
@@ -4,10 +4,10 @@ class EventsController < ApplicationController
   before_action :is_logged_in?, only: [:student, :coach]
 
   def index
-    events = [ Workshop.past.all ]
-    events << Course.past.all
-    events << Meeting.past.all
-    events << Event.past.all
+    events = [ Workshop.recent.all ]
+    events << Course.recent.all
```
matyikriszta: @dansteele we do not actually use courses so I think this could be removed for now. The types we use are Workshops, Meetings and Events.

matyikriszta:
This looks good to me @dansteele, besides the little comment saying that courses are not currently utilised in the app. I'm happy to merge but would love to get another ruby/rails devs eyes on this

mble:
LGTM 👍 :shipit:

app/controllers/events_controller.rb
matyikriszta: on further reflection we can just merge this PR in and address the course issue in another PR. It is peppered around the code but is unused, might be good to remove every reference to it except the Model etc. in case we need it in the future.

BPScott:
This results in different ordering behaviour to what is there currently. See my comment on #536.

Consider the case where we are at the start of June. There have been 30 workshops in May, 10 events in April, 10 courses in March  and 10 meetings in February.

The existing logic (and the logic in #538) would display the 30 workshops and the 10 events - the 40 latest things across all event types.
This new logic would display 10 workshops, 10 events, 10 courses and 10 meetings, missing the fact that there were 20 workshops that occurred more recently than any of the courses and meetings.


Admittedly this is a contrived example but the reality is the vast majority of events are workshops, while events, courses and meetings are all much much rarer. This PR in its current state shall result in hiding recent workshops at the expense of showing much older courses, meetings and events. To ensure we show the last X events across all event types that filtering needs to be done in code, which is what @dondonz did in #538.

dansteele:
Understood.
Perhaps it would be nicer to display everything from the last 2 months, rather than N records?

matyikriszta:
@dansteele @BPScott thinking about this, I think this needs to be pulled out into separate PRs. I've been thinking about the structure of the site and how we present workshops, events and meetings. I want to move meetings (that we use for Monthlies) into it's own page where we have a little blurb about what Monthlies are and a list of them (say for the past year). That leaves the Events page with workshops and events. I think this page is predominantly used to RSVP to upcoming events as opposed to looking at old ones. It's more a way for us to show off the type and breath of workshops and events we have run in the past than a strictly chronological order of events (these could be presented on the standalone chapter pages that currently have Upcoming events and Recent events that currently return 1 event. This could be expanded into returning the past 6 months or past 10 events?). 

To be frank my immediate concern is to reduce the size of this page while still showing some info to users. As long as the upcoming events section is accurate I don't really care if there are events in the past that don't show up (I wonder who looks at them anyways).

dansteele:
I agree with the thoughts about splitting it out
Are you saying the PR should be left as-is for now, and start work on breaking out in to separate pages?

matyikriszta:
@dansteele what I'm saying is that we should show the last X events across the workshops and events like @BPScott mentioned above. Then work on extracting Monthlies into their own page, yes. Sorry for the confusion and being indecisive 😬. Super appreciate your help.

BPScott:
+1 having this PR as "the latest 40 events across all event types" and have a separate PR for the next step of splitting pages out". Fixing the load times is a big win that we can have now, while a final layout is worked out.

(Why yes I've been dealing with working out how to wrangle "small iterative PRs that are quickly merged" vs "big bang PRs that last for ages before merge" at my day-job, how could you tell ;) )

zaccolley:
May be missing something but has the idea of having two pages "Upcoming events" and "Past events".

"Upcoming events" will load faster because there's less. They are also the most important?
"Past events" will still load slow. They're not as important and we can afford to do that?

Again may have missed this in the thread so sorry if i didnt understand the current solution.

matyikriszta:
@flimflamjim this is the PR I was talking about

matyikriszta:
has been foxed by #595. closing this PR, thanks for all who worked and commented on it

PR #552, opened 2017-05-29
Organising coaches page by chapter

### LINKED ISSUE #540: Coaches page displays every coach
Hello,

At the moment our coaches page is displaying all coaches. I think it's great to demonstrate the very high number of coaches at Codebar. At the moment we have a few hundred coaches on the Wall of Fame and it's a bit difficult to navigate. 

I'd like to hear what you think - should we:
+ Continue to display all coaches, or 
+ Only display a selection, with a line to recognise the total number of coaches that have visited Codebar since 2013, or
+ Introduce pagination

If we do decide to change the coaches page - I can also help with an earlier ticket regarding ranking coaches on the Wall of Fame (#191).

### LINKED ISSUE #544: Organisers (and Chapters) Page
It'd be nice to have a page listing all of the organisers behind the various chapters of codebar :) also, in the process, listing what all the current different chapters are!

MikeRogers0:
This breaks the coaches page into two main pages:

https://codebar.io/coaches - A list of chapters
https://codebar.io/coaches(/:chapter_name) - The coaches in the chapter

Closes #540 and closes #544

Previews:
image

This page could be better in the way each chapter is listed, happy to make amends based on feedback.

image


app/controllers/dashboard_controller.rb
```diff
@@ -27,28 +27,11 @@ def faq
...
 
-  def wall_of_fame
-    sessions_count = Workshop.count
-    @coaches = order_by_attendance(attendance_stats_by_coach).map do |member_id, attendances|
-      member = Member.unscoped.find(member_id)
-      member.attendance = attendances
-      member
-    end
-  end
-
```
MikeRogers0: I removed all the unused methods.

app/views/coach/_coach.html.haml
```diff
@@ -1,20 +0,0 @@
...
-
-              .small-10.columns
-                =link_to twitter_url_for(coach.twitter) do
-                  = coach.full_name
-                %p= coach.about_you
-                - if coach.skills.any?
-                  %p 
-                    skills: 
-                    - coach.skills.each do |skill|
-                      = link_to skill.name, skill_url(skill.name), class: "skill"
```
MikeRogers0: Rendering `coach/coach` and passing a collection of coaches seemed confusing to me, I renamed & broke out this file to make its purpose more clear.

db/seeds.rb
```diff
@@ -15,7 +15,7 @@
 
     40.times do |n|
       coach = coaches.sample
-      Fabricate(:attended_session_invitation, role: "Coach", member: coach, sessions: workshop.sample )  rescue "Coach already attended"
+      Fabricate(:attended_session_invitation, role: "Coach", member: coach, workshop: workshop.sample )  rescue "Coach already attended"
```
MikeRogers0: When I was seeding my DB I couldn't see any coaches. A validation error was being suppressed by the rescue, so I corrected the validation error.

spec/controllers/coaches_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    it "Returns an array of coaches with those with highest attendance first" do
+      # Make sure neither coach is the same.
+      expect(high_attendance_coach).to_not eq(low_attendance_coach)
+
+      # All coaches have a different attended sessions amount.
+      expect(high_attendance_coach.attended_sessions.count).to eq(2)
+      expect(coaches.collect { |coach| coach.attended_sessions.count }.uniq).to eq([1])
+      expect(low_attendance_coach.attended_sessions.count).to eq(0)
+
+      expect(controller.send(:coaches).first).to eq(high_attendance_coach)
```
MikeRogers0: Using `.send(:` is a bit of a yucky way of testing the `helper_method`, happy to swap out for better solution.

app/controllers/coaches_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  helper_method :chapter
+  def chapter
+    @chapter ||= chapters.find_by!(slug: params[:chapter_slug])
+  end
+
+  helper_method :coaches
+  def coaches
+    @coaches ||= chapter.coaches.sort_by do |coach|
+      coach.attended_sessions.count
+    end.reverse
```
MikeRogers0: I'm not crazy happy about this approach to loading coaches sorted by attended sessions (It's N+2), so there is room to move this over to a single SQL query if we need.

spec/controllers/coaches_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      # Add a session attended to remaining coaches
+      coaches.each do |coach|
+        Fabricate(:attended_session_invitation, role: "Coach", member: coach, workshop: Fabricate(:workshop, chapter: chapter, date_and_time: Time.zone.now+14.days) )
+      end
+    end
+
+    it "Returns an array of coaches with those with highest attendance first" do
+      # Make sure neither coach is the same.
+      expect(high_attendance_coach).to_not eq(low_attendance_coach)
+
```
dansteele: Think we can ditch this as `delete_at` modifies the array, so the same option can't be deleted twice

spec/controllers/coaches_controller_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      # All coaches have a different attended sessions amount.
+      expect(high_attendance_coach.attended_sessions.count).to eq(2)
+      expect(coaches.collect { |coach| coach.attended_sessions.count }.uniq).to eq([1])
```
dansteele: Prefer [`map` over `collect`](https://github.com/bbatsov/ruby-style-guide#map-find-select-reduce-size)

dansteele: Not too sure what this is doing actually, could you please expand a little? :sweat_smile: 
To me, it looks like you want to test that they all have a different sessions count, but then you're asserting they all have `1` :thinking:

spec/controllers/coaches_controller_spec.rb
dansteele: I'd argue we should generally stay away from this approach, as you have also caught a whiff of that code smell.
The way I'd go about testing this controller is to create the data, then `visit` the page, and assert the controller `[assigns](https://relishapp.com/rspec/rspec-rails/docs/controller-specs)` things correctly, and the page has the correct content.

Relevant video: https://www.youtube.com/watch?v=URSWYvyc42M

(Relevant screenshot from the video)
image

spec/features/coaches_by_chapter_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+      expect(page).to_not have_content student.name
+    end
+
+    # Does not show coaches from other chapters in the list
+    visit coach_path(chapter_slug: chapter_without_coach.slug)
+    within ".coach-list" do
+      expect(page).to_not have_content coach.name
+    end
+  end
+end
```
dansteele: Following on from my last point, I would simply expand this spec

app/controllers/coaches_controller.rb
MikeRogers0: I've moved this logic into the chapter model (I think it belongs there).

https://github.com/MikeRogers0/planner/blob/9aed586088fb5fd19943b828ee35ff4a772f1df9/app/controllers/dashboard_controller.rb#L30 - I started from here & found the DB to be pretty confusing to work with. 

I've not tried it with production levels of number, how many of members are there normally in a chapter?

> spending a load of time restructuring the data in the app (which definitely needs to be done at some point)

I was thinking adding the following fields:
`Member.coach` - boolean, updated when they attend a session as a coach.
`Member.sessions_attended_count` - A cache counter updated when sessions are attended.

app/controllers/coaches_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def show
+  end
+
+  private
+  helper_method :chapters
+  def chapters
+    @chapters ||= Chapter.by_name
+  end
```
MikeRogers0: I moved the `helper_method` to be above the `private`.

spec/controllers/coaches_controller_spec.rb
MikeRogers0: I hate the `map` method, to me `collect` it much clearer as to what it does :|

I was aiming to check that all the remaining coaches had only attended 1 session. I removed it though, it did feel silly reading it back.

spec/features/coaches_by_chapter_spec.rb
MikeRogers0: What would you expand on for this spec?

MikeRogers0:
Any updates on this or should I close & let someone take over?

dansteele:
Will review again on the weekend, been a very busy week 🙈

MikeRogers0:
🤘🤘🤘 Thank You!

spec/features/coaches_by_chapter_spec.rb
dansteele: It's fine, really. Only if you wanted to 'transfer' some of the controller spec in to here, but I don't think it's _needed_

app/models/chapter.rb
```diff
@@ -34,6 +33,12 @@ def coaches
     members.select(&:coach?)
   end
 
+  def coaches_by_attended_sessions
+    coaches.sort_by do |coach|
+      coach.attended_sessions.count
+    end.reverse
+  end
```
dansteele: @mble How would you avoid this? :)

KimberleyCook:
@MikeRogers0 Hi Mike, this was a very good PR, how come you closed it?

MikeRogers0:
@KimberleyCook I should have added a comment before I closed, but here are the main reasons:

1. I was clearing out my https://github.com/pulls lists & there wasn't much activity in the PR.
2.The code I was contributing was a mediocre & incomplete solution. Someone closer to the project would be better to do this.

PR #573, opened 2017-08-16
Use placekitten instead of dropbox pugs

zaccolley:
Some content doesn't show in the dev version because it was using dead dropbox URLs. This uses placeholder sites instead.


BPScott:
I for one welcome our new placekitten overlords

zaccolley:
:cat2:

matyikriszta:
I'll miss the pugs 😢

PR #576, opened 2017-08-18
Layout, type and headings

zaccolley:
Homepage:
homepag

Homepage mobile:
homepage-mobile

### DIFF
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -17,11 +17,6 @@
   padding-bottom: 20px;
 }
 
-#banner,
-#chapter {
-  margin-top: 30px;
-}
-
 .brand {
   line-height: 5.8rem;
   padding: 0 0.75em;
@@ -136,6 +131,12 @@ u {
 
 .stripe {
   padding: 20px 0;
+}
+
+@media (min-width: 40em) {
+  .stripe {
+    padding: 30px 0;
+  }
 }
 
 .stripe:not(.reverse):nth-child(even) {
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -132,7 +132,7 @@ u {
 }
 
 .stripe:not(.reverse):first-child {
-  padding: 80px 0px 40px;
+  padding: 40px 0;
 }
 
 .stripe:not(.reverse):nth-child(even) {
@@ -141,12 +141,12 @@ u {
 }
 
 .stripe.reverse:nth-child(even) {
-  padding: 80px 0;
+  padding: 40px 0;
 }
 
 .stripe.reverse:nth-child(odd) {
   background-color: #f5f6f7;
-  padding: 80px 0;
+  padding: 40px 0;
 }
 
 .stripe.reverse:first-child{
--- a/app/assets/stylesheets/main.scss
+++ b/app/assets/stylesheets/main.scss
@@ -154,7 +154,6 @@ u {
   margin-bottom: 20px;
 }
 
-
 .stripe:nth-child(even) .grouped-events:not(:last-child) {
   border-bottom: 1px dotted $white;
   margin-bottom: 20px;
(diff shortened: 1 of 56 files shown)


zaccolley:
Will try get this to pass for you tomorrow Kim :+1:

zaccolley:
Have made headings consistent on most pages, and added some where it makes sense. Will update the original post with screenshots.

zaccolley:
I've also moved the 'Blog' link to the right and moved the 'Menu' icon to the right also.

zaccolley:
This starts to cover some of the headings mentioned in #582

zaccolley:
Covers some colour accessibility issues around body content in #583

zaccolley:
hey @KimberleyCook can I get a review on this please? :+1:

app/assets/stylesheets/main.scss
```diff
@@ -131,34 +134,23 @@ u {
   padding-bottom: 10px;
 }
 
-.stripe:not(.reverse):first-child {
-  padding: 80px 0px 40px;
+.stripe {
+  padding: 20px 0;
 }
```
KimberleyCook: How come you removed these paddings?

For me the screenshots you've shown feel a little too squashed together now.

app/views/invitation/show.html.haml
```diff
@@ -1,12 +1,12 @@
 .stripe.reverse
   .row
-    .large-12.columns
-      %h2
+    .columns
+      %h1
```
KimberleyCook: in the world of foundation this `large-12` class should still be there right?

app/views/shared/_title.html.haml
```diff
@@ -1,6 +1,6 @@
-%section.hero.opaque
+%section.stripe.reverse
   .row
```
KimberleyCook: I haven't been able to double check the css so you may need to ignore me, but these deleted id's don't have any related styles that need deleting do they?

app/views/workshops/_workshop_actions.html.haml
```diff
@@ -1,4 +1,4 @@
-%p
+%p.h4
```
KimberleyCook: is this right? Wouldn't this mean it's a paragraph with a `h4` class attached?

zaccolley:
Thanks for this, will get to this on Tuesday!

app/views/invitation/show.html.haml
zaccolley: yep you're right, will add that in, thanks!

app/views/workshops/_workshop_actions.html.haml
zaccolley: These helper classes aren't in the Foundation we're using but the latest version does. The docs are here: http://foundation.zurb.com/sites/docs/typography-helpers.html#typescale

I've added them in to our version, it's so we can have it presentationally look the same as a h4 element but the markup is still a paragraph.

Might be a better way of styling that but I think the classes make some sense, "this should look the same as a h4"?

app/assets/stylesheets/main.scss
zaccolley: I removed these for mobile view, but I'll tweak them cause I agree

app/views/shared/_title.html.haml
zaccolley: I have fixed this in e95457f below

app/assets/stylesheets/main.scss
zaccolley: OK I have tweaked the spacing, and the type a bit more, you can see how that looks in the updated screenshots in the original comment on this PR.

zaccolley:
I've restructured the homepage a bit more. Have created a new 'Support codebar' section at the bottom as it seems the 'Sign up for codebar' is important to keep clearer.

What do you think? @KimberleyCook @matyikriszta

zaccolley:
closing cause i dont want to have to keep this synced with master

PR #577, opened 2017-08-18
Event component tweaks

zaccolley:
This updates the look of the event component across the site.
It also updates the link styling across the site.

Before:

image

After:

image

### DIFF
--- a/app/assets/stylesheets/foundation_and_overrides.scss
+++ b/app/assets/stylesheets/foundation_and_overrides.scss
@@ -130,11 +130,18 @@ $large-up: "#{$screen} and (min-width:#{lower-bound($large-range)})";
 
 // We use these to control header font sizes
 $h1-font-size: rem-calc(37);
-// $h2-font-size: rem-calc(37);
-// $h3-font-size: rem-calc(27);
-// $h4-font-size: rem-calc(23);
-// $h5-font-size: rem-calc(18);
-// $h6-font-size: 1rem;
+$h2-font-size: rem-calc(37);
+$h3-font-size: rem-calc(27);
+$h4-font-size: rem-calc(23);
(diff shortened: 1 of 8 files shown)


KimberleyCook:
What now happens to the `manage` button if you're an organiser?

screen shot 2017-08-19 at 19 52 01

zaccolley:
They go to the side of the title. I'll get you a screenshot tomorrow.

I also want to change the links back as I'm not convinced they are different enough from the body text. Will submit a better solution as a separate PR.

KimberleyCook:
@zaccolley cheers, and yeah I'm happy to change links as the dotted pink underline isn't that nice but not sure this solution makes it clear that it's a link.

zaccolley:
image

Have updated the links again to make them more blue so they stand out. Here's a screenshot of what it looks like to get attending or see manage button.

zaccolley:
can I get a review of this too @KimberleyCook ? thanks!

KimberleyCook:
I'm not 100% sure these now look like links, @matyikriszta what do you think?

zaccolley:
I'm going to remove the links change and then submit that a seperate PR

zaccolley:
image

OK - have reverted the link changes, here's what it looks like now

zaccolley:
closing cause i dont want to have to keep this synced with master

zaccolley:
@despo Need to fix this PR but in terms of the style what do you think?

despo:
Hi @zaccolley - I had a look at this last night and overall it looks good, however some of the partials for  the event types have not been converted - the event below specifically is of type `Meeting` so they need to be updated to.

(Excuse the bad screenshot, no images on local)
screen shot 2018-05-01 at 08 09 14

PR #579, opened 2017-08-20
Add Slack button to footer

zaccolley:
image


zaccolley:
This is buggy - can revisit may need to update the Heroku app

PR #592, opened 2017-08-29
Improve flakey jobs test

zaccolley:
PRs keep failing because there is an ambiguous test that sometimes passes sometimes fails. This should make it so this test passes reliably.


zaccolley:
@KimberleyCook can we get this in to reduce confusion for anyone opening a PR?

KimberleyCook:
@zaccolley All done, thanks for prodding me about merging :)

PR #595, opened 2017-10-01
Limit events page to only show 40 most recent past events

flimflamjim:
Combine all event types, sort by date and then return 40 most recent.
Introduce new temp variable events_hash_grouped_by_date to make type conversion from array more explicit.

We tried to make the simplest possible change which satisfies the requests as specified in @matyikriszta's comment: https://github.com/codebar/planner/pull/551#issuecomment-305225200, specifically:

> what I'm saying is that we should show the last X events across the workshops and events

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -8,8 +8,9 @@ class EventsController < ApplicationController
     events << Course.past.all
     events << Meeting.past.all
     events << Event.past.all
-    events = events.compact.flatten.sort_by(&:date_and_time).reverse.group_by(&:date)
-    @past_events = events.map.inject({}) { |hash, (key, value)| hash[key] = EventPresenter.decorate_collection(value); hash}
+    events = events.compact.flatten.sort_by(&:date_and_time).reverse.first(40)
+    events_hash_grouped_by_date = events.group_by(&:date)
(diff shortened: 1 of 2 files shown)


app/controllers/events_controller.rb
```diff
@@ -8,8 +8,9 @@ def index
     events << Course.past.all
     events << Meeting.past.all
     events << Event.past.all
```
barisbalic: If the limit is 40 events, you can afford to limit these three calls to fetching the last 40 results for `Course`/`Meeting`/`Event`.

barisbalic:
Sorry, at work so my replies are fairly brief, didn't mean to sound dismissive or arrogant, please forgive me if they came across that way!

kevinpmcc:
(context I've been pairing with @flimflamjim on this pull request)

No need for apologies, we really appreciate the review.  

We were very keen to keep the changes as small and explicit as possible so when the future changes are made its as easily as possible to move the repeated behaviour into whatever the future solution is.  More context for all of this is in the #551 thread. 

This request should solve the immediate issue of the website falling over/being very slow on event's listing page. We'd like to get this merged, and then further optimizations are done in further pull requests.

barisbalic:
@kevinpmcc @flimflamjim it makes good sense, glad my comment was useful!

matyikriszta:
thanks folks, this looks great, hopefully will help with our performance issues

PR #596, opened 2017-10-05
Switch to percentage font size to respect browser preferences

BNewing:
This sets the base font size to a percentage, so that the rems in the project respect browser preference. This previously wasn't happening.

We've set the base to 100%, which makes the base 16px. It was previously 14px. This makes the font more readable, particularly on the FAQ page.


BPScott:
What's the difference between doing this and setting the `$rem-base` to be 16px?

BNewing:
16px is a fixed pixel size, whereas if you set it as 100%, it means 100% of what the user has set in their settings page. (https://kyleschaeffer.com/development/css-font-size-em-vs-px-vs-pt-vs/)

It might be that I've misunderstood how the $body-font-size variable works and maybe there is a mixin on it which converts it to a non-pixel value somewhere. There is also probably a better way of implementing it! So my fix may not be the right or best one!

 But, at the moment for the live site, if in Chrome you go to 'Preferences...' and then change font-size to 'Very large', nothing happens on the codebar website. If you check out my PR and do the same, it does all change to respect browser preferences.

BNewing:
This is referenced in Heydon Pickering's 'Inclusive Design Principles':

image

I've never worked with Foundation before these recent PRs, but this post seems to indicate that others have done the same sort of thing with it:

https://foundation.zurb.com/forum/posts/25086-working-with-rem-unit-in-foundation

zaccolley:
Will review this on Sunday @BNewing looks cool nice one

Issue #543, opened 2017-04-04
About Page/FAQ Page in the Top Nav Bar?


BNewing:
I've always felt like the site could do with an about page, or maybe just a link to the FAQ page up on the top nav bar. If you know what codebar is, you don't need it, but with setting it up in a new city, people aren't always sure what it is.

Maybe the nav bar as a whole could do with a bit of a restructure, and some sub navs adding in, to fit it in. It may look a bit cluttered otherwise.

matyikriszta:
@BNewing I agree the nav bar could use some restructuring and would love to add the FAQ link up there, but I think we also need to do a better job explaining what codebar does on the homepage. I reckon if we do that we won't really need an about us page. Thoughts?

zaccolley:
Have some thoughts on the nav restructuring here: https://github.com/codebar/planner/issues/580

BNewing:
It would be great to have the FAQ page up there with maybe the answers padded out and line heights being changed a bit so it's a bit easier to read. Clearer/more info on the home page would be great too! I've directed lots of people to the site but they just don't get what it is atm 

I would also really love a page that has like a step by step guide on things like how to sign up, what is GitHub and why do you have to have an account for it and what is it used for (complete beginners won't know but need an account with it to get a log in for codebar)

I'm happy to have a go at any of this if I can get it set up! Might ask Marten (coorganiser at Oxford) to help me get it set up soon as he does Ruby :) (and I'm a front end dev with no Ruby experience)

zaccolley:
Would love to work with you on this too! Let me know if you need help setting up the environment on the Slack!

Gonna see if can find a content designer to help plan this work.

BNewing:
Ooh thank you!! I would love that :D When I get around to it I will DM you :) 

That sounds good!

BNewing:
I've had another thought about this. How does this idea sound: 

an about page with the subheadings, 'sponsors', 'coaches', and 'FAQ'.

Then the FAQ page is up the top, things are grouped together a bit more and the nav bar will also be a bit tidier. Open to other suggestions or counter thoughts though!

PR #597, opened 2017-10-05
Relocated sponsors, faq and coaches into submenu

BNewing:
Related to #543

This creates an 'About' section on the nav bar, which has a submenu. The submenu contains the FAQ page, the Sponsors page and the Coaches page. The 'About' doesn't link to anything.

We're not sure about the styling and wanted to get some more thoughts on it. What do you think?

screen shot 2017-10-05 at 20 14 57


notapatch:
Hi @BNewing

Not a maintainer. Just to say that what you were seeing was a flaky error that's now dead. I have re-based the pull request and it works fine, locally. Clearly, it's two years later... and I have no idea if the requirements are the same?

PR #598, opened 2017-10-05
Increases question font size and reduces line length on FAQ page

BNewing:
This increases the font size of the questions FAQ page so that it's a big easier to read. We've also added in some padding to the top of each question to space them out a bit.

We've also decreased the column width so that the line length is shorter. This also increases readability.

screen shot 2017-10-05 at 20 20 32

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -16,7 +16,7 @@
 
     = csrf_meta_tags
 
-  %body.no-js{ 'class': "#{params[:controller]} #{params[:action]}"}
+  %body.no-js{ 'class': "#{params[:controller]}-#{params[:action]}"}
     .off-canvas-wrap{ 'data-offcanvas': true }
       .inner-wrap
         #top
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -16,7 +16,7 @@
 
     = csrf_meta_tags
 
-  %body.no-js
+  %body.no-js{ 'class': "#{params[:controller]} #{params[:action]}"}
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.haml
```diff
@@ -16,7 +16,7 @@
 
     = csrf_meta_tags
 
-  %body.no-js
+  %body.no-js{ 'class': "#{params[:controller]} #{params[:action]}"}
```
BPScott: What about a single class that is the controller and action concatenated e.g. "dashboard-faq" rather than "dashboard faq"? There are plenty of routes where the action is just "show" so it seems odd to have lots of different pages all with a body class of "show".

BNewing:
That is a good idea @BPScott - I've implemented it :) 

The Travis CI build is failing, and looking through the details it seems to be linked to the jobs board area of the codebase, but I don't understand it beyond that (this is the first rails and Travis CI project that I've worked on). Any help would be appreciated if anyone has the time to have a look! Otherwise I'll ask @marten at the next codebar Oxford event Tuesday :)

BPScott:
You've updated the class that goes into the HTML, but not the CSS file, now `body.dashboard.faq dl dt` doesn't match anything as that's not what the classes on the body element are called

BNewing:
Oh of course >.<  - silly mistake!

KimberleyCook:
Heyy @BNewing, thanks for this,  could we makes one small amend please. Could we have it so that the text is left aligned with the title please? I feel this looks a bit strange the container being centred but the text still being left aligned.

BNewing:
Added in :)

image

Issue #599, opened 2017-10-09
Clarify text around the sign up process


BNewing:
At the moment, we publicise codebar Oxford by mostly directing people to codebar.io/oxford to sign up. However not all of them have much context for what codebar is, and so I imagine that the current sign up page comes across a bit empty. I'd like to propose some changes to the text on the sign up page, so that it reads as follows:

> "Hi and welcome to codebar! Check out our <a href='http://codebar.io/faq'>FAQ</a> page for any questions you might have, and feel free to email us at `chapteremailaddress` if you can't find the answer you're looking for. Otherwise, use the button below to start the sign-up process."

And then on the following page which talks about eligbility, could we add in a paragraph about GitHub? If they are complete beginners, they'll have no idea what it is or why they're being sent there and it might be super off-putting and mean they don't sign up. The text I'd like to suggest is:

> "To sign up, you'll be taken to something called GitHub. If you don't have an account with GitHub yet, you'll need to create one to sign up. It doesn't take long and it's something that will come in handy when you get started with coding. 

> If you have any questions about the process or would like help with it, feel free to contact the chapter at `chapteremailaddress` "

The above is just first draft text, very open to feedback! It do think adding more text in would make a huge difference to the user journey.

If it all sounds okay I'd like to have a go at implementing it with some help from @marten if he's interested :D

matyikriszta:
Fixed by #600

PR #600, opened 2017-10-11
Updates to text around the sign up process

BNewing:
At the moment, the codebar.io/chaptername page is the first page that our codebar Oxford posters direct to. That page doesn't have very much information on it, so I've added in some text which directs people to the FAQ page and clarifies things a bit:

image

I've also added some text about GitHub onto the second page of the sign up process to warn people that they're about to leave the codebar site, just in case the transition is unnerving. I've chosen to not include a link to GitHub here because they'll be taken there anyway.

image

Feel free to suggest changes to the wording; I'm a bit sleepy at the time of writing them. 

#599

### DIFF
--- a/app/views/chapter/show.html.haml
+++ b/app/views/chapter/show.html.haml
@@ -3,7 +3,7 @@
     .large-12.columns
       %h4= @chapter.name
       %p.lead
-        = t('chapters.intro')
+        = raw t('chapters.intro', email: @chapter.email)
       - if logged_in?
         = render partial: 'subscriptions'
       - else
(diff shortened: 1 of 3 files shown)


KimberleyCook:
@BNewing this is so great thank you, once Travis passes I'll merge in 😁

BNewing:
Thanks @KimberleyCook ! I've checked the Travis build fail details and it looks like it's something to do with the jobs stuff, but I've no idea how to fix it. Do you have any idea what might be causing that?

marten:
That test failure is unrelated. It's due to the job fabricator choosing job titles at random, and thus sometimes choosing duplicate ones, and in that case Capybara can no longer tell which link to click on (and thus errors). I'll make a PR to fix that, but I suggest we ignore the failing test and merge this if it's good to go.

KimberleyCook:
Wooo Travis likes us. Thanks so much for this @BNewing 😁

PR #602, opened 2017-10-11
Eligibility page wording

richardwestenra:
- Add "Our coaches, organisers and sponsors donate their time and money to help those who have faced unfair barriers to entry into the tech world." to the eligibility page.
- Add a paragraph about catering to disabled students on the eligibility page

### DIFF
--- a/app/views/dashboard/participant_guide.html.haml
+++ b/app/views/dashboard/participant_guide.html.haml
@@ -54,7 +54,7 @@
 
       %p Bearing that in mind, it is sometimes difficult for us to identify whether attendees meet the eligibility criteria based on their first and last name. We will try our best to verify via social media profiles beforehand, but if we cannot, we will contact you asking for a short confirmation that you have read this document and assert that you meet one or more of the criteria. We will never ask participants to specify which group.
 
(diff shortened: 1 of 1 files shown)


app/views/dashboard/participant_guide.html.haml
```diff
@@ -54,8 +54,12 @@
...
+      %p codebar is first and foremost a resource for students who cannot afford formal training through coding bootcamps or university degrees, for financial reasons or otherwise. Our coaches, organisers and sponsors donate their time and money to help those who have faced unfair barriers to entry into the tech world. We have strived to create an inclusive, welcoming community, and so far, we have not had to implement a formal policy to enforce the eligibility criteria -- we have left this to the honesty and good faith of the students.
 
       %p If you are unsure whether you are eligible, please get in touch. If we determine that you have signed up and do not belong to an eligible group, we will permanently suspend your student invitations without warning.
 
       %p We take this issue very seriously, and we appreciate your cooperation and understanding.
+
+      %h3#disability Students with Disability
```
KimberleyCook: This amend is good thanks, however I don't think we need the id of `disability` added to the h3...unless we're going to link at the top to anchor to this in which case we'll keep it but then that link needs adding :)

richardwestenra: @KimberleyCook I added it to the in-page nav in 5df2d4b, so this is resolved.

Though it's useful for headings to have unique anchors whether or not you explicitly link to them, because it's often useful for people to be able to link to individual sections, whether or not you have navigation on the page.

PR #601, opened 2017-10-11
add company number to footer

KimberleyCook:
Added the company number to the codebar website.

screen shot 2017-10-11 at 19 59 50

PR #605, opened 2017-10-12
Fix intermittent Jobs test failures

marten:
The fabricator for jobs was choosing job titles at random. This meant that tests could not rely on the presence or absence of them on HTML pages, at least when a test case generated multiple job postings. 

For instance, sometimes the random number generator would choose the same title for an expired posting as one for a current posting, and thus a `expect(page).to_not have_content(expired_job.title)` would fail (because it would have the title for the current job). Whenever asserting on a specific string of text coming from a model, it's usually best to have that text be explicitly set in the test itself.

I've moved the random job/company title selection into the seeds file because it actually is nice to have realistic names when looking at the site in development mode.

This should solve flakeyness experienced in #594 (well, one of the failing tests was due to this), #598, #600, #601, #602 and probably more.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -12,7 +12,25 @@ if Rails.env.development?
     tutorials = 10.times.map { |n| Fabricate(:tutorial, workshop: workshop.sample) }
     feedback_requests = 5.times.map { Fabricate(:feedback_request) }
     feedbacks = 5.times.map { Fabricate(:feedback, tutorial: tutorials.sample, coach: coaches.sample) }
-    jobs = 5.times.map { Fabricate(:job) }
+
+    job_titles = [
+      'Software Engineer',
+      'Software Developer',
+      'Front-end Developer',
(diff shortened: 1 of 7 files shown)


matyikriszta:
@marten another test seems to be failing now 😆 can you re-create it locally?

marten:
No, but I've seen those issues before. The thing with Capybara is that you have to make sure to check some text is on the page before you can make assertions on the current_path (because the page navigation happens asynchronously, so it might not have happened yet when the `expect(current_path).to eq(step2_member_path)` line executes. I'll add a fix for that.

marten:
Oh, these days there's a `expect(page).to have_current_path(...)` RSpec matcher that has the same auto-waiting behaviour that `have_content` has, so I could just use that. Should hopefully come up with green tests now. :)

matyikriszta:
awesome, thanks @marten

PR #608, opened 2017-10-19
Updates to post-sign up text

BNewing:
At the moment, it isn't very clear that you've not signed up to an event, when you first sign up to codebar. I've had a go at improving the text on page 2 of the sign up process! It now reads as follows:

image

I'm not a Ruby/Rails dev, I've just been poking around and seeing what I can get to work! Does anyone know if it's possible to put in a link to the original chapter page they signed up with? I'm not a huge fan of the bit of text that points them at the home page but wasn't sure what else to put.


KimberleyCook:
@BNewing This is great thank you. Do you mean a normal link inside the text like this `%p Lorem ipsum dolor sit amet, consectetur adipiscing elit.<a href='https://codebar.io/'>Link to codebar in the middle of a paragraph</a> continue with the lorem ipsum.` OR a button which is this `=link_to "I'm a button", "https://codebar.io/", class: 'button tiny'` ?

BNewing:
I mean more like a link that knows where they originally went to, to signpost them back to codebar.io/oxford or codebar.io/london - I've no idea if that is possible!

KimberleyCook:
Ohhh sorry!! Ah that is more advance than my Ruby knowledge too 😂

marten:
The standard Rails way of achieving that is to stick the current URL into the session object before redirecting to the login page. Usually after login you would redirect users back to that URL to continue what they were doing before needing to login.

zaccolley:
Cool, nice work. I'd open the other stuff you're thinking of in another PR and get this one merged.

BNewing:
:D. That sounds sensible if people are happy with the text as is - Marten and I have arranged to meet up to do some open source coding, but that won't be for another week and a half which is a while off

matyikriszta:
@BNewing made two small change requests, once those are addressed I'm happy to merge

BNewing:
Hi @matyikriszta - probably a silly question but where are the two small change requests?

matyikriszta:
@BNewing not silly, I forgot to submit the review. You should be able to see them above now.

BNewing:
Addressed the feedback, thank you Kriszta! :D

app/views/members/step2.html.haml
```diff
@@ -5,9 +5,12 @@
       %p
         You're now signed up to codebar!
       %p
-        You haven't signed up to any events yet. If you'd like to sign up to the next event at your local codebar chapter, visit its page. A full list of chapter pages can be found on the homepage. 
+        You haven't signed up to any events yet. If you'd like to sign up to the next event at your local codebar chapter, visit its page. A full list of chapter pages can be found on the 
+        = link_to("homepage", root_path(anchor: "chapters")) + "."
       %p
-        We also email out invites to workshops. If you tick any chapters in the list below, you'll be put on their mailing list. Just be aware that spots to some of our busier chapters fill up fast!
+        We also email out invites to workshops (aside from Central London ones, where you'll need to sign up to on the
```
matyikriszta: hey @BNewing thank you so much, this is totally on me as I said Central London but can we change this to just say London (that's the official name of the chapter). Once this change is done I'm happy to merge.

BNewing: No worries! Done :D let me know if there's anything else that you'd like changing, no rush or hurry on my end :)

BNewing:
Is there anything else I need to do on this PR for it to be okay :)?

matyikriszta:
@BNewing sorry for the delay on getting back to you on this. Reading through the text I think it is still somewhat confusing and could be made clearer. E.g. first we are saying if they'd like to sign up to the next event at their local codebar chapter, they should visit it's page, but they NEED to be subscribed to the chapter before they can RSVP. Only after this we are saying they need to subscribe to specific chapters. So I think the order is wrong here. It should be:

1. To start attending workshops in a specific location you need to RSVP to that chapter first (buttons below).

2. We email out invitations to workshop for the chapters you have signed up to and you can RSVP to them through the invitation, EXCEPT in London where you need to RSVP through the workshop page. Alternatively you can also RSVP to other chapters workshop through the respective workshop pages.

Would you mind rearranging the text to reflect this order?

Issue #534, opened 2017-03-08
Use thumbnail for avatars


conorcussell:
The avatars used throughout the site are displaying the full size images as they were uploaded, which means at its worst, a 7.3 mb image is being loaded to be displayed at 90x50px

<img width="818" alt="screen shot 2017-03-08 at 12 28 39" src="https://cloud.githubusercontent.com/assets/4960622/23704028/d3ac1290-03fa-11e7-9b95-dd36a69b94d5.png">

The [avatar uploader](https://github.com/codebar/planner/blob/master/app/uploaders/avatar_uploader.rb#L30) has a method to create the thumbs, so I guess this means they'll be available on `avatar.thumb` as per the [carrierwave docs](https://github.com/carrierwaveuploader/carrierwave#adding-versions). I was unable to test this locally as presumably images are only uploaded on production?

KimberleyCook:
Heyy @conorcussell, thanks for noting this down. All examples I just looked at on the website are either pulling in 26px x 26px (homepage and event page) or 56px x 56px (single workshop page)  size images. Are you able to tell me where on the website you saw this happening. Thanks.

conorcussell:
Hey @KimberleyCook. Sorry I thought it was everywhere, but it seems its just on the events page https://codebar.io/events

matyikriszta:
@conorcussell as Kim mentioned this seems to be a non issue, looking at the code the 56px z 56px version of the images are served on the events page. I'm going to close this issue as we could not replicate it.

conorcussell:
@matyikriszta its still an issue as far as I can see. Check the company images on this page https://codebar.io/events

Thoughtbot, Net Native etc are all over 1000px wide and displayed at a max width of 90px.

<img width="329" alt="screen shot 2017-10-22 at 18 15 19" src="https://user-images.githubusercontent.com/4960622/31864421-030992e8-b755-11e7-87ec-9a66330acd8d.png">

matyikriszta:
@conorcussell I'v opened a PR for this at #613, would be good if you could review it. After more digging I noticed we have a thumb size defined for sponsor avatars but we are currently not making use of it. As I don't have intimate knowledge of how image resizing works in Rails I'm not sure if this is a fix but maybe.

conorcussell:
[...]

PR #613, opened 2017-10-22
Use thumb version of sponsor avatar

matyikriszta:
This PR is a potential fix for #534. The sponsor avatar uploader had a thumb size defined but we were not making use of it in the templates.

### DIFF
--- a/app/models/sponsor.rb
+++ b/app/models/sponsor.rb
@@ -12,7 +12,7 @@ class Sponsor < ActiveRecord::Base
   default_scope -> { order('updated_at desc') }
   scope :active, -> { joins(:workshop_sponsors) }
 
-  mount_uploader(:avatar, AvatarUploader) if Rails.env.production?
+  mount_uploader(:avatar, AvatarUploader)
 
   accepts_nested_attributes_for :address, :contacts
 
--- a/app/views/events/_event.html.haml
+++ b/app/views/events/_event.html.haml
@@ -32,6 +32,6 @@
           - if event.sponsors.any?
             - event.sponsors.each do |sponsor|
               = link_to sponsor.website do
-                = image_tag sponsor.avatar, class: 'sponsor'
+                = image_tag sponsor.avatar.thumb, class: 'sponsor'
 
       %br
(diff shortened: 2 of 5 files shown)


conorcussell:
@matyikriszta yep - that should fix it. I was unable to test locally when I opened the issue because carrierwave is not enabled when developing locally. Do you have a staging server we can check it on?

matyikriszta:
@conorcussell yes, we can push it up to staging an see if it fixes the issue. Got to first fix all of the failing tests

conorcussell:
Let me know if you want help? Happy to take a look now if you do!

matyikriszta:
@conorcussell that would be great, thanks! I have tried a few things and nothing worked so far :) But the problem seems to be related to this article: https://github.com/carrierwaveuploader/carrierwave/wiki/How-to:-Use-test-factories

conorcussell:
@matyikriszta to be honest I'd be inclined to do something like this in the template.

```ruby
- if (sponsor.avatar.respond_to?(:thumb))
  = image_tag sponsor.avatar.thumb, class: 'sponsor'
- else
  = image_tag sponsor.avatar, class: 'sponsor'
```

I couldn't find any way of mocking different image sizes! Someone might have a better idea, but I don't see any real issue with this. Bare in mind I haven't been writing any rails code for at least a year though...

despo:
@matyikriszta  I've had a look at this and to resolve the failures we need to make the following changes:

https://github.com/codebar/planner/blob/master/app/models/sponsor.rb#L15 
Make uploader available in any environment, not just production

```ruby
 mount_uploader(:avatar, AvatarUploader) 
```
https://github.com/codebar/planner/blob/master/spec/fabricators/sponsor_fabricator.rb#L11
Update sponsor fabricator to associate an uploaded image with the avatar
```ruby
  avatar { Rack::Test::UploadedFile.new(Rails.root.join('app', 'assets', 'images', 'logo.png'), 'image/jpeg') }
```

https://github.com/codebar/planner/blob/master/spec/models/sponsor_spec.rb#L40
Set the subject to a new factory with a nil image as setting the image to nil, won't update it.
```ruby
subject(:sponsor) { Fabricate.build(:sponsor, avatar: nil) }
```
https://github.com/codebar/planner/blob/master/spec/controllers/admin/sponsors_controller_spec.rb#L8
Set the avatar to an uploaded image.
```ruby
 let(:avatar) { Rack::Test::UploadedFile.new(Rails.root.join('app', 'assets', 'images', 'logo.png'), 'image/jpeg') }
```

matyikriszta:
@despo thanks for looking into this, just testing this now

PR #614, opened 2017-10-23
Prevent main nav menu links from being focusable when closed

### LINKED ISSUE #609: [accessibility]Tab order - side nav still tab focusable when off page
At the moment, if you try to use the site with just a keyboard, the side nav isn't great. Even though it's hidden off the page initially, it's still focusable by tab. So if you're tabbing through the top nav bar, once you get through that the tab focus will disappear away off of the page. 

One possible fix could be this, from a video we watched at work the other day:

**https://www.youtube.com/watch?v=fGLp_gfMMGU**

richardwestenra:
This is unfortunately a very hacky fix, because [Foundation has terrible accessibility (which they freely admit in their documentation)](https://foundation.zurb.com/sites/docs/v/5.5.3/components/offcanvas.html#accessibility).

I think it works as a temporary workaround, which we can remove when we one day redesign the site and remove Foundation.

Fixes #609

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -45,5 +45,20 @@ $(function(){
 
   $('#member_lookup_id').chosen().change(function(e) {
     $('#view_profile').attr('href', '/admin/members/' + $(this).val())
-  })
+  });
+  
+  // Prevent tabbing through side menu links when menu is hidden.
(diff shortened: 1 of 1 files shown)


KimberleyCook:
Sorry this took so long to get merged in, doing it now :)

Issue #588, opened 2017-08-22
[Accessibility] Using labels for countries


zaccolley:
In the events component we use `<label>` elements when these could be using `<span>` maybe

clairgriffiths:
Hey @zaccolley I'd like to make this change but could I ask for a brief explanation as to why the change needs to be made? Thanks!

zaccolley:
Hi @clairgriffiths!

Labels are typically used for forms, specifically for labelling input fields. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

We use labels for the countries but I don't think we're using  them correctly.

Making sure we use the right elements for content would hopefully mean assistive technology can understand it better. (Screen readers)

clairgriffiths:
Awesome, thanks for the explanation, I'm going to have a go at making the change. Thanks!

PR #618, opened 2017-10-29
Change label tag to span for events

clairgriffiths:
Fix for https://github.com/codebar/planner/issues/588 which changes a couple of label tags to span tags for better accessibility.


zaccolley:
👍

PR #619, opened 2017-10-31
Updates welcome email instructions

### LINKED ISSUE #615: Welcome email instructions are inaccurate
The welcome email to students to students and coaches includes the following:

```
%p
  - if @member.subscriptions.any?
    You'll automatically receive invitations to our workshops via email.
  - else
    We send out invitations to our events via email.
    = link_to "Sign up from the subscriptions page", subscriptions_url
    to receive invitations to our workshops.
  Places are limited, so be sure to respond quickly if you want to attend.
```
(See `welcome_student.html.haml` and `welcome_coach.html.haml`)

This is inaccurate for the London chapter, and possibly some of the other chapters too, as we no longer send out invite emails for London workshops. We should realign our instructions with better information about how to sign up for workshops.

matyikriszta:
This PR updates the coach and student welcome email copy to include instructions on how RSVPs work for for the London chapter.

Resolves #615 

This is work in progress, any copy change suggestions are welcome.

### DIFF
--- a/app/views/member_mailer/welcome_coach.html.haml
+++ b/app/views/member_mailer/welcome_coach.html.haml
@@ -6,7 +6,11 @@
 
 %p
   - if @member.subscriptions.any?
-    You'll automatically receive invitations to our workshops via email.
+    For our London chapter, you will be able to RSVP to workshops via our website. We open RSVPs at 1PM on the Friday before the workshop. You will be able to register through the workshop page. Join the #london-chapters channel
+    = link_to 'on our Slack', 'http://codebar-slack.herokuapp.com/'
+    to receive a notification when we open RSVPs, or follow us on Twitter
+    - link_to '@codebar', 'https://twitter.com/codebar'
+    . For our other chapters, you'll automatically receive invitations to our workshops to your inbox.
   - else
     We send out invitations to our events via email.
     = link_to "Sign up from the subscriptions page", subscriptions_url
--- a/app/views/member_mailer/welcome_coach.html.haml
+++ b/app/views/member_mailer/welcome_coach.html.haml
@@ -6,11 +6,13 @@
 
 %p
   - if @member.subscriptions.any?
-    For our London chapter, you will be able to RSVP to workshops via our website. We open RSVPs at 1PM on the Friday before the workshop. You will be able to register through the workshop page. Join the #london-chapters channel
-    = link_to 'on our Slack', 'http://codebar-slack.herokuapp.com/'
-    to receive a notification when we open RSVPs, or follow us on Twitter
-    - link_to '@codebar', 'https://twitter.com/codebar'
-    . For our other chapters, you'll automatically receive invitations to our workshops to your inbox.
+    You'll automatically receive email invitations to our workshops as soon as a workshop is open for RSVP.
+    - if @member.chapters.pluck(:name).include?("London")
+      However, our London chapter doesn't send workshop invitations via email, so you won't receive email invites for regular London events. We open RSVPs at around 1pm on the Friday before the workshop, and you will be able to register through the workshop page on our website. To be notified when we open RSVPs, join the #london-chapters channel
+      = link_to 'on our Slack', 'http://slack.codebar.io'
+      or
+      = link_to 'follow us', 'https://twitter.com/codebar'
+      on Twitter.
   - else
     We send out invitations to our events via email.
     = link_to "Sign up from the subscriptions page", subscriptions_url
--- a/app/views/member_mailer/welcome_student.html.haml
+++ b/app/views/member_mailer/welcome_student.html.haml
@@ -4,7 +4,11 @@
 
 %p
   - if @member.subscriptions.any?
-    You'll automatically receive invitations to our workshops via email.
+    For our London chapter, you will be able to RSVP to workshops via our website. We open RSVPs at 1PM on the Friday before the workshop. You will be able to register through the workshop page. Join the #london-chapters channel
+    = link_to 'on our Slack', 'http://codebar-slack.herokuapp.com/'
+    to receive a notification when we open RSVPs, or follow us on Twitter
+    - link_to '@codebar', 'https://twitter.com/codebar'
+    . For our other chapters, you'll automatically receive invitations to our workshops to your inbox.
   - else
     We send out invitations to our events via email.
     = link_to "Sign up from the subscriptions page", subscriptions_url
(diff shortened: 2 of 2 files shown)


app/views/member_mailer/welcome_coach.html.haml
```diff
@@ -6,7 +6,11 @@
 
 %p
   - if @member.subscriptions.any?
-    You'll automatically receive invitations to our workshops via email.
+    For our London chapter, you will be able to RSVP to workshops via our website. We open RSVPs at 1PM on the Friday before the workshop. You will be able to register through the workshop page. Join the #london-chapters channel
+    = link_to 'on our Slack', 'http://codebar-slack.herokuapp.com/'
+    to receive a notification when we open RSVPs, or follow us on Twitter
+    - link_to '@codebar', 'https://twitter.com/codebar'
+    . For our other chapters, you'll automatically receive invitations to our workshops to your inbox.
   - else
```
KimberleyCook: `to your inbox` seems a little weird, could we just say `via email`?

app/views/member_mailer/welcome_coach.html.haml
```diff
@@ -6,7 +6,11 @@
 
 %p
   - if @member.subscriptions.any?
```
richardwestenra: Is there any way to automatically detect whether someone is subscribed to the London chapter? Is this exposed at all in `@member.subscriptions`?

app/views/member_mailer/welcome_coach.html.haml
```diff
@@ -6,7 +6,11 @@
...
-    You'll automatically receive invitations to our workshops via email.
+    For our London chapter, you will be able to RSVP to workshops via our website. We open RSVPs at 1PM on the Friday before the workshop. You will be able to register through the workshop page. Join the #london-chapters channel
```
richardwestenra: I think this is good, but here's how I'd word it (if it were up to me). Up to you which you think is better:

> If you're subscribed to our London chapter, you won't receive workshop invitations via email. We open RSVPs at around 1pm on the Friday before the workshop, and you will be able to register through the workshop page on our website. To be notified when we open RSVPs, join the #london-chapters channel [on our Slack](http://codebar-slack.herokuapp.com/), or [follow us on Twitter](https://twitter.com/codebar).
If you are subscribed to any of our other chapters, you'll automatically receive email invitations to our workshops as soon as a workshop is open for RSVP.

app/views/member_mailer/welcome_coach.html.haml
matyikriszta: sure, can update

app/views/member_mailer/welcome_coach.html.haml
matyikriszta: there is a way but it requires hardcoding the string `london`. I think it opens up a can of worms we could not open up, e.g. then other chapters will want special instructions too, etc.

richardwestenra: We're already hardcoding the string `London`, in the paragraph of instructions we're adding for London users. I don't think this is any different?

KimberleyCook: I think if it's not too difficult to change the text like @richardwestenra suggested depending on chapter subscribed to it would better as it seems a lot of wording, when in fact this is only relevant for 1 out of 17 chapters

KimberleyCook: Being a little bit custom for london alone is ok I think :)

matyikriszta: Not really because what if, for whatever reason, the name of the London chapter changes? Then the code has to be updated here and it might be easy to forget and we'll be banging our head against the table trying to figure out why the wrong instructions are sending. Again, it's very unlikely to happen but it's not good practice. That said if someone else vouches for this solution I'm happy to implement it. @KimberleyCook what do you think?

matyikriszta: ok, can put this in but just keep in mind this might cause some issues in the future 😂

KimberleyCook: We'll have to remember ha, at the level codebar is growing I don't think all text suits all chapters so a few exceptions for London I think is ok :)

richardwestenra: I agree, it's not ideal, but in this situation I think it's okay. Also if we just set it up with a conditional just for london (e.g. `@member.subscriptions === london` or whatever) then if the London chapter's name changes then worst case scenario, they'll just get the same text as before until we notice the error.

matyikriszta: I investigated this issue further and noticed a possible issue. The problem is when a user signs up they can subscribe to more than 1 chapter. So they can subscribe to both London and Brighton. What would the text be then in that case? Seems like to me to send tailored messages based on what chapter the user subscribed to would entail setting up some sort of conditional statement with a lot of edge cases. You'll also notice that a member can sign up and not subscribe to any chapters (and they would still receive this message). For now we are sending them this text: `We send out invitations to our events via email. Sign up from the subscriptions page" to receive invitations to our workshops.` Again, this is factually incorrect but on this case we don't even have the means to check for which chapter they have subscribed to as they have subscribed to none. Any thoughts of how we could get around this?

KimberleyCook: Ahh yes the several chapters problem, good point I forgot about this.

richardwestenra: What about something like this (pseudocode)?

> You'll automatically receive email invitations to our workshops as soon as a workshop is open for RSVP.
**if (member.subscriptions.includes(london)) {**
    However, our London chapter doesn't send workshop invitations via email, so you won't receive email invites for regular London events. We open RSVPs at around 1pm on the Friday before the workshop, and you will be able to register through the workshop page on our website. To be notified when we open RSVPs, join the #london-chapters channel [on our Slack](http://codebar-slack.herokuapp.com/), or [follow us on Twitter](https://twitter.com/codebar).
**}**

matyikriszta: @richardwestenra yep, that looks good, let me look into implementing that

matyikriszta: @richardwestenra @KimberleyCook this is now ready to review

matyikriszta:
@richardwestenra @KimberleyCook this is now ready to review

PR #526, opened 2017-02-10
Setup basic Dockerfile

alexanderkustov:
Fix #242

Since the Vagrant PR was apparently abandoned for over a year I decided to give Docker a go.

Hopefully its a painless setup process following 3 steps and already a basic docker container prepared for ruby/rails that is needed.

If anyone is here more experienced with dockerizing apps let me know how to improve this.

Issue #242, opened 2015-05-15
Have a VM to get started developing with


pedromaltez:
The project isn't that complex but would this be desirable? I've been building one out anyway for my own selfish purposes of learning more about provisioning, but I could clean it up and submit it if it's something that you want.

despo:
@pedromaltez that would be great idea. A VM with PG configured & data would be great, I think these are the main issues to get people started.

If you are up for it I can provide a fresh export of the data this weekend.

pedromaltez:
Oh, great. I was/(am?) having issues with the data and the fact that you need a mailchimp api key (which isn't mentioned anywhere, but `rake db:seed` fails without one!), which I don't know how to fix without setting a key. But it doesn't actually load any data from mailchimp because there is no data in my fresh account and no corresponding mailing list set up because my account is completely new and I don't know what the list is supposed to look like. Would we then seed it with your dbdump instead of the current db:seed script? Guess that would fix all the issues I've had.

despo:
Mailchimp is not really required for development - it's called when subscribing and unsubscribing to different groups to update our mailing lists. I missed this because I was using my own account when initially developing to verify.
 
https://github.com/codebar/planner/blob/master/lib/services/mailing_list.rb#L9 - this should probably change to reflect that, I should add an issue.

pedromaltez:
@ despo, so yes I'd like a dbdump if you can do that! Put it in a gist? Or somewhere that I can get to it. The VM is basically done. But I think I'll need to discuss some of the details with you. We can do that in the PR though. :)

gcapizzi:
Hi folks! Any news on this? Setting up my dev environment right now and going through the same problems (Postgres, Mailchimp) :smile:

pedromaltez:
Hey Giuseppe, I'll post the PR up within a couple of hours. But still no db dump, so hopefully we sort that out soon after I issue the PR.

pedromaltez:
Scratch that, I'll do it 'tomorrow'. Got home pretty late and I'm tired, but it will be up. First thing I'll do once I wake up. :P

gcapizzi:
No worries! Thanks a lot for working on this :)
We can take care of the dump later, seed data should be enough for development.

pedromaltez:
@gcapizzi, no, thank you for nudging me about this. I'm just sorting out some niggles with the provisioner.

pedromaltez:
@gcapizzi #244 fixes the Mailchimp problems.

[...]

PR #624, opened 2017-11-21
Dockerize

despo:
Tackles #242 

This is @alexanderkustov's [PR](https://github.com/codebar/planner/pull/526) with @spike01's [PR](https://github.com/alexanderkustov/planner/pull/1) merged in.

I thought it would be great if we can get this PR in before the 24pullrequests event.  work was already done by @alexanderkustov and @spike01!


despo:
Updated version of #526

zaccolley:
Looks like it's failing CI, know why?

despo:
@zaccolley no idea, It started happening after I rebased. I will try to have a look when I get a minute, hopefully, it's something random.

despo:
@zaccolley it seems to have been random timeout errors when docker was trying to satisfy some dependencies, but all good no.

config/database.yml
```diff
@@ -1,14 +1,13 @@
-development:
+development: &default
   adapter: postgresql
   encoding: unicode
   database: planner_development
   pool: 5
+  username: postgres
+  password:
+  host: db
```
despo: I would prefer this to be localhost so we can run the app with or without docker. Any ideas?

SamirTalwar: Even if you forward the PostgreSQL port, `localhost` still won't work because the Rails server is running in its own container with its own idea of what _localhost_ is. So I suggest an environment variable called `DB_HOST` which is used here, with a fallback of `localhost`.

No idea how to do that in a YAML file though. Perhaps the whole config needs to be overridden? If you set `RAILS_ENV` to `development-docker` in _docker-compose.yml_, and override here:

```yaml
development-docker:
  <<: *default
  host: db
```

Will that work?

Violation of [Rule III](https://12factor.net/config) makes me sad.

SamirTalwar: Looks like [you can embed ERB in _config/database.yml_](http://guides.rubyonrails.org/configuring.html#connection-preference) so this might work:

```yaml
   host: <%= JSON.generate(ENV['DB_HOST'] || 'localhost') %>
```

SamirTalwar: https://github.com/despo/planner/pull/4

despo:
@SamirTalwar thank you for the addition. I added a docker-compose.env to handle the docker configuration so now both using docker or a local environment should be supported! 👍

README.md
```diff
@@ -18,6 +18,21 @@ If you are considering making a PR, please take a look at the Waffle board to se
...
+You will need to have Docker installed, https://docker.com/
+
+The current Dockerfile and docker-compose were closely copied from the guide: https://docs.docker.com/compose/rails/
+
+1. Clone the project.
+2. Run `docker-compose build` to build a docker environment.
+3. Run `docker-compose run web rake db:create db:migrate db:test:prepare` to create and migrate the development and test databases
+4. Run `docker-compose run web rake db:seed` to seed the development database
+5. Run `docker-compose run web rake` to run all the tests.
+6. Run `docker-compose up` to start the app.
```
matyikriszta: I would add a note here about viewing the app at http://localhost:3000/

README.md
```diff
@@ -18,6 +18,21 @@ If you are considering making a PR, please take a look at the Waffle board to se
 
 The following steps walk through getting the application running. For contributing guidelines see [here](https://github.com/codebar/planner/blob/master/CONTRIBUTING.md).
 
+## Getting Started With Docker
+You will need to have Docker installed, https://docker.com/
```
matyikriszta: I would add a note here about making sure Docker is running

README.md
```diff
@@ -18,6 +18,21 @@ If you are considering making a PR, please take a look at the Waffle board to se
...
+
+The current Dockerfile and docker-compose were closely copied from the guide: https://docs.docker.com/compose/rails/
+
+1. Clone the project.
```
matyikriszta: to be very explicit add line about `cd`ing into repo

despo:
@matyikriszta added scripts to simplify the process and updated the README

PR #646, opened 2017-12-09
Update Model and Controller code

swatijadhav:
This PR has some small changes to model and controller.
- Use `before_action` in controller
- Removed `predicate methods` added for `boolean` attributes as rails auto-provides those methods.

Thanks.

### DIFF
--- a/app/models/tutorial.rb
+++ b/app/models/tutorial.rb
@@ -5,6 +5,6 @@ class Tutorial < ActiveRecord::Base
   default_scope -> { order(:created_at) }
 
   def self.all_titles
-    all.map(&:title)
+    pluck(:title)
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/tutorial.rb
```diff
@@ -5,6 +5,6 @@ class Tutorial < ActiveRecord::Base
   default_scope -> { order(:created_at) }
 
   def self.all_titles
-    all.map(&:title)
```
skattyadz: Oh this is a huge improvement

app/models/member.rb
```diff
@@ -60,9 +60,13 @@ def organised_chapters
...
-    return received_student_welcome_email if subscription.student?
-    return received_coach_welcome_email if subscription.coach?
-    true
+    if subscription.student?
+      received_student_welcome_email
+    elsif subscription.coach?
+      received_coach_welcome_email
+    else
+      true
+    end
```
skattyadz: Personally, I don't think this is an objective improvement, but happy to let it go given the rest of the work here!

Issue #173, opened 2014-12-05
The application should be able to support multiple languages


despo:
this involves mostly moving all the text in config/locales/en.yml

Unicornelia:
I will add some translations in German, let me know if somebody already started working on this.

despo:
Support for multiple languages has now been merged in 👍 
Can we close this one @matyikriszta and add individual tasks for areas we would like translated?

matyikriszta:
@despo yes, I will close this issue. @Unicornelia I will create a new ticket for the German translation and assign it to you.

revolunet:
Hi!

We didnt make it in Paris, but we have some crowdsourced CoC translation available here : https://github.com/codebar/paris/blob/master/code-of-conduct.md

happy to help for the FR translations !

Unicornelia:
Sounds good, thanks! I might ask some details as the whole repo is new to me still but im defo keen :)

PR #647, opened 2017-12-12
internationalization (I18n) work

matyikriszta:
## Description
This PR is the beginning of the internationalization work as per #173. I have added support for en, fr and de translations and it currently supports the French translation of the code of conduct (all other pages fall back to the English version for now). I have decided to go with the `codebar.io/locale/example` construct and use scoped routes. I have added a set_locale method to the application_controller which is required to set the locale. The problem is currently only the code-of-conduct route is wrapped in a scope. set_locale works fine on this URL, e.g. when I change `codebar.io/en/code-of-conduct` to `codebar.io/fr/code-of-conduct` it correctly serves the French version, but on all other URLs it appends a `?local=en` param to the URL. I would like all other URLs to fall back onto the English version and for the app not to append the param to the URLs. Not sure how to achieve this exactly.

## Status
Ready to Review

## Related Github issue
#173

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -20,6 +20,8 @@ class ApplicationController < ActionController::Base
   helper_method :current_user
   helper_method :current_service
 
+  before_action :set_locale
+
   def render_not_found
     respond_to do |format|
       format.html { render :template => "errors/not_found", layout: false, :status => 404 }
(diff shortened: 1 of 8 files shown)


matyikriszta:
@despo this is the localization/internationalization work I have started

matyikriszta:
@despo this PR is currently on 🔥 as all of the tests started failing all of a sudden.

PR #651, opened 2018-01-06
Use find_by helper instead of `where(...).first`

skattyadz:
We had some `where(...).first` queries where `find_by(my: thing, another: var)` would be clearer.

I've also moved any existing `find_by_my_variable` dynamic methods to keep it consistent. Initially, I thought they'd been deprecated but after googling it seems they're just not used as commonly as they used to be. eg: https://github.com/bbatsov/rails-style-guide#find_by

It's in two separate commits so I can leave the existing finders intact if needed

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -29,7 +29,7 @@ class EventsController < ApplicationController
     @host_address = AddressDecorator.new(@event.venue.address)
 
     if logged_in?
-      invitation = Invitation.where(member: current_user, event: event, attending: true).try(:first)
+      invitation = Invitation.find_by(member: current_user, event: event, attending: true)
       if invitation
         redirect_to event_invitation_path(@event, invitation) and return
       end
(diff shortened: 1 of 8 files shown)


KimberleyCook:
@skattyadz could you solve the conflicts in this PR please :)

skattyadz:
I will do if it gets approved. Otherwise I'll have to do it twice

matyikriszta:
@skattyadz could you resolve the conflicts so we can merge

skattyadz:
Yep, will do it this afternoon!

app/controllers/events_controller.rb
```diff
@@ -59,16 +59,11 @@ def rsvp
 
   def find_invitation_and_redirect_to_event(role)
     set_event
-    @invitation = Invitation.where(event: @event, member: current_user, role: role).try(:first)
-    if @invitation.nil?
-      @invitation = Invitation.new(event: @event, member: current_user, role: role)
-      @invitation.save
-    end
-
+    @invitation = Invitation.find_or_create_by(event: @event, member: current_user, role: role)
```
skattyadz: @matyikriszta coverage dropped because I shrunk this method. Am I ok to get a merge despite this github check failing?

matyikriszta:
@skattyadz would love to get that coveralls check to pass before we merge

skattyadz:
Rebased

PR #655, opened 2018-01-11
Add time zones to chapters. (Mostly) fixes #559

### LINKED ISSUE #559: Allow different timezone for chapters throughout website
## Description

Now that codebar is operating internationally outside of the UK, we need a way to store event times so they can be translated to different timezones. Currently all times are saved in UK time which is causing issue for our New York chapter. Please refer to @tomafro's description below to see what steps need to be taken to fix this bug.

## To do 📋

* [ ] Submit a pull request, make sure you mention #codebar24PR2017 in the description

skattyadz:
## The problem

#559 describes an issue where the New York chapter were creating events with GMT times. These times displayed correctly on the site but were broken in the application logic. When people tried to RSVP on the day, the cut-off time had already passed in the UK

## The solution

- Each chapter now has a time zone, defaulting to GMT (London). 

- When creating or editing an event, the workshop's `date_and_time` field is a date using the chapter's correct time zone. This means that application logic will do the right conversions when comparing and sorting dates. Dates get saved to the database in the UTC time zone (a Rails feature)

- `Workshop.date_and_time` wraps the `date_and_time` attribute so that's it's always converted to the chapter's time zone when read. This means it's printed correctly on the site rather than printed as that time in the UK

- I've done the same for the RSVP open date field

## Tidying

- On master there are places where we have a datetime AND a time field (saved as a datetime field) on a record. I'm assuming this was to have separate date and time fields in forms, but it's very very confusing and hopefully the new approach is better. For example, `rsvp_open_date` and `rsvp_open_time` had the same value but only one was used. I've replaced it with a single `rsvp_opens_at`

## Caveats

- I haven't done this for other event types, like meetings, courses, and events. It shouldn't be much work but workshops are the first priority

- If you change the time zone of a chapter then the existing workshop times will shift by a few hours. A clever fix would be to store an individual time zone for each workshop but it's overkill. It means we'll need to re-enter the times for any New York workshops already on the site as a one-off (should only take 5 mins).

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -15,10 +15,6 @@ module ApplicationHelper
     current_user.groups.include?(group)
   end
 
-  def can_access?(resource)
-    current_user.has_role?(:admin) || current_user.has_role?(:organiser) || current_user.has_role?(:organiser, resource)
-  end
-
   def has_permission?
     current_user.has_role?(:admin) || current_user.has_role?(:organiser) || Chapter.find_roles(:organiser, current_user).any?
   end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,13 +1,10 @@
 module ApplicationHelper
-  def humanize_date(date)
+  def humanize_date(date, with_time: false)
     human_date = "#{I18n.l(date, format: :day_in_words)}, "
     human_date << "#{ActiveSupport::Inflector.ordinalize(date.day)} "
     human_date << I18n.l(date, format: :month)
-  end
-
-  def humanize_date_with_time(date, time = date)
-    human_date = humanize_date(date)
-    human_date << " at #{I18n.l(time, format: :time)}"
+    human_date << " at #{I18n.l(date.time, format: :time)}" if with_time
+    human_date
   end
 
   def dot_markdown(text)
--- a/app/models/session_invitation.rb
+++ b/app/models/session_invitation.rb
@@ -14,9 +14,7 @@ class SessionInvitation < ActiveRecord::Base
   scope :to_coaches, -> { where(role: 'Coach') }
   scope :by_member, -> { group(:member_id) }
   scope :order_by_latest, -> { joins(:workshop).order('workshops.date_and_time desc') }
-  scope :past, -> { joins(:workshop).where('workshops.date_and_time < ?', Date.today) }
-  scope :last_six_months, -> { joins(:workshop).where('workshops.date_and_time < ? AND workshops.date_and_time > ?',
-                                                      Date.today, Date.today - 6.months)}
+  scope :last_six_months, -> { joins(:workshop).where(workshops: { date_and_time: 6.months.ago...Time.zone.now}) }
 
   def waiting_list_position
     @waiting_list_position ||= WaitingList.by_workshop(self.workshop).where_role(self.role).where(auto_rsvp: true).order(:created_at).map(&:invitation_id).index(self.id) + 1
--- a/app/models/workshop.rb
+++ b/app/models/workshop.rb
@@ -2,8 +2,7 @@ class Workshop < ActiveRecord::Base
   include Invitable
   include Listable
 
-  attr_accessor :local_date
-  attr_accessor :local_time
+  attr_accessor :local_date, :local_time, :rsvp_open_local_date, :rsvp_open_local_time
 
   resourcify :permissions, role_cname: 'Permission', role_table_name: :permission
 
@@ -19,8 +18,9 @@ class Workshop < ActiveRecord::Base
   scope :coaches, -> { joins(:invitations).where(invitations: { name: 'Coach', attended: true }) }
 
   validates :chapter_id, presence: true
+  validates :date_and_time, presence: true
 
-  before_save :combine_date_and_time, :set_rsvp_close_time
+  before_validation :set_datetime_fields
 
   def host
     WorkshopSponsor.hosts.for_session(self.id).first.sponsor rescue nil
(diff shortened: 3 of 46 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,13 +1,10 @@
 module ApplicationHelper
-  def humanize_date(date)
+  def humanize_date(date, with_time: false)
     human_date = "#{I18n.l(date, format: :day_in_words)}, "
     human_date << "#{ActiveSupport::Inflector.ordinalize(date.day)} "
     human_date << I18n.l(date, format: :month)
-  end
-
-  def humanize_date_with_time(date, time = date)
```
skattyadz: Date and time were always the same wherever this was called, so I binned it

app/models/session_invitation.rb
```diff
@@ -14,9 +14,7 @@ class SessionInvitation < ActiveRecord::Base
   scope :to_coaches, -> { where(role: 'Coach') }
   scope :by_member, -> { group(:member_id) }
   scope :order_by_latest, -> { joins(:workshop).order('workshops.date_and_time desc') }
-  scope :past, -> { joins(:workshop).where('workshops.date_and_time < ?', Date.today) }
```
skattyadz: `Date.today` breaks down when you have multiple time zones. Today to who?

This scope was dead code anyway so I removed it

app/models/session_invitation.rb
```diff
@@ -14,9 +14,7 @@ class SessionInvitation < ActiveRecord::Base
...
-  scope :last_six_months, -> { joins(:workshop).where('workshops.date_and_time < ? AND workshops.date_and_time > ?',
-                                                      Date.today, Date.today - 6.months)}
+  scope :last_six_months, -> { joins(:workshop).where(workshops: { date_and_time: 6.months.ago...Time.zone.now}) }
```
skattyadz: `Date.today` breaks down when you have multiple time zones. I changed it to "6 months ago from this time today"

app/models/workshop.rb
```diff
@@ -2,6 +2,8 @@ class Workshop < ActiveRecord::Base
   include Invitable
   include Listable
 
+  attr_accessor :local_date, :local_time, :rsvp_open_local_date, :rsvp_open_local_time
```
skattyadz: These temporary attributes hold the values from the date and time pickers in the form. We don't need to save them to the database

app/models/workshop.rb
```diff
@@ -51,24 +54,17 @@ def past?
     date_and_time.past?
   end
 
-  def today?
-    date_and_time.today?
-  end
```
skattyadz: Dead code

app/models/workshop.rb
```diff
@@ -98,21 +94,45 @@ def location
...
+  end
+
   def self.policy_class
     WorkshopPolicy
   end
 
+  def date_and_time
+    return nil unless super
+    super.in_time_zone(time_zone)
+  end
```
skattyadz: Rails assumes we want dates in GMT because that's the timezone set in the Rails config. Wrapping the `date_and_time` method means it's always in the correct time zone when we need it

app/models/workshop.rb
```diff
@@ -98,21 +94,45 @@ def location
...
+
   def date
     I18n.l(date_and_time, format: :dashboard)
   end
 
+  def time
+    date_and_time.try(:time)
+  end
```
skattyadz: For backwards compatibility

skattyadz:
Frustrating that this is failing coverage now (it wasn't before). Is it because I deleted some code that had coverage, thus reducing the coverage percentage?

skattyadz:
I have one more thing to do on this PR. I think there's an N+1 query if we ever show the date/time of a workshop on a list page. Because to show the date/time you need to know the time zome of its chapter

Can either: 
- Use `includes` to eager load the chapter of the workshop
- Cache the current time zone of the chapter when a workshop is saved

skattyadz:
Ok the feedback rake task has lost coverage in both of my PRs. I can't work out a reason why

matyikriszta:
@skattyadz would be good to figure out why the feedback rake task has lost coverage. Also, there are a number of conflicting files, can these be sorted? I will also wait for @despo to review this PR as well

skattyadz:
Rebased

margOnline:
Nice work @skattyadz. I love that you provided a detailed description of your approach to solving the problem, otherwise I probably would have asked a bunch of questions about why some code was deleted. 

The concern is that this is a big pr. The size may offput potential reviewers. More importantly, it increases the deployment risk.  Have you considered ways to reduce the risk of deployment? One suggestion:

Break the pr into several smaller ones e.g.
* one pr to address just removing redundant code
* one pr to address the back end stuff - migrations, controller logic
* one pr to address the views

app/models/workshop.rb
margOnline: 👍

app/models/workshop.rb
```diff
@@ -98,21 +94,45 @@ def location
...
   private
 
-  def combine_date_and_time
-    self.date_and_time = date_and_time.change(hour: time.hour, min: time.min)
+  def set_datetime_fields
+    new_date_and_time = datetime_from_fields(local_date, local_time)
+    self.date_and_time = new_date_and_time if new_date_and_time
+    new_opens_at = datetime_from_fields(rsvp_open_local_date, rsvp_open_local_time)
+    self.rsvp_opens_at = new_opens_at if new_opens_at
   end
```
margOnline: This method would benefit from splitting into several methods, maybe one for setting `rsvp_opens_at` and one for setting `date_and_time`

skattyadz:
Hi Margo, thanks for your feedback

You're right that this PR is too big and I should have put any tidying in a different PR

In terms of backend/frontend changes I was hoping that splitting into commits would be sufficient, but maybe I could have done that better. I find having dependencies between PRs can get confusing: "don't merge this PR until that PR gets merged" etc but will give that a go 👍

app/models/workshop.rb
skattyadz: Good point

Issue #635, opened 2017-11-29
Add and configure rubocop


matyikriszta:
## Description of the issue 📄

As a measure to achieve better code consistency throughout the app we would like to add and configure [rubocop](https://github.com/bbatsov/rubocop). Once rubocop is added and configured, it would also be great if you could go through the app and fix all rubocop offenses (there will likely be a lot).

## Steps to fix 🛠

Add rubocop to the Gemfile. Setting it up with a default configuration is ok for now. Once configured, run rubocop and fix the offenses.

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request, make sure you mention #codebar24PR2017 in the description
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

bellvat:
Hi!
I'd like to take a stab at this issue if possible!

Thanks!

Monica

despo:
@bellvat Hi Monica, I don't think anyone else has picked this up so that would be great. 😄

bellvat:
Hi @despo!

I was able to install the rubocop gem. Once I ran rubocop, it indicated that there were about 3,800 offenses. I went through a few files to see the general offenses, and the following are a few from the files I looked at:
  - Linelength exceeds 80 characters
  - Missing top level class documentation comment 
  - Single quote preferred over double quotes
  - Missing frozen string literal comment

Rubocop has an automated code review which fixes offenses for you, but it is in experimental stage.
I feel like the automated code review can be utilized, but in small groups of files, as I'm not sure we can trust it to not break code. 

There are some offenses, like line length exceeding 80 characters and adding top level class comments, that cannot be automated.

**Before I start fixing the offenses I wanted to ask if you wanted to take a look at rubocop configuration? There may be some things that you may not feel as strongly about, and may not want rubocop to mark as offenses. Here is the link for configuration in Rubocop documentation: [http://rubocop.readthedocs.io/en/latest/configuration/](url)**

Please let me know!

Monica

matyikriszta:
[...]

PR #656, opened 2018-01-11
Rubocop fixes

skattyadz:
Related to #635

They're mostly obvious except for:

## Increase rubocop line length to 100 chars

This is very subjective but I think this it's more achievable and friendly to contributors. It also removed the majority of offenses in that class

## Using `destroy` instead of `delete` when destroying records

This was my own addition, not from rubocop. Using `delete` skips callbacks and `dependent: destroy` on the record which can trip people up.

## Add policies instead of manual permission checks

I stumbled into this when fixing a rule. It's something I'm going to keep an eye on from now on; we should be using policies for authorization everywhere


skattyadz:
Yay 🙌

PR #649, opened 2017-12-30
Add and configure rubocop #codebar24PR2017

bellvat:
Hi @matyikriszta 

PR for Issue#635

- Added the Rubocop Gem
- Added rubocop configuration file .rubocop.yml at root directory
  - As discussed, disabled rubocop for 1.Missing top level comments 2.Frozen string literals
- Fixing `Line length Exceeding 80 chars`offenses would involve refactoring code into two lines, which might be better suited for someone more familiar with the code base
- Ran auto-correct for various Layout and Style cops using the following command: `rubocop --only <cop name> -a`
  - `Single quote preferred over double quote` offenses were corrected
  - Auto-correct brought down offenses from 3,800 to 1,254 offenses

PR #670, opened 2018-02-01
Regenerate db/schema.rb

skattyadz:
The schema file was automatically changed in #649 to have single quotes instead of double.

Now, when you migrate, Rails regenerates the file with double-quotes and the whole file appears in the git diff.

This PR reverts it to the format the Rails generates so running future migrations just modifies the correct lines in the git history

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,527 +11,529 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161228121833) do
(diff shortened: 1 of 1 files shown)


KimberleyCook:
@margOnline I know you were keen for @skattyadz to break down his previous PR, would you be able to take a look at this, https://github.com/codebar/planner/pull/671, https://github.com/codebar/planner/pull/672 and https://github.com/codebar/planner/pull/673 please?

PR #672, opened 2018-02-02
Tidy dead code

skattyadz:


matyikriszta:
@skattyadz can you comment on what dead code is this removing? has this been replaced by other code? Can you also fix the conflicts?

app/helpers/application_helper.rb
```diff
@@ -18,10 +18,6 @@ def belongs_to_group?(group)
     current_user.groups.include?(group)
   end
 
-  def can_access?(resource)
```
skattyadz: This function isn't called by anything

app/models/chapter.rb
```diff
@@ -18,7 +18,6 @@ class Chapter < ActiveRecord::Base
   def self.available_to_user(user)
     return Chapter.all if user.has_role?(:organiser) || user.has_role?(:admin) || user.has_role?(:organiser, Chapter)
     return Chapter.find_roles(:organiser, user).map(&:resource)
-    []
```
skattyadz: This can't be reached as the line before returns

app/helpers/application_helper.rb
matyikriszta: ok, cool

app/models/chapter.rb
matyikriszta: yep, makes sense

matyikriszta:
@skattyadz coveralls is not liking this one

PR #673, opened 2018-02-02
Add time zones to chapters

skattyadz:
This depends on #669, #670, and #671, so the diff looks larger than it is. Just look at the last commit's changes

## The problem

#559 describes an issue where the New York chapter were creating events with GMT times. These times displayed correctly on the site but were broken in the application logic. When people tried to RSVP on the day, the cut-off time had already passed in the UK

## The solution

- Each chapter now has a time zone, defaulting to GMT (London). 

- When creating or editing an event, the workshop's `date_and_time` field is a date using the chapter's correct time zone. This means that application logic will do the right conversions when comparing and sorting dates. Dates get saved to the database in the UTC time zone (a Rails feature)

- `Workshop.date_and_time` wraps the `date_and_time` attribute so that's it's always converted to the chapter's time zone when read. This means it's printed correctly on the site rather than printed as that time in the UK

- I've done the same for the RSVP open date field

## Caveats

- I haven't done this for other event types, like meetings, courses, and events. It shouldn't be much work but workshops are the first priority

- If you change the time zone of a chapter then the existing workshop times will shift by a few hours. A clever fix would be to store an individual time zone for each workshop but it's overkill. It means we'll need to re-enter the times for any New York workshops already on the site as a one-off (should only take 5 mins).


margOnline:
👍

Issue #677, opened 2018-02-22
Add venue accessibility info to monthlies page


KimberleyCook:
## Description of the issue 📄

In the admin area you are able to gather venue accessibility info, we've implemented this on the workshop pages but not monthlies. I'd like it added where I've put the arrow in the image below:

<img width="775" alt="screen shot 2018-02-22 at 10 30 11" src="https://user-images.githubusercontent.com/2683270/36556867-7a72fc9a-17bb-11e8-997b-9e40c4a6c72b.png">

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request, make sure you mention #codebar24PR2017 in the description
* [x] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

PR #684, opened 2018-03-04
Add venue accessibility info to monthlies page

matyikriszta:
## Description
This PR addresses #677 and adds venue accessibility info to monthlies pages.

### DIFF
--- a/app/views/admin/meetings/show.html.haml
+++ b/app/views/admin/meetings/show.html.haml
@@ -12,7 +12,6 @@
     %i.fa.fa-users
     %label Labels
 
-
 %section
   .stripe.reverse
     .row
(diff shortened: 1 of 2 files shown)


app/views/admin/meetings/show.html.haml
```diff
@@ -24,16 +23,20 @@
...
     .row
       .large-3.small-6.columns#host
         - if @meeting.venue.present?
           %h4 Venue
-          %strong= @meeting.venue.name
-          %br
+          %p
+            %strong= @meeting.venue.name
+          %strong Accessibility info
+          - if @meeting.venue.accessibility_info.present?
```
KimberleyCook: we should probably add line 35 inside the if statement

KimberleyCook:
@matyikriszta I don't think you pulled the latest from master based on the conflict.

matyikriszta:
@KimberleyCook yep, just rebasing, one sec

app/views/admin/meetings/show.html.haml
```diff
@@ -24,16 +23,20 @@
...
 
     .row
       .large-3.small-6.columns#host
         - if @meeting.venue.present?
           %h4 Venue
-          %strong= @meeting.venue.name
-          %br
+          %p
+            %strong= @meeting.venue.name
+          %strong Accessibility info
```
matyikriszta: @KimberleyCook I added it outside the of statement because on the admin page I want to show a message to update a sponsor with the information if there is no accessibility info yet. So the header would always be there.

KimberleyCook: Yeah I realised I didn't mean to add that comment as the code wrong hint I deleted it 😂

PR #692, opened 2018-03-13
Fix n-plus-1 queries

skattyadz:
When we load records from the database we want to include all the data we're going to need. If we haven't 'eager loaded' all the records needed by the views, Rails will automatically fetch the extra records one-by-one which is slow.

Imagine a page that shows 30 workshops and a chapter name and sponsor logos for each one. Normally we'd fetch those workshops with one database call. If we forget to 'eager load' the chapters and sponsors we turn 1 query into 61 queries.

```ruby
Workshop.limit(30)
```

vs

```ruby
Workshop.limit(30).includes(:chapter, :sponsors)
```

To detect these I added [Bullet](https://github.com/flyerhzm/bullet) and made tests fail if it detects this happening

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -59,6 +59,7 @@ group :development, :test do
   gem 'rspec-its'
   gem 'rspec-rails'
   gem 'timecop'
+  gem 'bullet'
 end
 
 group :test do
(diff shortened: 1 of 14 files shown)


BPScott:
Woot

skattyadz:
Don't get too excited; it's only a small number that got picked up
automatically!

despo:
@skattyadz thank you 👍

skattyadz:
Thanks for merging :)

PR #704, opened 2018-04-07
Update copyright to 2018

minaorangina:
Currently still reads as copyright 2017 :)

### DIFF
--- a/app/views/layouts/_footer.html.haml
+++ b/app/views/layouts/_footer.html.haml
@@ -2,7 +2,7 @@
   .row
     .large-3.columns
       %p.copyright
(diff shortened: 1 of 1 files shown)


KimberleyCook:
@minaorangina Is there some way in Rails this could be automated??

BPScott:
```
 &#169; #{Date.today.year} codebar 
```
Seems like it would work

KimberleyCook:
@BPScott exactly what I was after!

@minaorangina Are you able to make the above amend please? :)

minaorangina:
@KimberleyCook @BPScott sure, I'll add it

PR #705, opened 2018-04-08
Add Sentry exception tracking

sandfox:
Any exceptions in production can now be sent to [sentry](https://sentry.io/features/), which makes it easy to summarise and see exceptions as they happen (and in the past) instead of searching through heroku logs.

To enable capturing the exceptions the `SENTRY_DSN` enviroment variable needs to be set in heroku. To get a sentry DSN you'll need to [create an account](https://docs.sentry.io/quickstart/) with sentry (they have a free tier of about 10K events per month).

I haven't done anything fancy like add extra context info (e.g username) to the information that goes to sentry, but the basic info it provides should be a good starting point.

screenshot-2018-4-8 runtimeerror error

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -42,6 +42,8 @@ gem 'gibbon', '~> 1.1.5'
 
 gem 'stripe'
 
+gem "sentry-raven"
+
 group :development do
   gem 'better_errors'
   gem 'letter_opener'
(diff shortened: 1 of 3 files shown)


KimberleyCook:
I'm going to close this PR @sandfox as I don't think we intend to use it. Cheers though!

PR #707, opened 2018-04-18
Update loofah and omniauth to resolve security vulnerability concerns

despo:


greysteil:
Would you be up for creating PRs like this automatically? I just added a "security-update-only" option to [Dependabot](https://dependabot.com) which could be useful?

PR #720, opened 2018-04-30
Work through rubocop errors

despo:


despo:
Related to #635

app/controllers/events_controller.rb
```diff
@@ -23,15 +23,15 @@ def index
...
+    event = Event.find_by(slug: params[:id])
 
     @event = EventPresenter.new(event)
     @host_address = AddressPresenter.new(@event.venue.address)
 
     if logged_in?
       invitation = Invitation.find_by(member: current_user, event: event, attending: true)
       if invitation
-        redirect_to event_invitation_path(@event, invitation) and return
+        redirect_to event_invitation_path(@event, invitation) && return
```
EGiataganas: I think this should be

```ruby
return redirect_to event_invitation_path(@event, invitation)
```

Can you please double check if there is a spec for that?

app/controllers/feedback_controller.rb
```diff
@@ -4,7 +4,7 @@ def show
 
     if feedback_request.nil?
       flash[:notice] = I18n.t('messages.feedback_not_found')
-      redirect_to root_path and return
+      return redirect_to root_path
```
EGiataganas: 👍

app/controllers/events_controller.rb
EGiataganas: I think this should be

```ruby
return redirect_to event_invitation_path(@event, invitation)
```

Can you please double check if there is a spec for that?

despo: Good catch. I'll take a look.

PR #721, opened 2018-04-30
Add unique index to meeting slug and optimise queries based on bullet feedback

despo:
resolves bug noticed on the website last week where an old meeting was retrieved because the new meeting slug was identical.


geopolitis:
Seems good to me 
+1

PR #723, opened 2018-05-03
remove gratipay link from README

despo:


matyikriszta:
@despo not sure how but this change introduced a breaking test, could you fix this before we merge?

matyikriszta:
@despo possibly, let's see if it runs this time

PR #727, opened 2018-05-06
Don’t use docker on travis

despo:
- adds a 3 minute delay
- does not add any value
- breaks coveralls


despo:
@matyikriszta @EGiataganas thanks 👍

PR #728, opened 2018-05-07
Improve test coverage

despo:
Halfway through but the pull request is getting quite big so I think better do it in parts.

Brief overview
- fixes issue where a meeting organiser is unable to RSVP a member that has declined the invitation
- query optimisations
- extract text to locales
- use strong params
- remove non existing routes from configuration `/admin/chapter/workshops`
- display better error messages
- fix broken workshop validation as we can't check for date and time without the chapter's timezone
- optimise sending meeting invitations
- removes unused `invitation` param from `MeetingInvitationsMailer`
- fixes Feedback task so all workshop attendees from the previous day receive the email - as it currently only retrieves the latest one


app/controllers/admin/meeting_invitations_controller.rb
```diff
@@ -2,35 +2,42 @@ class Admin::MeetingInvitationsController < Admin::ApplicationController
...
+                       notice: t('admin.messages.invitation.update_rsvp', name: @invitation.member.full_name)
   end
 
   def create
     member = Member.find(params[:meeting_invitations][:member])
     meeting = Meeting.find_by(slug: params[:meeting_invitations][:meeting_id])
 
-    if meeting.invitations.accepted.map(&:member_id).include?(member.id)
-      return redirect_to [:admin, meeting], notice: "#{member.full_name} is already on the list!"
+    if MeetingInvitation.accepted.where(meeting: meeting, member: member, attending: true).exists?
```
barisbalic: Any reason `accepted` is a scope but `attending` is not?

app/controllers/admin/meeting_invitations_controller.rb
```diff
@@ -2,35 +2,42 @@ class Admin::MeetingInvitationsController < Admin::ApplicationController
...
 
   private
 
   def set_invitation
-    @invitation = MeetingInvitation.find_by(token: params[:id])
+    @invitation = MeetingInvitation.find_by(token: id)
+  end
+
+  def id
+    params.permit(:id)[:id]
```
barisbalic: For a basic read of a specific field, I probably wouldn't bother with this, I guess it's to taste but it doesn't add anything and I'd always go for less code.

despo: Really? I think it makes the code reasier to read. However, I was also wanted to use strong params in preparation for upgrading to rails 5.

app/models/concerns/listable.rb
```diff
@@ -15,7 +16,7 @@ def next
     end
 
     def most_recent
-      past.load.first
+      past.includes(:sponsors).load.first
```
barisbalic: This can be replaced with `joins`?

despo: I fixed this based on the warnings I got from [bullet](https://github.com/flyerhzm/bullet) - I'm not sure tbh but this seems to resolve the error.

despo:
@barisbalic thank you 👍  I'll take a look

PR #729, opened 2018-05-08
Add unsubscribe instructions to monthlies invitation emails

matyikriszta:
## Description

Monthlies invitation emails did not have the unsubscribe link in them and we were getting a lot of emails to remove people. This PR adds the current footer partial with the Unsubscribe link to the emails. It would be good if in the future we could improve upon the copy (let people know which chapter they should unsubscribe from).

## Related Github issue
#621


matyikriszta:
@despo I can't recreate this failing test on my local version. Any idea what might be happening?

PR #732, opened 2018-05-13
Tests and refactoring (with focus on InvitationManager)

despo:


app/models/invitation_manager.rb
```diff
@@ -1,111 +1,59 @@
...
-    end
-
-    if audience == 'coaches'
-      invite_coaches workshop
-      return
-    end
-
-    invite_students workshop
-    invite_coaches workshop
+    invite_students_to_workshop(workshop) unless audience.eql?('coaches')
```
barisbalic: Never see `eql?` used like this, is this a convention established for this codebase?

app/models/invitation_manager.rb
```diff
@@ -1,111 +1,59 @@
...
-  def self.send_monthly_attendance_reminder_emails(monthly)
-    monthly.attendances.map(&:member).each do |member|
-      MeetingInvitationMailer.attendance_reminder(monthly, member)
-    end
-  end
-
-  def self.send_course_emails(course)
-    course.chapter.groups.students.map(&:members).flatten.uniq.each do |student|
+  def send_course_emails(course)
+    students = Member.not_banned.joins(:groups).merge(course.chapter.groups.students)
```
barisbalic: Lot going on here, break it down maybe?

app/models/invitation_manager.rb
```diff
@@ -1,111 +1,59 @@
...
       invitation.update_attribute(:reminded_at, Time.zone.now)
     end
   end
+  handle_asynchronously :send_workshop_attendance_reminders
 
-  def self.send_workshop_waiting_list_reminders(workshop)
-    # Only send out reminders to people where reminded_at is nil, ie. falsey.
-    workshop.waiting_list.reject(&:reminded_at).each do |invitation|
+  def send_workshop_waiting_list_reminders(workshop)
+    workshop.invitations.where(reminded_at: nil).joins(:waiting_list).each do |invitation|
```
barisbalic: Is this an opportunity to use a scope?  "unsent_invitations" or something like that?

app/models/member.rb
```diff
@@ -23,6 +23,15 @@ class Member < ActiveRecord::Base
   validates_length_of :about_you, maximum: 255
 
   scope :subscribers, -> { joins(:subscriptions).order('created_at desc').uniq }
+  scope :not_banned, lambda {
```
barisbalic: Use either `->` or `lambda` for consistency?

app/models/invitation_manager.rb
despo: @barisbalic I don't believe there is any convention established for the codebase. `eql?` is quite common practise in rails project.

app/models/member.rb
despo: I did initially but then codeclimate was complaining about using -> for multiline lambdas so I switched to the suggested format.

Issue #699, opened 2018-03-20
Donate button does not work on Donate page in Chrome


matyikriszta:
## Description of the issue 📄
The Donate button stopped working on the donation page in Chrome and Firefox (it works fine in Safari). When clicked a popup should appear prompting you to fill out email and card details. We use Stripe to accept payments on the site.

## Screenshots 📷
<img width="1280" alt="screen shot 2018-03-20 at 10 15 06" src="https://user-images.githubusercontent.com/5873816/37670868-9eb9e6d8-2c27-11e8-89bf-43b8cd7c85bb.png">

<img width="1237" alt="screen shot 2018-03-20 at 10 14 35" src="https://user-images.githubusercontent.com/5873816/37670879-a3fc4082-2c27-11e8-9b45-e8391f3d195e.png">

## Steps to fix 🛠
This is likely an issue with the javascript Stripe provided that power the Donate button and the popup where you have to fill in your details. 

## To do 📋
* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request, make sure you mention #codebar24PR2017 in the description
* [x] Mention this issue in the PR description by including it's number
* [x] Have your pull request reviewed & merged by a codebar team member

BPScott:
I've done some research into this, and found [this here thing](https://wordpress.org/support/topic/stripe-payment-is-not-working-on-google-chrome/) that suggests CloudFlare's ["Rocket Loader"](https://support.cloudflare.com/hc/en-us/articles/200168056-What-does-Rocket-Loader-do-)  JS optimisation thing that rewrites and rehosts all JS does not play nicely with Stripe's embed JS. This is backed up by a note in the [Stripe API](https://stripe.com/docs/checkout#integration-custom) that says:

> Checkout must be loaded directly from https://checkout.stripe.com/checkout.js. Using a local copy of Checkout is unsupported, and may result in user-visible errors.

[...]

PR #734, opened 2018-05-22
Fix donate button not working in Chrome

matyikriszta:
## Description
This PR attempts to address the donate button not working in Chrome by loading the Stripe checkout script at the bottom of the page and adding `cfasync="false"` attribute to it.

## Status
Ready for Review

## Related Github issues
#699

### DIFF
--- a/app/views/donations/new.html.haml
+++ b/app/views/donations/new.html.haml
@@ -1,11 +1,8 @@
-- content_for :head do
-  %script{:type => "text/javascript", :src => "https://checkout.stripe.com/checkout.js"}
-
+- content_for :page_footer do
   :javascript
     var stripePublishableKey = "#{Rails.configuration.stripe[:publishable_key]}"
-
-
   = javascript_include_tag 'donations'
+  %script{ :src => 'https://checkout.stripe.com/checkout.js', :type => 'text/javascript', 'data-cfasync' => 'false' }
 
 .stripe.reverse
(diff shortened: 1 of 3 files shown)


app/views/donations/new.html.haml
```diff
@@ -1,11 +1,8 @@
-- content_for :head do
-  %script{:type => "text/javascript", :src => "https://checkout.stripe.com/checkout.js"}
-
+- content_for :page_footer do
   :javascript
     var stripePublishableKey = "#{Rails.configuration.stripe[:publishable_key]}"
-
-
   = javascript_include_tag 'donations'
+  %script{ :src => 'https://checkout.stripe.com/checkout.js', :type => 'text/javascript', 'data-cfasync' => 'false' }
```
BPScott: How does this render as the HTML?

IIRC the rocket loader stuff was very explicit about the data-cfasync attribute being before the src attribute. It seems like a weird request but I guess because rocket loader does some kind of naive string parsing rather than understanding the html as html for the sake of performance.

matyikriszta: it render's with the `data-cfasync` attribute before `src`, like this -> `<script data-cfasync="false" src="https://checkout.stripe.com/checkout.js" type="text/javascript"></script>`.

matyikriszta:
oh no, the flaky test attacks again :(

PR #737, opened 2018-06-03
This will have consistency between attending and attending_reminder

EGiataganas:
## Description of the issue 📄

The section with organisers name and mobile is not presented the same on attending and attending_reminder email.

## Screenshots 📷

**OLD**

**NEW**

## Steps to fix 🛠

The only difference is that `organiser.full_name` and `organiser.mobile` is under a paragraph tag. It can be moved to partial but not sure if it worth doing it


despo:
@EGiataganas thank you 👍

PR #740, opened 2018-06-12
Etag events

dansteele:
Etags allows the server to communicate to the browser if the page is unchanged.
If this is the case, it will tell the browser just that; Rails will not continue processing the action.

In this case, the events endpoint is quite slow. This mitigates that problem on subsequent page views.

([I wrote blog post](https://eola.co.uk/blog/how-to-make-your-site-faster) quite recently on some speed tricks aimed at more junior devs, if anyone's interested)

----------

This PR also adds public/uploads to the .gitignore, preventing accidental commits of local images

----------

Also, another tiny change - getting rspec to save the results of the last run allows a faster test cycle with `rspec --only-failures`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ TAGS
 
 # Local ruby version
 .ruby-version
+
+# Ignore local uploads
+public/uploads/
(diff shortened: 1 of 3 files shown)


dansteele:
Coveralls seems to think an untouched rake task has decreased in coverage ¯\_(ツ)_/¯

despo:
@dansteele thanks for the PR!

We get that all the time with the rake tasks. Probably something we should look into, I'm guessing related with our coveralls configuration.

PR #741, opened 2018-06-12
Fix events endpoint trying to show an inactive chapter

dansteele:


dansteele:
Change made to spec due to it sometimes returning a false negative (sorry for that phrasing :P)

screen shot 2018-06-12 at 15 20 46

PR #744, opened 2018-07-24
Force SSL on app

koddsson:
>Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.

This makes sure that codebar.io traffic goes over HTTPS rather then HTTP. Since [Chrome 68 will mark websites over HTTP as unsafe](https://security.googleblog.com/2018/02/a-secure-web-is-here-to-stay.html) I figured this would be a good thing to get done 😸


koddsson:
This seems to be done on production already at this point.

PR #747, opened 2018-07-27
Update links to slack invitation page

despo:


despo:
@indykid agreed, hopefuly we'll get to that  after this is in :)

PR #748, opened 2018-07-28
feedback improvements & view for chapter organisers

despo:
- exposes chapter specific feedback page for organisers
- removes workshop selection from feedback page as workshop is already known and could lead to mapping against wrong events or chapters
- extracts feedback text to locale


app/views/layouts/_organiser_menu.html.haml
```diff
@@ -15,3 +15,4 @@
     %label My Chapters
   - current_user.organised_chapters.each do |chapter|
     %li= link_to chapter.name, admin_chapter_path(chapter)
+    %li= link_to 'Feedback', admin_chapter_feedbacks_path(chapter)
```
matyikriszta: This is a minor issue but I'm wondering if it would be better to move this link into the organiser chapter view - if a person organises more than 1 chapters these links might become confusing (e.g. there would be two Feedback links). Or add the chapter name to the link text.

despo:
@matyikriszta I did considered this, but I think there is value in organisers being able to see feedback across all chapters, as it is at times food for thought, and also a way to enable them to make improvements by being aware of issues that other chapters may encounter.

PR #769, opened 2018-08-06
(fix) remove default_scope from Chapter

despo:
This should resolve most rollbar issues we’ve been encountering, most relating to past workshops allocated to inactive chapters causing application errors


despo:
@matyikriszta I don't know if we do use any html in chapter intros so that's why I wasnt keen in changing, I'll address in a different PR.

PR #776, opened 2018-08-14
Managin chapter organisers interface

matyikriszta:
## Description
This PR exposes an interface for managing chapter organisers, something we so far did on the console. This view is currently only available to admins. The potential organiser has to be subscribed to the chapter they are becoming an organiser of.

## Status
Ready for Review

## Screenshots
screen shot 2018-08-14 at 18 11 15


matyikriszta:
@despo am I ok to merge with the remaining CodeFactor issues?

despo:
@matyikriszta yes it's ok. Some are non issues either, need to figure out how to disable those checks.

I enabled it as part of trying out Github apps and what value they may add to the project. I'm not convinced yet we need it, but it has some additional checks, besides rubocop, that may come handy.

PR #813, opened 2018-09-05
L10n configure translations app

despo:
includes translations files added by Crowdin.


config/locales/fr.yml
```diff
@@ -230,17 +216,66 @@ fr:
...
+  feedback_form:
+    title: 'Workshop feedback'
+    description: Your submission will be completely anonymous. We read every piece of feedback and use it to help us improve our workshops.
+    note: Please do not mention any names, but instead choose your coach from the dropdown. This enables us to address issues but also to anonymously share feedback with coaches. Member privacy is very important to us!
+    rating: Your overall rating of the workshop
+    coach: 'Who was your coach?'
+    select_coach: '- Select Coach -'
+    tutorial: 'What tutorial did you go through?'
+    select_tutorial: '- Select Tutorial -'
+    request: 'How did you find the workshop?'
```
KimberleyCook: I feel we need to change the wording of this question as I often I see people reply with the word Google 😂

despo: Maybe we can approach that in a different PR as the intention of this is just to get in the translation files for all languages.

PR #830, opened 2018-10-10
Add instance variable @skill to SkillsController

danielthepope:
Instance variable `@skill` is used when rendering a /skills page e.g. https://codebar.io/skills/ruby

As of writing this PR, the header for that page says 

> Coaches skilled in ''

image

This PR adds the `@skill` instance variable to `app/controllers/skills_controller.rb` which is referenced in `app/views/skills/show.html.haml`. The title now says

> Coaches skilled in 'ruby'

I added an expectation to one of the tests. Should I check for a full sentence like that? Checking for content 'ruby' matches the list of coaches below...

### DIFF
--- a/app/controllers/skills_controller.rb
+++ b/app/controllers/skills_controller.rb
@@ -1,5 +1,6 @@
 class SkillsController < ApplicationController
   def show
+    @skill = skill
     @coaches = Member.includes(:skills)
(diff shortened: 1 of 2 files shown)


despo:
:shipit:

danielthepope:
🎉 
Thanks @despo 😄

Issue #627, opened 2017-11-24
Ability for admin to subscribe any member of a chapter to a workshop


matyikriszta:
## Description of the issue 📄

On the admin workshop view (admin/workshops/:id) admins have access to a dropdown that allows them to manually subscribe members of a chapter to any given workshop. Currently this dropdown is populated with members who have received but not accepted an invitation email. This is causing issues for the London chapter as we have stopped sending email invites for that chapter. Update this dropdown to show all members of a chapter who have not yet RSVP'd to the workshop.

## Screenshots 📷
<img width="571" alt="screen shot 2017-11-24 at 21 11 50" src="https://user-images.githubusercontent.com/5873816/33223901-244b5f88-d15c-11e7-8e95-c192859ea676.png">
Dropdown to update

## Steps to fix 🛠

Update the behaviour of the dropdown in the `app/views/admin/workshops/_invitation_management.html.haml` partial.

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request, make sure you mention #codebar24PR2017 in the description
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

matyikriszta:
This issue is still relevant, removing the stale label.

matyikriszta:
@jonodrew yes, this is still relevant. Can you point me to the line of code where you see the dropdown being commented out on master? Here's a screenshot of a Brighton workshop (currently live on the site) where you can see the dropdown (so it's still being used in it's current form).

<img width="1467" height="757" alt="Image" src="https://github.com/user-attachments/assets/44c9f5cb-9547-4d4a-a797-e7cb3d2145c5" />

The issue with the way the dropdown currently works is that if a member subscribes to a chapter after the invitations have been sent out for a workshop, you won't be able to find them in the dropdown. The dropdown loads all members who are subscribed to a chapter and have received an invitation to the workshop.

[...]

PR #840, opened 2018-10-13
Manually add any member of a chapter to a workshop

matyikriszta:
## Description
This PR updates the Select a member to RSVP dropdown on the admin workshop page to include all members of a chapter regardless whether they have received an invitation or not. 

## Status
Work in Progress

## Related Github issue
#627


matyikriszta:
This PR can probably be salvaged, removing the stale label for now.

Issue #849, opened 2018-10-17
Add description field to Chapter to be displayed on individual chapter pages


matyikriszta:
## Description of the issue 📄
Add description field to Chapter model to allow organisers to display a short description of their chapter on the individual chapter pages.

## Screenshots 📷
<img width="1440" alt="screen shot 2018-10-17 at 1 28 14 pm" src="https://user-images.githubusercontent.com/5873816/47114387-8bb4ed00-d210-11e8-879b-bc4941482a31.png">

## Steps to implement 🛠
Create a migration to add description field to the chapter model

## To do 📋
* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number
* [x] Have your pull request reviewed & merged by a codebar team member

PR #853, opened 2018-10-19
Add description field to chapter

matyikriszta:
## Description
This PR adds a description field to the Chapter model so chapter organizers can provide custom intro text for their chapters to be displayed on the individual chapter pages. Character count for the field is capped at 280. If there is no description field specified, a default welcome message is shown. This PR also extracts the current intro text into two strings: intro (can be customized on a chapter basis) and info (same for every chapter) and moves the info text (FAQ and chapter email) into it's own visually highlighted section with the CTA buttons (Sign in etc.).

## Status
Ready for Review

## Related Github ticket
#849 

## Screenshots
Chapter page with custom intro (logged out)
screen shot 2018-10-18 at 8 32 22 pm

Chapter page with custom intro (logged in)
screen shot 2018-10-18 at 8 33 58 pm

Chapter page with default intro (logged out)
screen shot 2018-10-18 at 8 34 47 pm

### DIFF
--- a/app/views/chapter/show.html.haml
+++ b/app/views/chapter/show.html.haml
@@ -7,12 +7,11 @@
       - if @chapter.description
         %p.lead= @chapter.description
       - else
-        %p.lead= t('chapters.intro').html_safe
-
+        %p.lead= t('chapters.intro_html')
 .panel
   .row
     .small-12.medium-6.columns
-      = t('chapters.info', email: @chapter.email).html_safe
+      = t('chapters.info', email: h(@chapter.email)).html_safe
     .small-12.medium-6.columns.text-center
       - if logged_in?
         = render partial: 'subscriptions'
--- a/app/views/chapter/show.html.haml
+++ b/app/views/chapter/show.html.haml
@@ -2,14 +2,22 @@
   .row
     .large-12.columns
       %h1= @chapter.name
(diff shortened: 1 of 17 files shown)


app/views/chapter/show.html.haml
```diff
@@ -2,14 +2,22 @@
...
     .large-12.columns
       %h1= @chapter.name
-      %p.lead
-        = t('chapters.intro', email: @chapter.email).html_safe
+  .row
+    .large-7.columns
+      - if @chapter.description
+        %p.lead= @chapter.description
+      - else
+        %p.lead= t('chapters.intro').html_safe
```
despo: @matyikriszta  there is a trick you can do with the locale to resolve the codeclimate warning. When a locale's name is suffixed with `_html` then you don't need to use `html_safe` in the views.

e.g. in this instance it should be changed to `chapters.intro_html`.

matyikriszta: @despo cool, I'll give that a try

matyikriszta: @despo unfortunately that trick doesn't solve the issue where it's actually reported: `t('chapters.info', @chapter.email).html_safe`, but I fixed it by manually escaping the model attribute with `h()`.

PR #854, opened 2018-10-19
Fix Seeding of coaches and students and Counter of coaches

MarckK:


despo:
Thank you @MarckK - deployed! :shipit:

Issue #848, opened 2018-10-17
Fix broken organisers grid on chapter page


matyikriszta:
## Description of the issue 📄
The chapter bridge organisers grid currently does not support variable height elements. Fix the grid to appear uniform even when the items are of varying heights.

## Screenshots 📷
<img width="1279" alt="screen shot 2018-10-16 at 21 59 51" src="https://user-images.githubusercontent.com/5873816/47063427-c1fe5800-d18f-11e8-8604-838abea57d12.png">

## To do 📋
* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number
* [x] Have your pull request reviewed & merged by a codebar team member

PR #862, opened 2018-10-24
Fix and improve appearance of chapter page organisers grid

matyikriszta:
## Description
This PR refactors the chapter page organisers grid to fix items breaking out of the grid because of differing heights.

## Status
Ready for Review

## Related Github ticket
#848 

## Screenshots
screen shot 2018-10-23 at 9 11 14 pm

### DIFF
--- a/app/assets/stylesheets/partials/_organisers.scss
+++ b/app/assets/stylesheets/partials/_organisers.scss
@@ -1,3 +1,32 @@
 .user-link {
   border-bottom: none;
 }
+
+.organisers-grid {
+  display: flex;
+  flex-wrap: wrap;
+}
+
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/partials/_organisers.scss
```diff
@@ -1,3 +1,32 @@
...
+  flex-wrap: wrap;
+}
+
+.organiser-item {
+  display: flex;
+  margin-bottom: 2rem;
+}
+
+.organiser-content {
+  display: flex;
```
KimberleyCook: I don't think you need `display: flex;` on chid elements, only the parent one.

Issue #846, opened 2018-10-17
Order chapters alphabetically on admin dashboard page


matyikriszta:
## Description of the issue 📄
Chapters in the Chapter section of the admin dashboard page are currently ordered by [some order]. Order chapters alphabetically so it's easier to find them.

## Screenshots 📷
<img width="467" alt="screen shot 2018-10-16 at 21 53 58" src="https://user-images.githubusercontent.com/5873816/47062981-1274b600-d18e-11e8-83a9-09d32187f524.png">

## To do 📋
* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

shanodin:
Hello, I'd like to look into fixing this issue!

despo:
@shanodin thank you, that would be great! Let me know if you need any help.

shanodin:
@despo oops I deleted my comment because I wasn't sure if I could do it. The issue I am having is that I don't know how to create an admin account to see the page the issue is referencing so I can't see the effect of my changes. If you could tell me how to do that I might make some progress.

shanodin:
Oh it looks like someone else has submitted a PR for this :( nevermind then

pawgrmr:
Hello, is this still open? I noticed someone has submitted a PR for this issue but it's still open

matyikriszta:
@pawgrmr feel free to pick this up, I have closed the previous PR as the work is not what we have asked for and there hasn't been much progress on it

PR #864, opened 2018-10-28
Order chapters alphabetically on admin dashboard page

pain11:
#846


app/controllers/admin/events_controller.rb
```diff
@@ -27,6 +27,7 @@ def show
     @attending_students = InvitationPresenter.decorate_collection(@original_event.attending_students)
     @attending_coaches = InvitationPresenter.decorate_collection(@original_event.attending_coaches)
     @host_address = AddressPresenter.new(@event.venue.address)
+    @chapters = @event.chapters.asc(:name)
```
despo: This breaks the app. I think chapters need to be retrieved from `@original_event`

matyikriszta:
@pain11 thanks for your work on this so far. I can see you are trying to alphabetise the chapters on the admin event page but that is not what the issue that was raise was about. I meant the list of chapters on the admin dashboard page which is the index action in this controller: https://github.com/codebar/planner/blob/master/app/controllers/admin/portal_controller.rb. Could you please update your PR to reflect this?

matyikriszta:
@pain11 thank you for your work but since there hasn't been any activity on this PR and the work you've done is not the work we have asked for I'm going to close this PR.

PR #889, opened 2018-12-01
Spelt GitHub in the correct format

rubyrox:
Changed the small h in GitHub to a capital. As it is meant to be. 
CC @joenash :octocat:


koddsson:
Awesome first PR @rubyrox 💖

PR #895, opened 2018-12-12
Delete skipped tests

koddsson:
Hey hey 👋! 

Seems like these tests either were skipped when commited initially or have been skipped since 2015 so I figured we can delete 'em?

### DIFF
--- a/spec/features/admin/manage_workshop_attendances_spec.rb
+++ b/spec/features/admin/manage_workshop_attendances_spec.rb
@@ -38,16 +38,5 @@ feature 'managing workshop attendances' do
 
       expect(page).to have_content('2 are attending as students')
     end
-
-    xscenario 'can rsvp an invited student the the workshop' do
-      other_invitation = Fabricate(:workshop_invitation, workshop: workshop, attending: nil)
-
-      visit admin_workshop_path(workshop)
-      expect(page).to have_content('1 are attending as students')
-
-      select "#{other_invitation.member.full_name} (#{other_invitation.role})", from: 'workshop_invitations'
-
-      expect(page).to have_content('2 are attending as students')
-    end
   end
 end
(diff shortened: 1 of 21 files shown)


spec/features/admin/manage_workshop_attendances_spec.rb
```diff
@@ -38,16 +38,5 @@
...
-    xscenario 'can rsvp an invited student the the workshop' do
-      other_invitation = Fabricate(:workshop_invitation, workshop: workshop, attending: nil)
-
-      visit admin_workshop_path(workshop)
-      expect(page).to have_content('1 are attending as students')
-
-      select "#{other_invitation.member.full_name} (#{other_invitation.role})", from: 'workshop_invitations'
-
-      expect(page).to have_content('2 are attending as students')
-    end
```
despo: Hm. I wonder why this was commented out and if we should maybe try and fix this spec rather than removing it.

spike01:
I've started looking at this as skipped tests also bother me. The workshop attendances scenario test doesn't work because JS is not enabled on the test - however, when it is enabled, the current PhantomJS/Poltergeist is not set up correctly and so still fails. The scenario does work as expected when tested manually.

I don't think is worth fixing as PhantomJS is effectively abandoned. I am preparing a PR stripping out PhantomJS/Poltergeist as a first step (no other tests currently use it). I will look into setting up headless Chrome so that a.) this test will pass and b.) future contributors needing JS enabled for feature tests will be able to use it.

despo:
Thanks for the feedback @spike01. Given that, happy to approve the PR 👍 @koddsson

spike01:
Whoops, reading back through this, I think I fixed these tests in #941 (which has been merged) as a separate issue to the JS tests. That wasn't clear from the discussion in this thread.

So not sure if this should be closed rather than merged. Up to maintainers 👍

koddsson:
@spike01 I rebased this of master where I think one test is still being skipped but has pulled in your changes so this patch doesn't delete them anymore.

koddsson:
This PR is now at 0 changed files from changes to `master`.

PR #899, opened 2018-12-19
Give example of how to run a single test file

koddsson:
The README doesn't include info on how to run a single test. This patch adds that info below the info on how to run the whole test suite.

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,11 @@ bundle exec rails server
 bundle exec rake
 ```
 
+If you just want to run a single test file you can pass the path to the file via the `SPEC` variable:
+```bash
+bundle exec rake SPEC=<path to test>
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -131,6 +131,11 @@ bundle exec rails server
 bundle exec rake
 ```
 
+If you just want to run a single test file you can pass the path to the file via the `SPEC` variable:
+```bash
+bundle exec rake SPEC=<path to test>
```
despo: This looks good.  
However, can we also add an example using rspec, which I think is a more common way to know and run individual specs.

matyikriszta:
@koddsson @spike01 have addressed this in #946 which just got merged in, therefore I'm closing this issue. Thanks for suggesting this change though.

PR #941, opened 2019-03-05
Fix skipped mailing list test

spike01:
These tests were skipped because there of a guard clause preventing
mailing list subscribing/unsubscribing unless in production. Since
neither tests nor code have been touched in about a year I think this
is an acceptable workaround (I really dislike pending tests!).

A further explanation of why this works can be found here:
https://stackoverflow.com/a/25134591


spike01:
This partially addresses #895, working on the other skipped test

PR #943, opened 2019-03-06
Updates Gemfile.lock after poltergeist gem has been removed from the Gemfile

matyikriszta:
## Description
After merging in #942 auto deployment to codebar-staging failed because the updated Gemfile.lock wasn't committed after removing the poltergeist app from the Gemfile. Per the error message from the build log:

```
You are trying to install in deployment mode after changing
your Gemfile. Run `bundle install` elsewhere and add the
updated Gemfile.lock to version control.

You have deleted from the Gemfile:
* poltergeist
```

This PR updates and commits the new Gemfile.lock.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -81,7 +81,6 @@ GEM
       railties (>= 3.0)
       sass-rails (>= 3.2)
     chunky_png (1.3.10)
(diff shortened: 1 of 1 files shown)


spike01:
D'oh, I forgot how to Ruby

PR #946, opened 2019-03-08
Add support for JS feature tests w/Chrome headless

spike01:
I previously removed Poltergeist (as it is no longer being maintained and wasn't configured in a way that worked) and said I would replace it in a future PR - this is that PR.

At the moment there is only one test which uses `js: true`, which after being skipped for a few years is now working. There is another PR (#945) that this will unblock.

However, Selenium tests are fragile, not as intuitive to write as Rack-based Capybara tests and slow - so I would definitely advocate for trying to keep the number of tests to a minimum, and eventually phasing them out completely if possible.


spike01:
If this is merged, #895 can be closed as well, as well as #899 (I've added those examples as well as others for JS tests to the README) - with thanks to @koddsson for raising both PRs that started me thinking about doing this.

Dockerfile
```diff
@@ -5,6 +5,14 @@ FROM ruby:2.4.2
 RUN curl -sL https://deb.nodesource.com/setup_8.x | bash -
 RUN apt-get install -y build-essential libpq-dev nodejs
 
+# Install latest chrome dev package
+RUN set -ex; \
```
koddsson: The documentation below seems to suggest that running the tests on a headless chrome doesn't work against a docker run instance. Do we need the chrome install if that is the case?

spec/features/admin/manage_workshop_attendances_spec.rb
```diff
@@ -39,13 +39,20 @@
...
+    scenario 'can rsvp an invited student to the workshop', js: true do
+      login_as_admin(member)
+
       other_invitation = Fabricate(:workshop_invitation, workshop: workshop, attending: nil)
 
       visit admin_workshop_path(workshop)
       expect(page).to have_content('1 are attending as students')
 
-      select "#{other_invitation.member.full_name} (#{other_invitation.role})", from: 'workshop_invitations'
+      # Unclear why this has to be done twice, when tested manually it works
```
koddsson: If this gets merged it would be great to open up a issue to debug and fix this.

Dockerfile
spike01: I'll make the documentation clearer - headless Chrome is all working fine in Docker (and necessary to run tests in Docker correctly), but not running with a visible (headful?) browser

koddsson: >headful

😂

Issue #867, opened 2018-10-31
Prevent accidental removal of RSVP with a confirm dialogue 


khamiltonuk:
## Description of the issue 📄
As an admin it is a little too easy to remove an RSVP by mistake.
I propose we add a confirm dialogue; just to make sure.

Dialogue should explain the attendee is removed from the workshop, but not blocked, the student can easily sign up again, I think the functionality is fine, but could be explained somewhere. 

### Helpful guide 
 [https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm)

## Screenshots 📷
<img width="1426" alt="screen shot 2018-10-31 at 09 36 14" src="https://user-images.githubusercontent.com/4013283/47775710-ba42c580-dcf0-11e8-9d0d-9485efeaf4c3.png">

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

matyikriszta:
@khamiltonuk these are great suggestion but I feel like these are separate issues and should be addressed as separate PRs. Would you mind breaking these into separate issues accordingly? For the icon bit, do you have an idea what icon would be better? Maybe a trash can?

Re: the email, I think it would be nice to send an email. This again could be it's own item though.

khamiltonuk:
@matyikriszta Ok updated the issue just to be about the confirm dialogue

matyikriszta:
@khamiltonuk great, thanks

matyikriszta:
@khamiltonuk there's a PR for this now at #948

khamiltonuk:
Done

PR #948, opened 2019-03-13
Add confirm dialog to cancel attendance button

matyikriszta:
## Description
This PR adds a confirm dialog to the cancel attendance button to help prevent accidental attendance removal.

## Related Github issue
#867 

## Screenshots
Screen Shot 2019-03-12 at 10 30 30 PM


app/views/admin/workshop/_attendances.html.haml
```diff
@@ -11,7 +11,7 @@
               - else
                 %i.fa.fa-check-square-o
             - else
-              = link_to admin_workshop_invitation_path(@workshop, invitation, attending: false), class: 'cancel_attendance', title: 'Cancel RSVP', method: :put, remote: true do
+              = link_to admin_workshop_invitation_path(@workshop, invitation, attending: false), class: 'cancel_attendance', title: 'Cancel RSVP', data: { confirm: "Are you sure you want to remove #{invitation.member.full_name} from the list of attendees?" }, method: :put, remote: true do
                 %i.fa.fa-minus-square-o
```
KimberleyCook: Shouldn't this string come from the i18n files?

matyikriszta: Yes if we wanted to translate this page. Currently none of the other strings on this page are coming from i18n. I guess the question is do we want to translate this page/is it priority?

despo: I think for consistency we should aim to have all the newly introduced text in the locale files, this should also help us translate easier moving forward. 

However I'm happy to approve the PR and we can address that seperately.

matyikriszta: @despo, sounds good, I'll merge this in and start a PR moving the strings into a i18n. Or we can wait to do that until the need to translate this page arises.

PR #957, opened 2019-03-28
Explicitly cast year to string in tests

koddsson:
When running the tests there is some output and one of those output is a deprecation warning from Capybara. This PR explicitly casts those variables to string.

The command I ran to produce the following output is:

```sh
bin/drake SPEC=spec/features/listing_coaches_spec.rb:25
```

### Before

```
Locator 2018 must be an instance of String or Symbol. This will raise an error in a future version of Capybara.
Locator 2016 must be an instance of String or Symbol. This will raise an error in a future version of Capybara.
.

Finished in 3.85 seconds (files took 5.04 seconds to load)
1 example, 0 failures
```

### After 

```
.

Finished in 3.98 seconds (files took 4.49 seconds to load)
1 example, 0 failures
```


matyikriszta:
hey @koddsson, could you be able to rebase this PR against master so we can merge it in?

koddsson:
@matyikriszta Should be good now ✨

matyikriszta:
@koddsson great, thanks for the PR :)

PR #958, opened 2019-03-28
Update rails to version 5.1.7 and ruby to version 2.5.5

koddsson:
I could not build the image from 2.4 since the node sources are not available. I think they have been moved and so the sources link will 404 now. Instead of finding the correct source link in the archives I figured it was a good excuse to bring us to ruby 2.5?


koddsson:
If I'm reading this error message right, we might need to update rails before we can update ruby.

koddsson:
Haven't actually manually tested this so I'm gonna do that and fix those codeclimate warnings. Not sure how I can fix the snyk warnings since I don't get access to that.

koddsson:
This is getting big and hard to manage. I think I might break it out into smaller PRs as much as I can to make this easier.

koddsson:
Actually re-reading I'm not sure if this can broken up any more. I thought I had jumped more versions but this is just one major version.

I'll work on cleaning up the non-test errors.

koddsson:
Not sure why these things are failing. We already have nokogiri on `master` and there isn't a update for it. Also those two codeclimate issues seem to be unrelated to what I changed in this PR.

I think this is ready for review.

notapatch:
Hello koddson

I'm not a maintainer and have no official capacity - just a website user who would like to see it, and the data, on a supported version of Rails. 

Are you still interested in upgrading? As you said yourself there's quite a large number of changes. How about putting some changes through which aren't Rails 5 related?  I think it would be sensible to get as much through as possible before hitting Rails 5.  By eye I would estimate the following have a good chance:

```
s/before_filter/before_action
fix spec deprecation warnings
s/uniq/distinct * 2
moved common code out of conditional
make method be less then 10 lines 
reduce methods sizes by moving DB fetching into `before_action`
fix syntax error
fix order of associations

the show method doesn't exist
Revert "the show method doesn't exist”
define a empty show method to shut up linter

```

Also wondering about these but ...
```
render plain not text  -  this one sounds interesting but may have to wait for 5?
dont raise on n+1 in tests - I was wondering why this was Rails 5 specific ? 
cast year to string in test -  Is this rails 5?
```

re: `ignore rubocop rule`
I also think any rubocop rules should go into .rubocop.yml and not into the codebase

If you aren't interested I can put as much of this as I can through.

koddsson:
@notapatch I don't have any capacity to work on this. I'll close the PR but feel free to cherry-pick if you think that any of my commits are helpful 😸

notapatch:
I quite understand. I will pick away, and thank you for doing the work!

Issue #964, opened 2019-04-01
Add note about Google Search optimisation on new job page


despo:
## Description of the issue 📄

Add note on new job page, about sharing jobs with Google search when address information is filled in.

**Text to add:**
> The information below only required if you want this job post to be shared with Google Search UK.

We are also open to suggestions about how this text can be improved.

## Screenshots 📷
<img width="1231" alt="new-job" src="https://user-images.githubusercontent.com/159200/55354821-944b6800-54be-11e9-9d8f-e7cec5170f66.png">

## Steps to fix 🛠

1. Add text to locale file (there is already a members/job/new section)
https://github.com/codebar/planner/blob/master/config/locales/en.yml#L337

2. Refrence locale entry from view
https://github.com/codebar/planner/blob/master/app/views/member/jobs/_form.html.haml#L37
 
3. Verify with a test

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

klmansel:
I would like to claim this issue.

KimberleyCook:
@klmansel yes please do :D

klmansel:
Done :) PR submitted @KimberleyCook

PR #995, opened 2019-06-08
Added text to new job form

klmansel:
Added provided text to new job form explaining it is only required if user wants job to be shared with Google Search UK.

I was unsure of the required text size, so I went with small font.

Fixes Issue #964 

Screen Shot 2019-06-08 at 1 20 56 PM

### DIFF
--- a/app/views/member/jobs/_form.html.haml
+++ b/app/views/member/jobs/_form.html.haml
@@ -34,6 +34,10 @@
   .row
     .large-6.columns
       = f.input :location, required: true
+
+  %small
+    = t('member.jobs.new.optional_details_message')
+
   .row
     .large-6.columns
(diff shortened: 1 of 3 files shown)


klmansel:
I think you may have to @KimberleyCook - Merging is blocked for me.

matyikriszta:
hey @klmansel this looks awesome, thanks for the PR. I was wondering though, could we wrap that `%small` tag into a `%p` so there's more spacing between the text and the label for the Address input?

klmansel:
@matyikriszta does this look better? 
Screen Shot 2019-06-08 at 8 29 02 PM

matyikriszta:
@klmansel yep, looks good, thank you so much for contributing to the project.

PR #1035, opened 2019-09-05
Show total number of spots in workshop invitation

legoscia:
I thought it would be nice for registered attendees to see the total number of spots available at the workshop, to give an idea of how full the workshop is and whether it's worth trying to invite more people.

What do you think?


matyikriszta:
@legoscia I think this is a great idea, could you post a screenshot into this thread showing us how this looks like in the site?

legoscia:
It looks like this:

image

notapatch:
Hello @legoscia 

Yeah, that has been fixed. You need to rebase your change on top of the latest code.
When you've done it [this file which is causing the problem ](https://github.com/legoscia/planner/blob/show-number-of-spots/.travis.yml).

 [Will look like this one](https://github.com/codebar/planner/blob/master/.travis.yml).... 

Key line is: `dist: xenial`

legoscia:
Seems to have worked. Thanks @notapatch !

PR #1039, opened 2019-09-10
Fix: flaky test breaks when fabricator produces apostrophe name

notapatch:
- Nokogriri::CSS::SyntaxError: expect(page).to
    have_css("span[title='#{permissions.members.last.full_name}']")
  - the programmer needed to use two sets of quotes. Double quotes
    for the have_css string and single quotes for the nested CSS
    attribute for title's value.
  - When the fabricator produced a standard name like smith
    or Jones it worked - about 9 times in 10.
  - When the fabricator produced a name with an apostrophe in it
    it broke the test, about 1 in 10 times, because the CSS
    attribute was: title='O'Conner' and the string became
    unbalanced

### DIFF
--- a/spec/features/admin/meeting_spec.rb
+++ b/spec/features/admin/meeting_spec.rb
@@ -58,8 +58,8 @@ feature 'Managing meetings' do
       click_on 'Update'
 
       expect(page).to have_content('You have succesfully updated the details of this meeting')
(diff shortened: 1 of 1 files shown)


notapatch:
[If you look through the past builds on Travis you can see it pop up](https://travis-ci.org/codebar/planner/builds/564847119)

docljn:
I have also seen this - the apostrophe is handled correctly in the ui but not in test.

PR #1044, opened 2019-09-17
spelling mistakes esp successfully

notapatch:
* mainly effects the tests but also user messages and out to the
    translations
  * Note: submitted has been spelt wrongly but it also has a
    database table attribute spelt as 'submited' .... so I am
    leaving well alone in this commit
  * there are other spelling mistakes to review

### Update
Clearly, changing spelling in a project doesn't make it more vulnerable (see: security/snyk). So I'll stop committing until resolved.


KimberleyCook:
Thanks for doing this PR @notapatch and nice spot :)

PR #1045, opened 2019-09-17
Make new job salary field mandatory

KimberleyCook:
I think the salary field in the new job form should be mandatory.


matyikriszta:
@KimberleyCook as an aside I really hate the current form validation - it just doesn't work very well. For example it's kind of hard to tell which form fields are being validated, and I wish it showed all errors all at once as opposed to error to error. Not sure we even need the "fancy" javascript validations, just show errors inline on top of the inputs or all at once at the top of the page. What do you think?

KimberleyCook:
@matyikriszta Yeah I agree, I think showing some text above the input field that has errored is what it should do. I hate it when forms show it at the top (or bottom) and I have to scroll and figure out what input field is erroring. 

I'll create a new issue for this.

PR #1047, opened 2019-09-18
gitignore byebug_history

notapatch:
Now in by default in Rails - and very useful when debugging


KimberleyCook:
Thanks @notapatch

PR #1048, opened 2019-09-19
deprecation fix: paragraph faker uses keyword argument

notapatch:


notapatch:
When you submit a PR you have been working with the code for an hour or two and everything is obvious to you. This is _exactly_  the wrong time to have to write about the PR. In an ideal world I would write a PR and then review in two days later when I've forgotten most of it and have to explain it to myself all over again. I've re-explained a couple of the PRs to help the reviewer or, indeed, myself in two days time.

PR #1049, opened 2019-09-19
fix warning: 'checking for expected text of nil is confusing'

notapatch:
When you run the tests you get this warning `checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead.`

Screenshot 2019-09-22 at 10 18 57

### This is the fix

  - Two steps required to fix issue
    1) The course fabricator was not producing a URL
    2) The test was testing for text and not a URL

  - 1) Course Fabricator should have generated a URL
    - to help the reader understand what attributes the Fabricator
      had and what it should have had I moved the Fabricators
      attributes to the database table's schema order. This is to
      avoid mistakes in the future.
    - Finally, I added the missing url faker

  - 2) expectation testing for have_content of URL when it
    should have been testing for have_link

### DIFF
--- a/spec/fabricators/course_fabricator.rb
+++ b/spec/fabricators/course_fabricator.rb
@@ -1,11 +1,11 @@
 Fabricator(:course) do
-  description { Faker::Lorem.paragraph }
-  short_description { Faker::Lorem.sentence }
   title { Faker::Lorem.sentence }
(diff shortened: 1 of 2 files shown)


notapatch:
rebased!

PR #1050, opened 2019-09-19
Fix: schema synchronised with ExtendedJobModel migration

notapatch:
- schema includes in Job table the attribute/column company_region
 - extend_job_model that introduced these set of changes did *not*
   include company_region
 - no reference to company_region in the code (and would have
   to be added to a new migration anyway)
 - removing company_region
 - referenced commit 85b9db678365b01623df567190f84a5b1cbd8042

--------------
Odd one. The attribute is only in the schema and not in migrations .. when you run the full migrations again after trashing the database it removes it from the schema. 
You are left with a change in your repository.

@despo - might want to check it out?

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -309,7 +309,6 @@ ActiveRecord::Schema.define(version: 20181021014646) do
     t.datetime "updated_at"
     t.string   "company"
     t.integer  "approved_by_id"
-    t.string   "company_region"
     t.string   "company_website"
     t.string   "company_address"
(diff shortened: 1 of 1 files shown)


notapatch:
Hello @matyikriszta,

I was thinking this should go through. The schema remains out of step with the migrations and that is an odd state of affairs. The review would only be useful to @despo if there was ongoing work which doesn't seem to be the case.

notapatch:
rebased

PR #1060, opened 2019-09-23
fix: listable_spec flaky scope test - introduce match_array

notapatch:
- Problem:
    - the tests fabricate three future workshops without specifying an order
    - the test asserts that three future workshops appear in date order
  - Fix: test asserts workshops are present but not the order of appearance

[This broke PR #1056 earlier today on "#upcoming returns a list of all upcoming workshops" but its sibling test could have just as easily broken](https://github.com/codebar/planner/pull/1056)

Screenshot 2019-09-23 at 11 51 04

 note: the bot forced pushed a couple of hours later and passed the second time

### DIFF
--- a/spec/models/concerns/listable_spec.rb
+++ b/spec/models/concerns/listable_spec.rb
@@ -9,7 +9,7 @@ describe Listable do
         Fabricate.times(5, :past_workshop)
         future_workshops = Fabricate.times(3, :workshop)
 
-        expect(Workshop.upcoming).to eq(future_workshops)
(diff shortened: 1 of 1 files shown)


notapatch:
I change the test code and coverage has decreased??

matyikriszta:
@notapatch yes, sometimes it does that and I have no idea why, this really shouldn't decrease coverage in my opinion. I'm still happy to merge it if you are.

notapatch:
Absolutely!

PR #1062, opened 2019-09-28
update_attribute to update

notapatch:
- rails 4 aliased update_attribute to update
 - rails 6 deprecates update_attributes
 - rails 6.1 removes update_attributes
 - https://edgeguides.rubyonrails.org/6_0_release_notes.html#active-record-deprecations

It's been aliased for a _long_ time but only in Rails 6 is it getting removed. The shorter syntax makes the code look neater and it will need to be removed... eventually, anyway.

[Rails 4.2 update/update_attributes code](https://github.com/rails/rails/blob/4-2-stable/activerecord/lib/active_record/persistence.rb#L244-L256)

```ruby
    # Updates the attributes of the model from the passed-in hash and saves the
    # record, all wrapped in a transaction. If the object is invalid, the saving
    # will fail and false will be returned.
    def update(attributes)
      # The following transaction covers any possible database side-effects of the
      # attributes assignment. For example, setting the IDs of a child collection.
      with_transaction_returning_status do
        assign_attributes(attributes)
        save
      end
    end

    alias update_attributes update
```


app/controllers/events_controller.rb
```diff
@@ -51,7 +51,7 @@ def rsvp
     invitation = member.invitations.where(event: @event, role: 'Student').try(:first)
     invitation ||= Invitation.create(event: @event, member: member, role: 'Student')
 
-    invitation.update_attributes attending: true
+    invitation.update attending: true
```
matyikriszta: @notapatch can we add parenthesis to this update call so we are using it consistently across the codebase?

matyikriszta: @notapatch can we add parenthesis to this update call so we are using it consistently across the codebase?

notapatch:
Sounds good. I was going to do a pass with Rubocop at some point but there's no reason not to do it now.

app/controllers/events_controller.rb
notapatch: I've typed the answer in the wrong box. Yes, consistency is always best.

notapatch:
@matyikriszta 

This one you had one problem with which I fixed - a bracket around the call was missing.

PR #1075, opened 2019-10-03
Add year to workshop date on admin workshops page

### LINKED ISSUE #1041: Admin workshop list does not contain year for event date
## Description of the issue 📄

The admin workshop list page (Chapter -> View all workshops) shows the date of the event but this date does not show the year. This makes the list confusing if the chapter has been around for more than a year.

## Screenshots 📷

image

## Steps to fix 🛠

Add the year.

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number

matyikriszta:
## Description
This PR adds the year to workshop dates on the admin/workshops#index page. Also fixes a few Rubocop offenses in `app/helpers/application_helper.rb`.

## Status
Ready for Review

## Related Github issue
Resolves #1041 

## Screenshots
Screenshot 2019-10-02 at 8 52 37 pm


notapatch:
When I saw the original issue I did look over humanize_date and played around with it but I thought "scary method" and did not try to create a PR for the issue. Methods should have one purpose but it is trying to be a date, datetime, different formats within all these demands. Further, it hard codes an order of day and month which Americans, and no doubt others, will not like. 

I think date variations should be handled directly by i18n. 

i18n won't handle [ordinalizing numbers natively until Rails 6.0](https://github.com/rails/rails/pull/32168) - if you decide you don't need an ordinalized number in an admin list of dates then you have workshop format that already covers what you need (it isn't used anywhere at all at the moment and probably should be renamed after the data it is formatting not where it is used - the workshop). So the solution would be [something like this](https://github.com/notapatch/planner/pull/14/commits/82c894a9624e8d920349cf56804a11af48eb5322). 

If we can't wait for Rails 6 and we want ordinalize now a [solution something like this](https://github.com/notapatch/planner/pull/14/commits/24d40e9aea16f9ae4431b1022034d49e83688e2a). I'm not sure how that works with the other en-GB, en-AU formats as I only ever got en?

Other than that, the date formatting and the test is at the wrong level - feature tests are slow and we always display a date. The formatting and testing should be at the presenter level which is fast - and if you want to test across different localisations you could end up with running 5 tests, one for each language.

```ruby
class WorkshopPresenter
  def day_and_date
     I18n.l(model.date_and_time, format: :workshop)
  end
end
```

despo:
@notapatch  as this is something we desperately need and `humanize_date` is already in place I decided to merge this and have created issue 
https://github.com/codebar/planner/issues/1204 quoting your comment, where the focus can be on removing `humanize_date` and moving responsibility to the relevant Presenters - as it's currently used in a number of areas in the app.

notapatch:
That makes sense 👍

PR #1069, opened 2019-09-30
Feature/chapters index page

yihyang:
Issue: https://github.com/codebar/planner/issues/738

Screenshot:

Screenshot 2020-05-17 at 11 31 30 PM

PR #1079, opened 2019-10-04
[WIP] chapter with basic layer of testing

notapatch:
I'll wait on @yihyang pull request #1069  - by eye I haven't seen any merge conflict but I want to keep things easy.

---------------------------------------------------------------
 - using subject(:chapter) where possible
 - trying is_expected.to it can automatically be changed by
   Rubocop
 - validating all attributes as required on model
 - slug is an on save test - explicit save! and reload before
   testing again
 - scope tests only need one object of each type and see if it
   is allowed or denied

 - mystery
   - available_to_user
   - not sure if we need a URL safe slug or not but it isn't
     being enforced

-----------------

And, coverall, like one word in production code removed. I'm hoping for at least +0.001% increase in coverage *fingers crossed*


notapatch:
The refactoring of tests can wait.

PR #1081, opened 2019-10-05
Spec layout spacing issues

notapatch:
There's two problems with Rubocop, first there's 1800 'issues', second there's 100 or so cops. This address about 8 space related cops with 1 or two errors in the thousands of lines _but_ focused on the test code only - no deploys necessary.

Oh, and an empty example group... I'm glad it picked that out as I had looked at the code previously and thought WT Hell.


matyikriszta:
@notapatch I understand Rubocop can be overwhelming but the intention was not to fix all of the offenses to to ensure new code is offense free. Then we can get to current offenses slowly as we work on parts of the app/

matyikriszta:
@notapatch oh no, please feel free to open Rubocop only pull requests, my comment was only meant ease your worry about the number of Rubocop offenses and to point out we do not have to fix them all all at once. But if you enjoy fixing them please feel free to do so (I know I do). BUT if you want to focus on other work that is totally fine as well.

notapatch:
Marvellous

I think it's easier to get "one offence" out the way rather than deal with 5 varied offences with each significant commit.

The good news is, and you won't believe this, I do have a few rubocop pr's left on my local repo but I was "easing them into the system" on the off chance you didn't want to spend your open source programming-time reviewing them.

PR #1082, opened 2019-10-05
Remove respond_to from specs

notapatch:
- removes 40 respond_to tests
    - inconsistent: respond_to covers 40 of 240 possible attributes
    - adds nothing over the rails tests. If it's in the migration
      it will 'respond_to'
    - adds nothing as a reference over the source of truth:
      schema.rb
----------------------------
40 tests removed, coverage remains the same.


notapatch:
done!

PR #1083, opened 2019-10-05
Remove testing of active record associations

notapatch:
- 5 association tests in the 100s of possible associations
 - reproduces what rails has already tested
 - the source of truth is the model's rb file
   - it it's in the model's rb then it's working


notapatch:
Re-base!

PR #1084, opened 2019-10-07
[WIP] Fix: take students from waiting list separate of coach spots

notapatch:
### Fix: take students from waiting list separate of coach spots

Another issue I saw when testing: the limiting of emails to students on the wait list: 

> - students are only emailed on the wait list if there were free seats for students *and* coaches

It looked like an indentation issue but maybe this is a business choice?? But why would you want a workshop full of coaches and the students not being offered spots that have become free? Commit 1e0084e will give maxed out coach workshops waitlisted students.

### DIFF
--- a/spec/fabricators/workshop_fabricator.rb
+++ b/spec/fabricators/workshop_fabricator.rb
@@ -70,6 +70,16 @@ Fabricator(:workshop_no_spots, class_name: :workshop) do
   end
 end
 
+Fabricator(:workshop_no_coaches, class_name: :workshop) do
+  date_and_time Time.zone.now + 2.days
+  title Faker::Lorem.sentence
+  description Faker::Lorem.sentence
+  chapter
+  after_build do |workshop|
+    Fabricate(:workshop_sponsor, workshop: workshop, sponsor: Fabricate(:sponsor, number_of_coaches: 0), host: true)
+  end
+end
+
 Fabricator(:past_workshop, from: :workshop) do
   date_and_time 3.months.ago
   ends_at 3.months.ago + 2.hours
--- a/spec/models/invitation_manager_spec.rb
+++ b/spec/models/invitation_manager_spec.rb
@@ -130,4 +130,48 @@ RSpec.describe InvitationManager, type: :model  do
       invitations.each { |invitation| expect(invitation.reload.reminded_at).to_not be_nil }
     end
   end
+
+  describe '#send_waiting_list_emails' do
+    it 'emails coaches when there are free coach spots' do
+      waitinglist_invitation = Fabricate(:waitinglist_invitation, workshop: workshop, role: 'Coach')
+
+      expect(WorkshopInvitationMailer).to receive(:notify_waiting_list).once
+        .with(waitinglist_invitation)
+        .and_call_original
+
+      manager.send_waiting_list_emails(workshop)
+    end
+
+    it 'does not email coaches when no coach spots are available' do
+      workshop = Fabricate(:workshop_no_coaches)
+      waitinglist_invitation = Fabricate(:waitinglist_invitation, workshop: workshop, role: 'Coach')
+
+      expect(WorkshopInvitationMailer).not_to receive(:notify_waiting_list)
+        .with(waitinglist_invitation)
+        .and_call_original
+
+      manager.send_waiting_list_emails(workshop)
+    end
+
+    it 'emails students when there are free student spots' do
+      waitinglist_invitation = Fabricate(:waitinglist_invitation, workshop: workshop, role: 'Student')
+
+      expect(WorkshopInvitationMailer).to receive(:notify_waiting_list).once
+        .with(waitinglist_invitation)
+        .and_call_original
+
+      manager.send_waiting_list_emails(workshop)
+    end
+
+    it 'does not email students when no student spots are available' do
+      workshop = Fabricate(:workshop_no_spots)
+      waitinglist_invitation = Fabricate(:waitinglist_invitation, workshop: workshop, role: 'Student')
+
+      expect(WorkshopInvitationMailer).not_to receive(:notify_waiting_list)
+        .with(waitinglist_invitation)
+        .and_call_original
+
+      manager.send_waiting_list_emails(workshop)
+    end
+  end
 end
--- a/app/models/concerns/workshop_invitation_manager_concerns.rb
+++ b/app/models/concerns/workshop_invitation_manager_concerns.rb
@@ -34,8 +34,8 @@ module WorkshopInvitationManagerConcerns
     def send_waiting_list_emails(workshop)
       workshop = WorkshopPresenter.decorate(workshop)
 
-      retrieve_and_notify_waitlisted(role: 'Coach') if workshop.coach_spaces?
-      retrieve_and_notify_waitlisted(role: 'Student') if workshop.student_spaces?
+      retrieve_and_notify_waitlisted(workshop, role: 'Coach') if workshop.coach_spaces?
+      retrieve_and_notify_waitlisted(workshop, role: 'Student') if workshop.student_spaces?
     end
     handle_asynchronously :send_waiting_list_emails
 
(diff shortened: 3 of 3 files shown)


notapatch:
@matyikriszta - I don't know if you've reviewed this or not but if you have the time -[ I think emails telling people a spot has become available aren't being sent](https://github.com/codebar/planner/pull/1084/commits/0677735c33e6552c7023059dfafa2214ea1e9340). If you like I can break the pull request up and leave the other bugfix for another day.

app/models/invitation_manager.rb
```diff
@@ -61,12 +61,12 @@ def send_waiting_list_emails(workshop)
...
         waiting_list.destroy
       end
+    end
 
-      if workshop.host.seats > workshop.attending_students.length
-        WaitingList.by_workshop(workshop).where_role('Student').each do |waiting_list|
-          WorkshopInvitationMailer.notify_waiting_list(waiting_list.invitation).deliver_now
-          waiting_list.destroy
-        end
+    if workshop.host.seats > workshop.attending_students.length
```
matyikriszta: @notapatch Rubocop is reporting a `Use a guard clause instead of wrapping the code inside a conditional expression.` offence for me on this line. Could we fix this as part of this PR?

spec/fabricators/workshop_fabricator.rb
```diff
@@ -54,6 +54,16 @@
   end
 end
 
+Fabricator(:workshop_barring_coaches, class_name: :workshop) do
```
matyikriszta: @notapatch similarly, can we rename this `workshop_no_coaches` please?

spec/fabricators/workshop_fabricator.rb
```diff
@@ -43,7 +43,7 @@
   end
 end
 
-Fabricator(:workshop_no_spots, class_name: :workshop) do
+Fabricator(:workshop_barring_spots, class_name: :workshop) do
```
matyikriszta: @notapatch I actually prefer `workshop_no_spots` name here, barring might be hard to understand for some contributors (I myself am an English as second language speaker).

matyikriszta:
@notapatch sorry, just got around to reviewing this, I'm 99% percent sure the fact we are only emailing student when coach spots become available as well is an indentation issue and not a conscious business decision (which would be weird because that logic makes no sense at all). I left you a few comments, once those are addressed I'm happy to merge.

spec/fabricators/workshop_fabricator.rb
notapatch: RE: second language: I did not know that.
I'll aim to keep the language straight forward - I didn't think about that despite talking to two Fins this week.

app/models/invitation_manager.rb
notapatch: I think that it breaks the symmetry you have one mailer in a block and the other protected by guard clause - so you are going to wonder why it's different - after some time you would realise there isn't a difference. But we can have it both ways. I'll refactor out the blocks and make both of them guard clauses.

spec/fabricators/workshop_fabricator.rb
notapatch: I've pressed "resolve conversation" as you probably know by now I have no idea what all these buttons do.

notapatch:
security/snyk putting vulnerabilities on all pull requests rather just notifying you that a Gem should be updated. Bad dog! 

I think the Rubocop errors show why it's important to try and clear them out the way as systematically as possible as it slows down development for the developer and the reviewer.

matyikriszta:
@notapatch so I was thinking about this issue this weekend (as you do) and I remembered a scenario in which you might not want to automatically RSVP more students from the waiting list unless another coach signs up. Imagine if a coach has dropped out and the coaches list is short by 1 person - in this case we do not want to bump any more students from the waiting list as they might not have a coach to mentor them. We need to devise some sort of check to ensure that student and coach numbers stay balanced. At our workshops we aim for a 2:1 student to coach ration (3:1 if we are super strapped for coaches). So if the coach list is not full, before we bump another student from the waiting list we need to check if the students and coach ration is 2:1 at most, if not we should not RSVP the student. If the coaches list if full we can proceed with RSVPing until a coach drops out. Let me know if this makes sense or you want me to clarify anything.

notapatch:
Makes sense ... I'll have a think and get back to you if I have any questions.  👍

notapatch:
[I've moved the other commit out of here](https://github.com/codebar/planner/pull/1132) and reassuringly it now breaks this build which this commit depends on).

The requirement is making sense in my head (code review completed during sleep). The main issue is the guard clauses. They are getting information from the workshop and the sponsor and then making comparisons rather than the OO way of asking the workshop if I can have more students and coaches. In other words, spreading the workshop full logic around the program rather than keeping it in, say, the Workshop object.
```
      # bad OO
      if workshop.host.coach_spots > workshop.attending_coaches.length
           "mail the coaches code"
       end

      if workshop.host.seats > workshop.attending_students.length
         "mail the students code"
      end

       # Better OO (current aim)
        "mail the coaches code" if workshop.coach_spaces_free?
        
       "mail the students" if workshop.students_spaces_free?
```

I'll need to review workshop and probably sponsor production and tests. To this end, [I will need the removal of monkey patching code back (any reservations?)](https://github.com/codebar/planner/pull/1107) and I'll try and get a few more general test code changes before doing this. (I am assuming that this isn't urgent - just a nice to have when I have the time).

notapatch:
Was thinking of something like this:  

```ruby
class Workshop
  def coach_spaces_free?
    host.coach_spots > attending_coaches.length
  end
        
  def students_spaces_free?
    return false unless host.seats > attending_students.length
    return false unless enough_coaches_for_students

    return true
  end

  private

  def enough_coaches_for_students
    attending_coaches.length * COACH_TO_STUDENT_RATIO  > attending_students.length
  end
end

```

Any good? Only thing it won't cope with is coaches suddenly dropping out. Probably want a separate panic email to all none-attending coaches to see if they would appear...

notapatch:
@despo 

I'm not sure what is happening in the case of `WorkshopInvitationManagerConcerns::send_waiting_list_emails(workshop)` as the code is broken because the call to `retrieve_and_notify_waitlisted` requires `workshop` parameter which is missing. I am assuming you are aware - it is also untested and I have updated the testing code from this PR to get it working with the new codebase.

If the fix d192c29 and/or testing 47f45ec is useful for you then feel free to merge if this has no value now then I'm happy to see this closed.

despo:
@notapatch so this code is not actually used so this is why it hasn't been noticed.
This is because of https://github.com/codebar/planner/blob/master/app/controllers/waiting_lists_controller.rb#L22 - the plan was to enable auto_rsvp but **I think** we never got around to exposing the `auto_rsvp`.

However this is a good opportunity to first fix the email and maybe take steps into implementing the auto_rsvp feature so happy to look at this again.

spec/models/invitation_manager_spec.rb
```diff
@@ -130,4 +130,48 @@
...
+      waitinglist_invitation = Fabricate(:waitinglist_invitation, workshop: workshop, role: 'Coach')
+
+      expect(WorkshopInvitationMailer).to receive(:notify_waiting_list).once
+        .with(waitinglist_invitation)
+        .and_call_original
+
+      manager.send_waiting_list_emails(workshop)
+    end
+
+    it 'does not email coaches when coach spots full' do
```
despo, line 145: maybe reword to `does not email coaches when no coach spots are available`

spec/models/invitation_manager_spec.rb
```diff
@@ -130,4 +130,48 @@
...
+      waitinglist_invitation = Fabricate(:waitinglist_invitation, workshop: workshop, role: 'Student')
+
+      expect(WorkshopInvitationMailer).to receive(:notify_waiting_list).once
+        .with(waitinglist_invitation)
+        .and_call_original
+
+      manager.send_waiting_list_emails(workshop)
+    end
+
+    it 'does not email students when student spots full' do
```
despo, line 166: maybe reword to `does not email students when no student spots are available`

spec/models/invitation_manager_spec.rb
notapatch: done!

spec/models/invitation_manager_spec.rb
notapatch: done!

PR #1107, opened 2019-10-18
[no deploy] RSpec modernizing - remove monkey patch and tag with metadata

notapatch:
The current suite is a mix of new and old style boiler plate and this moves the suite to all modern.
1. Remove the 'describe' and 'feature' monkey patch
2. add metadata types where appropriate: models, features and mailer specs

- only test suite changes / no production code changed

---
### Monkey Patch

Modern RSpec suites remove the monkey patching on the keywords 'describe' and 'feature' by adding 'RSpec' name space.

describe => RSpec.describe
feature => RSpec.feature

[RSpec.describe vs describe](https://stackoverflow.com/questions/24747448/rspec-describe-vs-describe)

---
### Metadata types

RSpec 2 guessed the type of test (model or mailer or feature) from directory structure. RSpec 3 now uses metadata types instead.

RSpec.describe "some test" do => RSpec.describe "some test" type: :model do
RSpec.feature "some test" do => RSpec.feature "some test" type: :feature do

[RSpec directory Structure
](https://relishapp.com/rspec/rspec-rails/docs/directory-structure)

[What does including the type in the describe block of a rspec do?
](https://stackoverflow.com/questions/45023022/what-does-including-the-type-in-the-describe-block-of-a-rspec-do)

### DIFF
--- a/spec/controllers/admin/member_notes_controller_spec.rb
+++ b/spec/controllers/admin/member_notes_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe Admin::MemberNotesController, type: :controller do
+RSpec.describe Admin::MemberNotesController, type: :controller do
   let(:member) { Fabricate(:member) }
   let(:admin) { Fabricate(:chapter_organiser) }
   let!(:member_note) { Fabricate(:member_note) }
--- a/spec/controllers/admin/sponsors_controller_spec.rb
+++ b/spec/controllers/admin/sponsors_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe Admin::SponsorsController, type: :controller do
+RSpec.describe Admin::SponsorsController, type: :controller do
   let(:member) { Fabricate(:member) }
   let(:member1) { Fabricate(:member) }
   let(:address) { Fabricate(:address) }
--- a/spec/controllers/admin/workshops_controller_spec.rb
+++ b/spec/controllers/admin/workshops_controller_spec.rb
@@ -1,6 +1,6 @@
 require 'spec_helper'
 
-describe Admin::WorkshopsController, type: :controller do
+RSpec.describe Admin::WorkshopsController, type: :controller do
   let!(:workshop) { Fabricate(:workshop) }
   let(:admin) { Fabricate(:member) }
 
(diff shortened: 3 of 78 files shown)


notapatch:
@matyikriszta I'm worried that I hadn't explained this one very well and it got skipped.

---
The RSpec gem version passed 3.0 without updating the test code. This introduces bugs. 

> In order to include the correct rspec-rails support functions, the specs need
> to have the appropriate corresponding metadata :type value:

From: [RSpec 3.9-  Directory Structure](https://relishapp.com/rspec/rspec-rails/v/3-9/docs/directory-structure)

This PR is just a massive search and replace.

This times 78

```ruby
  # monkey patch adds RSpec code to the keyword  'describe'
  #   |                   Namespace with RSpec and add type of test with a keyword
- describe Address do   #     |
+ RSpec.describe Address, type: :model  do
  subject(:address) { Fabricate.build(:address) }
end
```

The other types are `feature` and `mailer` but some of them don't have a `type:` and I only add `RSpec` namespace.




No production issues.

notapatch:
@matyikriszta - this is ready for review

spec/presenters/jobs_presenter.rb
```diff
@@ -1,6 +1,6 @@
 require 'spec_helper'
```
matyikriszta: @notapatch I noticed there's a `spec/presenters/job_presenter_spec.rb` file and a `spec/presenters/jobs_presenter.rb` file, do we need both?

notapatch: Interesting. So, job_presenter is a singular presenter and job**s**_presenter is a collection of presenters. I would say that the collection tests were named incorrectly. 

| Model | Tests |
|-----|---------|
| `app/presenters/job_presenter.rb`  | `spec/presenters/job_presenters_spec.rb` |
| `app/presenters/jobs_presenter.rb` | ~spec/presenters/jobs_presenters.rb~  |

It should have been named **jobs_presenter_spec.rb** . I will make that change. 

I am not sure why this is the only collection presenter and it might be worth knowing why that is. I don't know enough about jobs_presenter other than it is used in the app.

matyikriszta: @notapatch makes sense, thanks for updating the name of the spec.

PR #1109, opened 2019-10-18
Fix notify_waiting_list.html.haml template

jabawack81:
the body of the `if organiser.mobile.present?` statement wasn't indented enough and this was breaking the `HAML` engine


notapatch:
Agreed. I put a pull request in covering it - it was also missing tests which allowed it to happen and not get picked up.
https://github.com/codebar/planner/pull/1084

jabawack81:
Oh sorry I missed your PR I'll close this one

PR #1111, opened 2019-10-19
Add instructions for adding yourself as an admin locally

turger:
I tested planner locally and found out that I could't access to admin page to test some functionalities. I didn't find instructions to add myself as and admin locally, so I added these instructions for everyone to use.


CONTRIBUTING.md
```diff
@@ -17,26 +17,39 @@
...
+5. (Optional) Note that to be able to use the page as an admin, you must first give yourself admin privileges.
+  Make sure you have started your app in docker (if using locally, skip the docker-related parts in these instructions) and signed up as a user on your locally running app.
+  * First get your id:
+    * Run `docker ps` to get the id of your running postgres container id.
+    * Run `docker exec -it <your-id-here> psql -U postgres planner_development` where planner_development is the database name.
+    * Query `SELECT * from members WHERE name = '<Your first name>';` (remember the ; in the end of the line, it's very important!)
+    * Exit with `exit` twice.
+  * Then access rails console:
+    * Run `./bin/drails console`
+    * Then in the console run: `Member.find(<your-id>).add_role(:admin)`
```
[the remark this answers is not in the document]
turger: Thanks for your comments! I tried this with `"Member.where(name: 'Taru').add_role(:admin)"` but only got error:
```
/usr/local/bundle/gems/activerecord-4.2.11.1/lib/active_record/relation/delegation.rb:136:in `method_missing': D, [2019-10-20T15:44:22.443021 #1] DEBUG -- :   Member Load (1.1ms)  SELECT "members".* FROM "members" WHERE "members"."name" = $1  [["name", "Taru"]]
undefined method `add_role' for #<Member::ActiveRecord_Relation:0x00000000066e9b18> (NoMethodError)
```
But the other one `Member.last.add_role(:admin)` worked like a charm! I will add `dadmin`-script as you suggested, because I too think that's a much nicer way for solving this. I will also add your instructions for non-docker users to contributing-readme. 👍

turger:
Made suggested changes, ready for a new check!

CONTRIBUTING.md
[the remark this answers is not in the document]
notapatch: re: Where Oops.

### Where solution

`where` returns, sort of in the loosest way, an array of Members and an array won't know what an `add_role` method is so you got the "undefined method" exception. A fix is to get the actual member from the sort-of array: 

`Member.where(name: '<you first name'>).first.add_role(:admin)` 

Assuming your name is unique.

### find_by solution

`find_by` returns an actual Member and not an array of members. Assuming your the first person named 'Taru' *thinks* which I reckon there's a decent chance. 

`Member.find_by(name: '<your first name>').add_role(:admin)`



## References
[find_by](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by)
[where](https://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-where)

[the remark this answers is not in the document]
turger: Cool, thanks! I edited instructions and `dadmin` script to use `find_by`.

biggianteye:
This is great. I've been looking for how to do this. I think this is a great addition.

> Related to this PR: I work in native mode and not docker so I can become admin at the command line (without even using the console):
> 
> `rails runner "Member.where(name: '<Your first name>').add_role(:admin)"`
> 
> or
> 
> 1. sign up
> 2. `rails runner "Member.last.add_role(:admin)"`

Adding `docker-compose run --rm web` before those commands is enough to get them running in docker. I've just done the second option like this and it worked perfectly.

```
docker-compose run --rm web rails runner "Member.last.add_role(:admin)"
```

I'm sure a wrapper script could be written to encompass both. eg. no arg = make the last user an admin, arg supplied = make that specific user an admin.

bin/dadmin
```diff
@@ -0,0 +1,19 @@
...
+  echo "cleaning up"
+
+  # ignore errors
+  set +e
+  docker-compose down
+
+  exit $code
+}
+
+trap cleanup EXIT
```
biggianteye: I don't recommend having this boilerplate taken from the other scripts. It will shut everything down any time you run this script. Probably not what you want to do.

Something like this should be sufficient:
```sh
#!/bin/sh

if [ $# -gt 0 ]
then
  docker-compose run --rm web rails runner "Member.find_by(name: '$@').add_role(:admin)"
else
  docker-compose run --rm web rails runner "Member.last.add_role(:admin)"
fi
```

notapatch: OOOkay, I get it. Doh! Thanks biggianteye. The start of the original script clears up the current docker instance? Marvelous.

biggianteye: This is a tangent, but there are two things happening.

1. `trap cleanup EXIT`

The  `trap` keyword allows you to intercept various signals received by the script. This particular line means that when the `EXIT` signal is received (ie. normal termination), the `cleanup` function is called.

This minimal example demonstrates this behaviour:
```sh
#!/bin/sh

function cleanup
{
    echo "Cleanup function being called"
}

trap cleanup EXIT

echo "Hello world"
```

```
[burhan@enceladus ~] {0} $ ./test.sh
Hello world
Cleanup function being called
```

2. `docker-compose down`

This is the important line in the `cleanup` function. It tells docker to bring down all the containers defined in `docker-compose.yml`. ie. `web` (the web server) and `db` (the database).

notapatch: Thanks that's really clear. I've only recently worked out how important scripts are for getting work done. Marvellous, a 3 person solution.

CONTRIBUTING.md
```diff
@@ -17,26 +17,31 @@
     ```
 
 4. Have a look around to get a feel for the app.
-5. Run the tests. We only take pull requests with passing tests, and it's great to confirm that you have a clean slate.
+
+5. (Optional) Note that to be able to use the page as an admin, you must first give yourself admin privileges. Make sure you have started your app and signed up as an user on your locally running app.
+  * Non-docker users can run this on command line: `rails runner "Member.find_by(name: '<your first name>').add_role(:admin)"`
```
matyikriszta: @turger this is a minor point but can we suggest people find themselves by email address? I think that's a more reliable piece of information then first name.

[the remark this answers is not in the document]
matyikriszta: @turger as @notapatch mentioned can we add this to the native installation steps as well?

matyikriszta:
@turger this is a great addition, thanks for the PR. I left you two small comments. As a more general note, upon further introspection I find it weird we have setup instruction in 3 different places, e.g. the readme, the contributing guide and also the guide on how to set up the app directly to you local environment.

CONTRIBUTING.md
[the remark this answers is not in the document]
turger: Sure, I'll move the instructions!

turger: Very good point, I'll change name to email!

bin/dadmin
turger: Cool, thanks! I made suggested changes to a script and now it's much better.

biggianteye:
> @notapatch @biggianteye let me know if that sounds good/you have any feedback.

Sorry, I missed this. This patch is wonderful and is a great quality of life improvement for new people to the project. I've already used it a number of times to test some things out. I'm all for more things like this.

As to `CONTRIBUTING.md`, I'm more than happy to look at any changes that are made to it. Github provides some "[anatomy of an open source project](https://opensource.guide/how-to-contribute/#anatomy-of-an-open-source-project)" guidance which talks about what to put in different documentation. Things grow organically so it's useful to occasionally take a step back and re-evaluate if things are in the right place.

Issue #616, opened 2017-10-27
[accessibility] Keyboard focus - visual issues


BNewing:
When tabbing through the site and you get to the footer, the focus changes the text colour so that links are very hard to read:

![image](https://user-images.githubusercontent.com/16868713/32127042-061d04a4-bb6c-11e7-880b-6daa1cb7ce36.png)

It's also recommended in Heydon Pickering's "Inclusive Design Patterns" that you add a background colour to things keyboard focus is on (something that the gov.uk website does):

![image](https://user-images.githubusercontent.com/16868713/32127027-ec2d8816-bb6b-11e7-9833-510469e2dcc0.png)

Lastly, the outline on nav elements looks a little odd:

![image](https://user-images.githubusercontent.com/16868713/32126961-8505385a-bb6b-11e7-8c1f-6e382b98b0b4.png)

turger:
Seems that the footer links are nowadays easier to read, but I can fix outline from nav elements and add a background colour to hovered items.

turger:
https://github.com/codebar/planner/pull/1114

BNewing:
Looks good! 

If you wanted to go one step further, you could also check out the colour contrast against AA and AAA :) I think at the moment from looking at it, it might pass AA but not AAA. I often use this [colour contrast tool](https://contrast-checker.glitch.me) for that, as you can tweak around the hues and things. 
 
It is still great as it is though, just sharing in case you're interested :)

turger:
Cool thanks Newing! I will check the colour contrasts and make some edits if needed.

turger:
I edited colours so that they now get AAA from text contrast with object background and AA from object contrast with page background.

BNewing:
Awesome! :D

PR #1114, opened 2019-10-20
Keyboard focus - visual issues #616

turger:
Fixed outline from nav elements and added a background colour to some hovered items.

menu
stripe-buttons-2
footer2
event-listing2

### DIFF
--- a/app/assets/stylesheets/foundation_and_overrides.scss
+++ b/app/assets/stylesheets/foundation_and_overrides.scss
@@ -1390,6 +1390,31 @@ body.dashboard-faq dl dt {
   border-bottom: none;
 }
 
+a:focus {
+  outline: 5px auto $codebar-blue;
+  color: darken($codebar-blue, 30%);
+}
+
+a.button:focus {
+  background: darken($codebar-blue, 50%);
+  border-color: darken($codebar-blue, 50%);
+  color: $white;
+  outline: none;
+}
+
+.top-bar-section li:not(.has-form) a:not(.button):focus,
+.title > a:focus {
+  background: darken($codebar-blue, 50%);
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/foundation_and_overrides.scss
```diff
@@ -1390,6 +1390,31 @@ body.dashboard-faq dl dt {
   border-bottom: none;
 }
 
+a:focus {
```
notapatch: I have not changed one line of CSS in Codebar planner and I have never used foundation!

Is it _possible_ to fit it into `assets/stylesheets/partials/_links.scss ` and while you're at it. change the code to use SCSS?

Something like:

```scss
a {
    &:hover {
       etc etc 
    }
   &:focus {
       your stuff etc etc
    }
} 
```

app/assets/stylesheets/foundation_and_overrides.scss
```diff
@@ -1390,6 +1390,31 @@ body.dashboard-faq dl dt {
...
+}
+
+.top-bar-section li:not(.has-form) a:not(.button):focus,
+.title > a:focus {
+  background: darken($codebar-blue, 50%);
+  color: $white;
+  outline: none;
+}
+
+.main-footer a:focus {
```
notapatch: Same deal but in `assets/stylesheets/partials/_layout.scss` ?

```scss
 a {
    color: $white;
    border: none;

    &:focus {
        // awesome code
    }

    &:hover {
      color: $secondary-color;
    }
  }

```

There may be places for the other code? But I'll shut up until you've had a look. _If_ I am right all the css should probably find a new home.

Issue #783, opened 2018-08-20
Enable description field in workshop creation


nadiakiamilev:
## Description of the issue 📄
Currently, when you create a workshop, you are only able to add basic details about the workshop (date, time, number of seats available). Workshop sessions are missing a description field that would allow you to add text about the workshop. 

## Screenshots 📷
N/A

## Steps to fix 🛠
* add a form field for a description in the workshop creation form

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

KimberleyCook:
@nadiakiamilev what type of information are you thinking of wanting to add?

nadiakiamilev:
We wanted to be able to add personalised text about the session. Something like:

"Join us for our first Codebar workshop at Designit! If you’ve been keen on learning how to code, this is the place for you. 

Snacks & refreshments will be provided. Please bring your laptop & charger. We have coaches that can help in both Norwegian & English."

We also felt that it would allow us to have a bit more of a personalised voice in the workshop description :)

matyikriszta:
@nadiakiamilev @KimberleyCook I think this is a good idea. I think especially for chapters outside the UK it's useful to say people can help in the local language (as currently the website is all in English and some people interpret that as the workshop being run in English as well). Maybe next to the text  area input we could show what the default text is (I think we should keep this), so people don't repeat themselves, but they could add additional details if the wanted to.

AlexAvlonitis:
Hi, can i take this?

turger:
I can start doing this.

turger:
Here's the pull request: https://github.com/codebar/planner/pull/1115

matyikriszta:
@nadiakiamilev this has now been implemented and will be deployed shortly, closing this issue.

PR #1115, opened 2019-10-20
Enable description field in workshop creation #783

turger:
Added description field to workshops.

admin-workshop-edit
admin-workshop-view
workshop-view-description

### DIFF
--- a/app/assets/stylesheets/partials/_admin_workshop.scss
+++ b/app/assets/stylesheets/partials/_admin_workshop.scss
@@ -2,3 +2,6 @@
   padding-top: 10px;
 }
 
+#workshop_description {
+  height: 150px;
+}
--- a/app/views/admin/workshops/_edit_form.html.haml
+++ b/app/views/admin/workshops/_edit_form.html.haml
@@ -13,9 +13,19 @@
   .row
     .large-6.columns
       = f.input :host, as: :select, collection: Sponsor.all, include_blank: true, selected: (@workshop.host.id rescue '')
+  %br
   .row
     .large-6.columns
       = f.association :sponsors
+  %br
+  .row
+    .large-6.columns
+      %h6 Default description text:
+      %small= t("workshops.lead")
+  .row
+    .large-6.columns
+      = f.input :description, as: :text, label: 'Additional description (supports HTML)',
+      placeholder: 'Use this space to write a description of the workshop.'
   %br
   .row
     .large-6.columns
--- a/app/views/admin/workshops/show.html.haml
+++ b/app/views/admin/workshops/show.html.haml
@@ -47,6 +47,9 @@
           %strong.label.round.primary #{@workshop.venue.seats} student spots, #{@workshop.venue.coach_spots} coach spots
           - unless @workshop.spaces?
             %label.label.round.warning Full
+          %p.description
+            %br
+            = sanitize(@workshop.description)
     %br
     .row
       .large-3.small-6.columns#host
--- a/app/views/workshops/show.html.haml
+++ b/app/views/workshops/show.html.haml
@@ -18,7 +18,9 @@
     .row
       .medium-12.columns
         %p.lead
-          Attend our workshops to learn programming in a safe and supportive environment at your own pace, or to share your knowledge and coach our students.
+          = t("workshops.lead")
+        %p.description
+          = sanitize(@workshop.description)
 
         - unless current_user and current_user.banned?
           = render 'workshop_actions'
--- a/app/controllers/admin/workshops_controller.rb
+++ b/app/controllers/admin/workshops_controller.rb
@@ -100,7 +100,7 @@ class Admin::WorkshopsController < Admin::ApplicationController
   def workshop_params
     params.require(:workshop).permit(:local_date, :local_time, :chapter_id,
                                      :invitable, :seats, :rsvp_open_local_date,
-                                     :rsvp_open_local_time, :ends_at, sponsor_ids: [])
+                                     :rsvp_open_local_time, :ends_at, :description, sponsor_ids: [])
   end
 
   def chapter_id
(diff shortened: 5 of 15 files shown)


KimberleyCook:
@turger thank you for this contribution Taru 🎉

app/assets/stylesheets/partials/_admin_workshop.scss
```diff
@@ -2,3 +2,10 @@
   padding-top: 10px;
 }
 
+#workshop_description {
+  height: 150px;
+}
+
+.lead {
```
matyikriszta: @turger was wondering if there's a Foundation class we could use here instead of defining our own.

matyikriszta: @turger I wonder if there's a Foundation class we can use here instead of defining our own. If not there must be another place we have defined a small class in the CSS, I'm not sure where off the top of my head but I would have a quick look. Or perhaps use the `<small>` tag?

app/views/admin/workshops/_edit_form.html.haml
```diff
@@ -11,10 +11,19 @@
...
     .large-6.columns
       = f.association :sponsors
   %br
+  .row
+    .large-6.columns
+      %h6 Default description text:
+      %p= t("workshops.lead")
+    .large-6.columns
+      = f.input :description, as: :text, label: 'Additional description',
+      placeholder: 'Use this space to write a description of the workshop. You can use HTML here!'
```
matyikriszta: @turger can we move the bit about HTML support out from the placeholder and into the label? For example `Additional description (supports HTML)`.

app/views/workshops/show.html.haml
```diff
@@ -13,7 +13,9 @@
     .row
       .medium-12.columns
         %p.lead
-          Attend our workshops to learn programming in a safe and supportive environment at your own pace, or to share your knowledge and coach our students.
+          = t("workshops.lead")
+        %p.description
+          = @workshop.description.html_safe
```
matyikriszta: @turger brakeman is reporting a `Unescaped model attribute` error on this line. Could we address this before merging?

notapatch: Brakeman is probably right. The description field is a particular problem for XSS issues on Planner. I have a PR sitting on my machine which would covers many issues on the application - I think most of them are descriptions (it's not so bad as a Student can't cause them. It would take a malicious organiser to cause a problem). I was kinda waiting for the PR load to decrease but it seems the opposite is happening.

turger: Not sure how I should fix this error? I could remove the possibility to add html to the description or just somehow ignore the error?

app/views/admin/workshops/_edit_form.html.haml
turger: Good idea, I'll do the change.

app/assets/stylesheets/partials/_admin_workshop.scss
turger: Yes! I changed to the small tag and removed this.

app/views/workshops/show.html.haml
notapatch: Remove `html_safe` string property and call `Sanitize` method instead. The difference is that they both allow HTML tags through but `html_safe` allows Javascript and `sanitize` doesn't.

In my fix I've added sanitize to the base class of the presenters so it is a matter of adding a single function call to the output string - it's a bit long otherwise:

`ActionController::Base.helpers.sanitize('<b>bold</b>')`

### Reference
[Sanitize](https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html) 
[html_safe](https://api.rubyonrails.org/classes/String.html#method-i-html_safe)
[Using sanitize within a Rails controller](https://stackoverflow.com/questions/3985989/using-sanitize-within-a-rails-controller)

matyikriszta: @turger as @notapatch pointed out your best bet is sanitize, in a haml template you should be able to do `= sanitize(t("workshops.lead") )`.

app/views/workshops/show.html.haml
```diff
@@ -18,7 +18,9 @@
     .row
       .medium-12.columns
         %p.lead
-          Attend our workshops to learn programming in a safe and supportive environment at your own pace, or to share your knowledge and coach our students.
+          = t("workshops.lead")
+        %p.description
+          = sanitize(@workshop.description.html_safe)
```
notapatch: Almost there ... so html_safe allows HTML tags and Javascript and sanitize allows only HTML tags. so you can remove the html_safe and it does the same thing. We want to get rid of html_safe because it's very misleading.

app/views/workshops/show.html.haml
turger: I added sanitize and now brakeman seems to pass!

app/views/admin/workshops/show.html.haml
```diff
@@ -47,6 +47,9 @@
           %strong.label.round.primary #{@workshop.venue.seats} student spots, #{@workshop.venue.coach_spots} coach spots
           - unless @workshop.spaces?
             %label.label.round.warning Full
+          %p.description
+            %br
+            = @workshop.description.html_safe
```
notapatch: So this is the same security hole ... you need to sanitize this as well. Same fix. But you can see the issue here every time you use description you are having to think about security. Rather than sanitizing it within the WorkshopPresenter.description - so you never have to think about it (which is my future pull request).

app/views/workshops/show.html.haml
turger: Thanks for clarifying this! Removed html_safe.

notapatch:
Looks good. Same deal, sometime during the week and probably later in the day it will be reviewed.

Question: have you rebased it recently? I can't tell (there may be a way but I don't know how)

It's a good idea to keep it rebased while waiting otherwise you wake up with the message "I want to merge .... _but_ can you rebase first" instead of the much more enjoyable "merged!"

turger:
@notapatch yup I rebased this branch today 👍

PR #1122, opened 2019-10-25
Replace string array with word array (Rubocop autocomplete)

notapatch:
- rubocop default
 - rubocop --only Style/WordArray -a

---

[Rubocop WordArray
](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/WordArray)
 - fixed via autocomplete
 - consistency as already used in 12 of the 18 possible places. 

```ruby
EnforcedStyle: brackets

# good
['foo', 'bar', 'baz']

# bad
%w[foo bar baz]
```

### DIFF
--- a/app/controllers/admin/chapters_controller.rb
+++ b/app/controllers/admin/chapters_controller.rb
@@ -50,7 +50,7 @@ class Admin::ChaptersController < Admin::ApplicationController
     authorize chapter
     type = params[:type]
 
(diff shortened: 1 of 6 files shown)


notapatch:
If a fix the codeclimate issue (which this change does not introduce) - it then breaks another type (Layout/IndentationWidth). I think it is cleaner to work through a class of error than trying to work through a cascade of rubocop errors.

PR #1123, opened 2019-10-25
Pr layout - rubocop spacings

notapatch:
Covers most of Layout Rubocop - lots of space changes for the good

### DIFF
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -2,5 +2,4 @@ class Admin::ApplicationController < ApplicationController
   include Pundit
 
   before_action :authenticate_admin_or_organiser!
-
 end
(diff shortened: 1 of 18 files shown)


matyikriszta:
@notapatch could you please fix that one codeclimate issue that's breaking the checks please? It would be good if we could merge in PRs with passing checks (otherwise I don't see the point of having them).

notapatch:
Understood. Got it passing.

The code will be properly refactored when I have tests in place - I daren't do this without tests...

```
  def invitability
    errors.add(:coach_spaces, 'must be set') if coach_spaces.blank?
    errors.add(:student_space, 'must be set') if self.student_spaces.blank?
    errors.add(:invitable, 'Fill in all invitations details to make the event invitable') if coach_spaces.blank? || student_spaces.blank?
  end
```

notapatch:
You're right no excuses coveralls is green at -0.2% - only green should go in.

PR #1125, opened 2019-10-26
Use ajax for step2 on registration

jabawack81:
as per title, this PR updates the forms in Step 2 (subscription to chapters) of the registration process to Ajax.

I think it will improve the UX.

This PR is meant to be the first proposal and open a discussion on the implementation:

I've added a plus icon to the groups the user is not subscribed.
Screenshot_2019-10-26 codebar

While the Ajax call is performed (the spinner spins).
Screenshot_2019-10-26 codebar(1)

On success.
Screenshot_2019-10-26 codebar(2)

On error.
Screenshot_2019-10-26 codebar(3)


notapatch:
Hi  jabawack81

I have not played with the pull request (it's late for me)  but it would make sense to make joining smoother - with the standard warning that it's not my decision.

Codeclimate bugs will need to be fixed - even though you did not originate them. If you find that annoying I recommend joining me in submitting Rubocop fixes for classes of bugs - there are many hundreds left.
 
Not going to be much help on this but a couple of ideas are: 
 - the long strings might fit into the en.yml? Not sure till I've tried.
 
_Maybe_ you break it into methods for success and failure? Your guess is as good as mine. I looked at a couple of refactoring controller blog posts but nothing interesting.

 ```ruby
      if @subscription.save
        unless current_user.received_welcome_for?(@subscription)
          MemberMailer.welcome_for_subscription(@subscription).deliver_now
        end
        format.html do
          flash[:notice] = "You have subscribed to #{@subscription.group.chapter.city}'s #{@subscription.group.name} group"
          redirect_to :back
        end
        format.js { render 'create_success' }
      else
        flash[:notice] = @subscription.errors.full_messages.join('<br/>')
        format.html { redirect_to :back }
        format.js { render 'create_error' }
      end
```

You might want to put the Rubocop fixes in a separate commit to help make it clearer what is Ajax related or not.

**Comment by jabawack81** [comment]
@notapatch that's a good point let me do that, I'm planning open a PR taking care of all the Rubocop infraction left around the code.

**Comment by jabawack81** [comment]
@notapatch that's a good point let me do that, I'm planning open a PR taking care of all the Rubocop infraction left around the code.

**Comment by notapatch** [comment]
Do you mean a PR or just a commit in _this_ PR? I've not thought about which way might be better - It might make sense as a PR if you're not sure the actual code is going to be accepted. If it were me I would keep as the first commit with the offer of making it a PR if this PR gets held up.

BTW: have you [seen this sign up Issue](https://github.com/codebar/planner/issues/442)? It's like one of the earliest not dealt with. I know it's still active issue... and must be a concern.

**Comment by jabawack81** [comment]
an actual PR only for `rubocop` fix codebase wise and limit the fixes in this PR to only the code I touch, basicly applying the S of SOLID to it ;)

**Comment by notapatch** [comment]
Yes, It sounds like a good idea. Better to have 2 simple PRs than one complicated one.

**Comment by jabawack81** [comment]
I've update the PR fixing most of the RUbocop errors there are two errors left:

```

app/controllers/subscriptions_controller.rb:8:3: C: Metrics/AbcSize: Assignment Branch Condition size for create is too high. [24.37/15]
  def create ...
  ^^^^^^^^^^
app/controllers/subscriptions_controller.rb:8:3: C: Metrics/MethodLength: Method has too many lines. [17/10]
  def create ...
```

but I can't see any real way to fix this two

Jaba

PR #1128, opened 2019-10-28
Fix: lock travis to bundler = 1.17.3

notapatch:
- travis has been ignoring restriction and building with 2.0.2
   commit fixes bundler to 1.17.3 matching Gemfile.lock
   warning appears as:
   rails (4.2.11.1) has dependency bundler (< 2.0, >= 1.3.0),
   which is unsatisfied by the current bundler version 2.0.2,
   so the dependency is being ignored
 - rails_5 remove restriction

---
In development we use bundler 1.17.X we can see it in the Gemfile.lock at the bottom

However, Travis has been using 2.0.2 :-|

travis-putting-own-bundler-in
Say what now?
Screenshot 2019-10-28 at 12 51 16

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 dist: xenial
 language: ruby
-bundler_args: --without development
 rvm:
   - 2.4.2
 services:
   - postgresql
+before_install:
+  - gem install bundler --version '~> 1.17.3' --no-document
+  - bundle _1.17.3_ install --without development
 before_script:
   - bundle exec rake db:create db:schema:load
 


notapatch:
If you aren't Heroku ... this is a good time to say :)

notapatch:
@matyikriszta  - this is ready for review too. Esp if I'm right about Heroku

matyikriszta:
@notapatch we are hosted on Heroku, yes, if that's what you meant.

notapatch:
I've mothballed this PR. It doesn't make sense to be building on bundler version outside where it was tested (Rails 4.2 only tested with bundler 1.17.X) however, the builds are still passing and there's no interest in it. Closing!

PR #1131, opened 2019-10-29
[Rubocop] prefer square brackets for percent literal delimiters

notapatch:
- rubocop --only Style/PercentLiteralDelimiters


notapatch:

PR #1132, opened 2019-10-29
Fix: notify_waiting_list exception

notapatch:
- added basic test that email created without error

---
@matyikriszta ready for review - this is the exception commit in the notify_waiting_list  the other pull request will deal with ratios of coaches / students.

the original message follows:

> The only untested mailer is WorkshopInvitationMailer.notify_waiting_list. During the testing I found out that it was creating exceptions when rendering the email - Commit eb94549 fixes and tests that the page is now rendered.
> 
> You can easily see the exception using the action mailer preview:
> http://localhost:3000/rails/mailers/workshop_invitation_mailer/notify_waiting_list
> 
> Without live data I can't be definitive on this pull request, but given the exception, it's possible that no notify_waiting_list emails are being sent.
>


notapatch:

notapatch:
@matyikriszta, @KimberleyCook  - this is a fix to a critical bug at the heart of the system - messaging people when a waiting list spot becomes free. Things can get lost under a dozen bump bot requests and perhaps this happened here? 

When either of you are next free can you look at it? I am not alone in noticing it [@jabawack81 has also put a PR with the same fix.](https://github.com/codebar/planner/pull/1109) I have never asked a PR to queue jump before; I feel this is important but has been left unmerged for a number of weeks having tried to get it through with two different PRs (this one is simpler).

The bug is easy enough to demonstrate if you have a development server open and click on this link you get a syntax error. My bet is you would get the same in production which means no waiting free emails are being sent.

http://localhost:3000/rails/mailers/workshop_invitation_mailer/notify_waiting_list

```
SyntaxError at /rails/mailers/workshop_invitation_mailer/notify_waiting_list

syntax error, unexpected keyword_ensure, expecting keyword_end
Util.html_safe(_erbout);ensure;@haml_buffer = @haml_buffer.u
```

Once you've seen the bug it's a matter of a couple of git commands and you can see the fix working.

```
 git fetch upstream pull/1132/head:pr-fix-send_waiting_list_emails-exception
 git checkout pr-fix-send_waiting_list_emails-exception

```

The fix adds some missing spaces that broke the HAML and we're off working again.

notapatch:
Thank you. That's good news.

PR #1139, opened 2019-10-31
[Rubocop] Refactor unless/x.present? to if/x.blank?

notapatch:
- rails defines present? as !blank? if we take advantage of that
   we can make it easier to reason
 - rubocop --only Rails/Blank
 - rubocop rails default

---

```
  # activesupport/lib/active_support/core_ext/object/blank.rb
  # An object is present if it's not blank.
  #
  # @return [true, false]
  def present?
    !blank?
  end

```

[Rubocop Rails Blank
](https://www.rubydoc.info/gems/rubocop-rails/RuboCop/Cop/Rails/Blank)


notapatch:
@matyikriszta  - ready for review

PR #1141, opened 2019-10-31
Refactor conditionals to guard clauses

notapatch:
This is a Rubocop but not automatic replacement - so it needs to be read through a little slower.  It does tell you what you need to do. The replacement pattern is: 
 
| Old Conditional | New guard clause |
|:-----------|:--------------|
|  `if ( clause ) ` |  `return unless (clause)` |
|  `unless ( clause )`  |  `return if (clause`) |

### Tests

Before I changed anything I put `raise 'error'` in the conditional to see if some tests fail - these all do so there is some protection.  There are a couple of them I haven't converted, yet, because they had no tests to protect them.

```ruby
unless (clause)
  raise "error"
  # code to run if clause passed
end
```

#### To see the errors:

`rubocop --only Style/GuardClause`

### DIFF
--- a/app/controllers/admin/workshops_controller.rb
+++ b/app/controllers/admin/workshops_controller.rb
@@ -122,12 +122,11 @@ class Admin::WorkshopsController < Admin::ApplicationController
 
   def set_host(host_id)
     return unless host_id
-
     host = @workshop.workshop_sponsors.find_or_initialize_by(sponsor_id: host_id)
-    unless @workshop.host.eql?(host.sponsor)
-      @workshop.workshop_sponsors.where(sponsor: @workshop.host).destroy_all
-      host.update(host: true)
-    end
+    return if @workshop.host.eql?(host.sponsor)
+
+    @workshop.workshop_sponsors.where(sponsor: @workshop.host).destroy_all
+    host.update(host: true)
   end
 
   def set_organisers(organiser_ids)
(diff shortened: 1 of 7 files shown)


notapatch:

notapatch:
@matyikriszta so this is the place to review the guard clauses. In answer to your question to change a conditional to a guard clause we have to "not" the clause as described above.

app/controllers/admin/workshops_controller.rb
```diff
@@ -122,12 +122,11 @@ def workshop_id
 
   def set_host(host_id)
     return unless host_id
-
     host = @workshop.workshop_sponsors.find_or_initialize_by(sponsor_id: host_id)
-    unless @workshop.host.eql?(host.sponsor)
-      @workshop.workshop_sponsors.where(sponsor: @workshop.host).destroy_all
-      host.update(host: true)
-    end
+    return if @workshop.host.eql?(host.sponsor)
```
notapatch: @matyikriszta - If you change this from `if` to `unless` it breaks the code. You can change any of these guard clauses in the commit from `unless` to `if` or `if` to `unless` and tests break. I didn't check in a couple of the refactorings because they weren't backed by tests and I won't do a change until it is.

notapatch:
Marvelous, a new flaky test. This will keep me quiet.

PR #1142, opened 2019-11-01
[Rubocop] Freeze mutable constants

notapatch:
- without this change constants can be changed without any
   error.
 - rubocop --only Style/MutableConstant
 - also changed parenthesis to square brackets
---

[Rubocop Mutable Constant](https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/MutableConstant)

### Further Reading
[Why are there frozen constants everywhere?](https://stackoverflow.com/questions/27702540/why-are-there-frozen-constants-everywhere)

### DIFF
--- a/app/presenters/event_presenter.rb
+++ b/app/presenters/event_presenter.rb
@@ -2,7 +2,7 @@ class EventPresenter < BasePresenter
   PRESENTER = { workshop: 'WorkshopPresenter',
                 course: 'CoursePresenter',
                 meeting: 'MeetingPresenter',
-                event: 'EventPresenter' }
+                event: 'EventPresenter' }.freeze
 
   def self.decorate_collection(collection)
     collection.map { |e| PRESENTER[e.class.to_s.downcase.to_sym].constantize.new(e) }


notapatch:

app/controllers/admin/workshops_controller.rb
```diff
@@ -122,12 +122,11 @@ def workshop_id
 
   def set_host(host_id)
     return unless host_id
-
     host = @workshop.workshop_sponsors.find_or_initialize_by(sponsor_id: host_id)
-    unless @workshop.host.eql?(host.sponsor)
-      @workshop.workshop_sponsors.where(sponsor: @workshop.host).destroy_all
-      host.update(host: true)
-    end
+    return if @workshop.host.eql?(host.sponsor)
```
matyikriszta: @notapatch this doesn't seem correct to me, shouldn't this be `return unless @workshop.host.eql?(host.sponsor)`? E.g. you only want this block code to run if the sponsor is also a host.

notapatch: there's a rebase issue here ... this should only be the "Freeze mutable constants code"

notapatch:
@matyikriszta so this has become a 1 line change - it was a 2 line change but one of the freezes has been introduced by previous commit. The guard clauses are kept for a different pr. Sorry for the rebase issues.

PR #1143, opened 2019-11-02
 Fix spec: randomiser collision of active & expired titles (flaky)

notapatch:
666a56b 

 - the test relies on expired job titles not appearing on the page.
   However, the job titles of active and expired tests are created
   randomly. Any expired job title that happen to be the same as
   an active job title will cause this test to fail
 - adding expired to the test offers no new coverage.
     - active scope is tested at the model level. The jobs index
       has a single conditional
         - if active returns nothing it shows "no jobs"
         - else display the active jobs.
     - any further test adds nothing beyond these two branches
  - expired should be removed because it's causing problems and
    offering no new protection. Expired already covered by the
    scope model tests of Jobs.active.
---

93ce2f6

To make the above commit I had to make sure that the model was tested properly too. 

 Changed Job scope tests to show only what is and is not returned

 - change active to test jobs that are available and jobs that
   have expired *only*
 - change pending_or_published to test if a job is draft or
   a job is pending or published *only*
 - why?
   - scopes work through sql.
   - sql is set theory: you are in the set or you are not
   - anything more than one item is more complicated and adds no
     more coverage

To add to this commit message:

The `active` scope wasn't being tested with what is active and what is expired but instead the different Job states (draft, pending and published) which are not what active filters on (which if a job is active or expired).

What `pending_or_published` scope returned was tested but not what is not returned. The tests were simplified where possible.


notapatch:
@matyikriszta - so I said something stupid like "the test show it works" [and then the pull request fails](https://travis-ci.org/codebar/planner/builds/606214522?utm_medium=notification&utm_source=github_status).   Because that's how the gods work. Turns out it was a flaky test and this is the fix. Ready for the hardest working reviewer. There's no rush on this it's a 1 in 100 event.

Issue #1144, opened 2019-11-02
Failing test: session not created


jhsu802701:
## Description of the issue 📄
When I run the test suite, one of the tests fails.  The screen output is:
```
Failures:

  1) managing workshop attendances an admin can rsvp an invited student to the workshop
     Failure/Error: visit admin_workshop_path(workshop)
     
     Selenium::WebDriver::Error::SessionNotCreatedError:
       session not created: This version of ChromeDriver only supports Chrome version 78
     # #0 0x55967636b7e9 <unknown>
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/response.rb:72:in `assert_ok'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/response.rb:34:in `initialize'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/http/common.rb:88:in `new'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/http/common.rb:88:in `create_response'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/http/default.rb:114:in `request'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/http/common.rb:64:in `call'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/bridge.rb:167:in `execute'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/bridge.rb:102:in `create_session'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/remote/bridge.rb:56:in `handshake'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/chrome/driver.rb:43:in `initialize'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/common/driver.rb:46:in `new'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver/common/driver.rb:46:in `for'
     # /usr/local/bundle/gems/selenium-webdriver-3.142.6/lib/selenium/webdriver.rb:88:in `for'
[...]

PR #1149, opened 2019-11-03
Change Dockerfile to install google-chrome-stable

notapatch:
- change google chrome to stable version instead of an unstable newer version
- today:
  - stable chrome is   78
  - unstable chrome is 80
- depending on where we are in the chrome release cycle the drivers may not work.

---
Fixes https://github.com/codebar/planner/issues/1144

We require ChromeDriver to run our browser tests. This is currently installed by Chromedriver helper which by default installs the latest driver. 

[New ChromeDrivers are released when Chrome goes beta](https://sites.google.com/a/chromium.org/chromedriver/) - See image below: 
Screenshot 2019-11-03 at 23 32 05

So, as of today Chrome 79 is in beta and this is the latest driver.  However, Planner’s Docker image installs the [pre-beta, unstable, Chrome 80 version.](https://www.ubuntuupdates.org/ppa/google_chrome)

Screenshot 2019-11-03 at 23 37 07

We are installing a Chrome version, 80, that has no chromedriver. When we run tests they fail. Saying we don't support this version of Chrome: 

Screenshot 2019-11-03 at 23 51 12

This change installs Stable which will be covered by Chromedriver.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ RUN set -ex; \
     wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
     && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
     && apt-get update \
-    && apt-get install -y google-chrome-unstable --no-install-recommends \
+    && apt-get install -y google-chrome-stable --no-install-recommends \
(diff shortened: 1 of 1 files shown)


notapatch:
@matyikriszta - ready for review - turns out we've been installing the wrong version of Chrome - the unstable rather than stable. This would then break the tests when running them in Docker depending on the release cycle of Chrome. Details in the commit. 

Thank you @jhsu802701 for reporting the issue

jhsu802701:
This branch works for me.

Issue #1148, opened 2019-11-03
Excessive screen output during the build process


jhsu802701:
During the build process, my terminal is flooded with messages about successful database transactions.  The configuration should be changed to skip printing so many messages, because they make it harder to troubleshoot in the event that something goes wrong.

jhsu802701:
I have submitted Pull Request #1150 to resolve this issue.

notapatch:
So there's an alternative. You can have a rake file and all it does is sets the logging level to info and then call the other rake commands so db:migrate db:seed etc. That way you get lower output not ideal but not bad and it wouldn't have the downside of switching off debug for all development.

[The code looks like this.](https://github.com/notapatch/planner/pull/12/commits/56858bc3dcaa008592febb97776e2e8985097d8c) and the [output looks like this](https://gist.github.com/notapatch/8565634c81e2535d9ab3dbe3d9a7e256)

Does this look OK to you? If it does the next step is to ask the maintainers if they are interested.

biggianteye:
I agree that the screen output when building the project is excessive. It also feels quite hostile to newcomers. I agree that it needs reducing to be more useful.

The approach offered by @notapatch seems useful as it offers a way to maintain debug during application use (where it's useful) but shuts it up during setup. Can rails take in values from environment variables? I would assume yes. If so, that might be a better approach as it requires less new code.

eg. `bin/dbuild` would contain something like:

```
LOG_LEVEL="INFO" docker-compose run --rm web rake db:create db:migrate db:seed db:test:prepare
```

`config/environments/development.rb` would then be tweaked a little to use `LOG_LEVEL` if it exists.

notapatch:
I was thinking the script should try to be as much like a standard Bash script as possible:

`bin/dbuild`  - defaults to warn or higher
`bin/dbuild -h` - displays help about -i and -w etc
`bin/dbuild -i` outputs at info 
`bin/dbuild -w` outputs at warn etc

biggianteye:
> I was thinking the script should try to be as much like a standard Bash script as possible:

Sounds sensible. Though the common pattern would be silent with no params and `-v`, `-vv`, `-vvv`, etc for increasing levels of verbosity.

notapatch:
Very good `-v, -vv, -vvv`. I intend to pass or my _brilliant_ script ideas through you.  👍

notapatch:
@matyikriszta has "OKed" the work.

[...]

PR #1150, opened 2019-11-03
Changed log_level in development environment

jhsu802701:
I changed the log_level from debug to info in order to cut down on excessive screen output.  Now the build process clearly displays the deprecation warnings that were previously pushed off the terminal scrollback by gratuitous news of successful database transactions.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -34,7 +34,7 @@ Planner::Application.configure do
   config.assets.debug = true
 
   config.logger = Logger.new(STDOUT)
-  config.log_level = :debug
+  config.log_level = :info
 
   config.after_initialize do
     Bullet.enable = true


notapatch:
Hello jhsu802701 

And welcome to codebar!

I agree that the messages are distracting during the seeding and it would be good to quieten them and you've shown a way of fixing it. 

I only run seeding for a few minutes in a day. The rest of the time I'm developing. So I want to keep Rails set to something that most commonly helps me develop. Genuine question. when you're developing do you normally set logging to :info? I normally leave it at :debug.

You might be able to change Rails logging levels with ENV variables? I haven't thought it through to be honest. 🤔 

That said it's the maintainers call to make and they'll check over PR's during the week. Good luck!

jhsu802701:
I normally don't think about the log_level, but it's clearly excessive here.

matyikriszta:
@jhsu802701 I think I would also prefer if we did not turn off full logging in the development env, this is definitely useful when you are developing. If we can find a way to turn it off for seeding/building I'm ok with that if it's bothering people.

notapatch:
I managed to ignore the output as well, but, I think jhsu is right it's hiding warnings which have now become [new issues](https://github.com/codebar/planner/issues/1151). And in general, Bash commands should only output when there's a problem. 

A Thor task with minimal logging will run the same commands we do already do (db:migrate db:seed etc) and only change the logging level during setup. The task should allow for additional flags so we can change, in future PRs - for example seeding levels.  [I've done a rake version to give an idea](https://github.com/notapatch/planner/pull/12/commits/56858bc3dcaa008592febb97776e2e8985097d8c) but rake doesn't handle flags cleanly - the PR should try with Thor.  ⚡️ 🔨

I will leave the issue open if @jhsu802701 wants to do it, or anyone else.  

#### This message is copied to open issue

#### Reference
[Open Issue](https://github.com/codebar/planner/issues/1148)
[Thor - Railscasts - old but good overview](http://railscasts.com/episodes/242-thor?autoplay=true)
[Calling Rake Tasks with Thor](https://github.com/erikhuda/thor/wiki/Calling-Rake-tasks-with-Thor)
[Thor](https://github.com/erikhuda/thor)

notapatch:
@despo  - this can be closed. jhsu802701 had some good ideas but this PR which is to change _everyone's_ log level to  info was never going to fly.

despo:
@notapatch closed!

Issue #1118, opened 2019-10-22
Refactor the workshop edit to use the _form partial


jabawack81:
## Description of the issue 📄
Not a real issue but working on the issue #1104 I noticed that in the `edit` and `new` views for the workshops are using two different partials.

Comparing the two files I excluding `CSS` differences I can't see a reason for this difference:

<pre>
diff _edit_form.html.haml _form.html.haml
1c1
< = simple_form_for [:admin, @workshop], url: :admin_workshop<b>, method: :put,</b> do |f|
---
> = simple_form_for [:admin, @workshop], url: :admin_workshops do |f|
13c13
<       = f.input :host, as: :select, collection: Sponsor.all, include_blank: true, selected: (@workshop.host.id rescue '')
---
>       = f.input :host, as: :select<b>, required: true</b>, collection: Sponsor.all, include_blank: true, selected: (@workshop.host.id rescue '')
16c16
<       = f.association :sponsors
---
>       = f.association :sponsors<b>, required: true</b>
26,27c26,27
<       <b>%label Organisers</b>
<       <b>= f.collection_select :organisers, Member.all, :id, :full_name, { selected: @workshop.organisers.map(&:id) }, { multiple: true }</b>
---
>
>
</pre>

* using `PUT` on the `edit` form.
* marking `host` as required in the `edit` form
* marking `sponsors` as required in the `edit` form
* adding `organisers`  in the `edit` form

there is a reason behind this situation? if not I would like to open a PR (after my other PR, https://github.com/codebar/planner/pull/1108, is merged to avoid conflicts) to fix this.

Consideng the target odience of Codebar I think the planner could be the first webapp students look at so it should follow rails best practises as much as possible.

## Steps to fix 🛠

Update the `edit` view to use `_form.html.haml` as partial instead `_edit_form.html.haml`

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

notapatch:
When creating a new workshop it doesn't validate/error out 

```
1. http://localhost:3000/admin/workshops/new
2. Fill in nothing (not the hardest step)
3. Press Save.

[...]

PR #1155, opened 2019-11-12
Refactor the workshop edit to use the form partial

jabawack81:
working on this issue I had to disable the Foundation validation on the form because it was preventing the sending of the form so I disabled it and now it works as expected.

In reference to issue #1118


app/views/admin/workshops/_form.html.haml
```diff
@@ -35,6 +35,8 @@
         = f.input :rsvp_open_local_time, as: :string, input_html: { data: { value: @workshop.rsvp_opens_at.try(:time).try(:strftime, '%H:%M') } }
   .row
     .large-6.columns
+      %label Organisers
```
notapatch: So, my first PR to the project was accepted in September. I say this because I have no project/business knowledge to fall back on.  

We are making a change in the behaviour of the code. Before `Create` was different from `Update` and with this commit they are the same. It is unusual that we have differences in forms but it's not unheard of. So, have you understood the reasons @ + despo (an original commiter but no longer active) wrote the behaviour the way they did and allowed for that? I say this because the maintainers are going to ask.

app/views/admin/workshops/_form.html.haml
```diff
@@ -35,6 +35,8 @@
...
+      = f.collection_select :organisers, Member.all, :id, :full_name, { selected: @workshop.organisers.map(&:id) }, { multiple: true }
```
notapatch: So, my first PR to the project was accepted in September. I say this because I have no project/business knowledge to fall back on.  

We are making a change in the behaviour of the code. Before `Create` was different from `Update` and with this commit they are the same. It is unusual that we have differences in forms but it's not unheard of. So, have you understood the reasons @ + despo (an original commiter but no longer active) wrote the behaviour the way they did and allowed for that? I say this because the maintainers are going to ask.

app/views/admin/workshops/_form.html.haml
```diff
@@ -1,4 +1,4 @@
-= simple_form_for [:admin, @workshop], url: :admin_workshops do |f|
+= simple_form_for [:admin, @workshop], html: { novalidate: true } do |f|
```
notapatch: I haven't looked at this yet but I did a quick search for `novalidate` so my first question when I look is why novalidate isn't used in any other forms?

jabawack81: that is a good point argument but I've looked at the history of the two files and they were created almost identical in the same commit https://github.com/codebar/planner/commit/7ef5049753b1bafe324a6d18ca8765c12196a10e and they had a fairly parallel edit history so I'm confident there should be no problem with this PR

jabawack81:
that is a good point argument but I've looked at the history of the two files and they were created almost identical in the same commit https://github.com/codebar/planner/commit/7ef5049753b1bafe324a6d18ca8765c12196a10e and they had a fairly parallel edit history so I'm confident there should be no problem with this PR

notapatch:
Right, fairly parrell, _except_ this line so lets make sure. If you change out of the branch to master. And do git blame against that file.

`git blame app/views/admin/workshops/_edit_form.html.haml`

And roll down to that line and copy the sha and do git show on it. You'll notice that the committer is despo (who made the original commit of the project) had particular requirements when she introduced this code. 

>     default chapter organisers as workshop organisers on creation
>     when updating enable user to set new/remove organisers

So, if I was a maintainer, I would want to know that the person making this PR had seen that and knew that this code wouldn't change that - this is the only line we can get wrong :-) And because this business knowledge will disappear with the deletion of the file (to all but the most determined reader). 

So, and I haven't tried this, but with the change:

- Does a new workshop still add the chapter organisers automatically?
- Is this tested so we can't fork up?
<BR>

Because if we break that we're going to be really unpopular with the most influential people in codebar :) And after a while of hunting you down they'll work out that the [X] isn't actually a mis-loaded image and then you'll be in big trouble.

despo:
@jabawack81 due to a number of changes we had to make in a rush, to make sure we support virtual workshops, this is no longer required/applicable as the admin workshop form has been extended to include some additional fields and also extracted to a reusable partial that now both the new and edit pages use, so I will be closing this PR. Apologies we didn't get to it on time.

jabawack81:
@despo no worries at the end was a unrequired improvement done out of curiosioty.

Issue #1167, opened 2019-12-03
Use social media metadata for better link previews


biggianteye:
## Description of the issue 📄
Workshop links at the moment expand to a generic preview. This is not very useful. The picture below shows how a workshop url looks like when unfurled in Slack.
![](https://user-images.githubusercontent.com/1482649/70055875-5749ef00-15d2-11ea-95a5-e4ceeb1134e9.png)

It would be useful to make use of social media metadata to provide more information and a richer preview. The picture below shows a mock up of what a workshop url might look like in Slack. This was created using the [Slack message builder](https://api.slack.com/docs/messages/builder?msg=%7B%22attachments%22%3A%5B%7B%22author_name%22%3A%22codebar.io%22%2C%22author_link%22%3A%22https%3A%2F%2Fcodebar.io%22%2C%22author_icon%22%3A%22https%3A%2F%2Fslack-imgs.com%2F%3Fc%3D1%26o1%3Dwi32.he32.si%26url%3Dhttps%253A%252F%252Fcodebar.io%252Fassets%252Ffavicon-2781597e1d2b243721b0d14f6b2ac5d04d1ee747fe8977abd59a29f0513b41b4.ico%22%2C%22title%22%3A%22Workshop%3A%20AZA%20Finance%22%2C%22title_link%22%3A%22https%3A%2F%2Fcodebar.io%2Fworkshops%2F1555%22%2C%22text%22%3A%22Attend%20our%20workshops%20to%20learn%20programming%20in%20a%20safe%20and%20supportive%20environment%20at%20your%20own%20pace%2C%20or%20to%20share%20your%20knowledge%20and%20coach%20our%20students.%22%2C%22fields%22%3A%5B%7B%22title%22%3A%22Where%22%2C%22value%22%3A%22AZA%20Finance%2C%20Cranmer%20House%2C%2039%20Brixton%20Rd%2C%20London%2C%20SW9%206DZ%22%2C%22short%22%3Atrue%7D%2C%7B%22title%22%3A%22When%22%2C%22value%22%3A%22Tue%2C%2010th%20December%20at%2018%3A30%22%2C%22short%22%3Atrue%7D%5D%2C%22thumb_url%22%3A%22https%3A%2F%2Fassets.codebar.io%2Fb%2F%2Fuploads%2Fsponsor%2Favatar%2F775%2F_AZA_LOGO_high-67.png%22%7D%5D%7D)

[...]

PR #1169, opened 2019-12-07
Use social media for metadata for better link previews  #1167

Liatmoss:
Almost solved this issue. Some tags `og:title` and `og:description` are already set in the layout so these need to be overridden in workshop pages.


Liatmoss:
This is how it appears on Slack with the changes
Screenshot 2019-12-07 at 15 25 54

PR #1175, opened 2019-12-24
Upgrade ruby to 2.4.9

tlwr:
What
----

~The current ruby version is 2.4.2; ruby 2.4.2 is no longer supported and 2.5.7 is the latest release marked "stable" which is supported by the status checks (e.g. codecoverage)~

The current ruby version is 2.4.2; ruby 2.4.2 is no longer supported and 2.4.9 is the latest release  which is supported by rails core, the status checks (e.g. codecoverage), and Heroku

(see comments below)

How to review
----

- [x] Tests should pass

- [x] `git grep -F 2.4.2` should return no results

- [ ] Deploy to staging should not fail

- [x] Ensure that Heroku [supports ruby 2.4.9 by reviewing this changelog](https://devcenter.heroku.com/changelog-items/1687)


notapatch:
Hi @tlwr

I'm not the reviewer / maintainer. I'm just a nosy committer.

> The current ruby version is 2.4.2 ruby 2.4.2 is no longer supported

Agreed.

> and 2.5.7 is the latest release marked "stable" which is supported by the status checks

I'm a bit confused here. Heroku status checks? So we're talking about what Ruby version of 2.5 Heroku is calling stable?

This is what I would say Planner's situation is:

- [Planner is on  Rails 4.2 and Ruby 2.4.2](https://github.com/codebar/planner/blob/master/Gemfile#L2-L4).
- [Rails 4.2  was tested with 2.4.5 but not 2.5.X](https://github.com/rails/rails/commit/7a48f5869e7b9723cf94b5366ccaeb1e17060807)
- [ It was a _lot_ of work to backport Ruby 2.4 onto Rails 4.2](https://github.com/rails/rails/pull/27473)
- [I can't find any backport to Ruby 2.5 in Rails 4-2-stable  commits](https://github.com/rails/rails/commits/4-2-stable)

Given the above points. The community repo hasn't tested Rails 4.2 on Ruby 2.5. When I googled around all I have to decide is CI configurations (Travis CI used Ruby 2.4. [They have now moved CI to BuildKite which uses Ruby of 2.5](https://www.hmallett.co.uk/2018/08/ruby-and-ruby-on-rails-version-compatibility/) but I don't think it's ever been run against it).

[Rails LTS would cover Ruby 2.5 it but it's pricey. ](https://railslts.com/)

### Questions
- Do you have better information than the Rails CI configuration? 
- Have you or anyone been running Rails 4.2 on Ruby 2.5?

I am only seeing bad or expensive options at this point. I would like to be wrong from what I have found Ruby 2.5 means Rails 5.0.

### Further Reading
[Has ruby 2.5 been tested with Rails 4.2?](https://stackoverflow.com/questions/53213498/has-ruby-2-5-been-tested-with-rails-4-2)

tlwr:
You are correct, in that 2.5 is not tested by rails core; and I am incorrect in my assertion that 2.4 is EOL, the actual 2.4 EOL date is end of March 2020. With that in mind I will amend this PR to use the latest Ruby 2.4

By status checks, I am referring to:

- codeclimate
- coverage/coveralls
- travis
- snyk

where either coveralls/codeclimate (I cannot remember which) did not support Ruby 2.6 onwards.

Additionally I am looking at a 4.2 -> X upgrade path in my limited spare time

Gemfile.lock
```diff
@@ -452,7 +452,7 @@ DEPENDENCIES
   will_paginate
 
 RUBY VERSION
-   ruby 2.4.2p198
+   ruby 2.4.9p362
 
 BUNDLED WITH
-   1.17.1
+   1.17.2
```
notapatch: #### Minor correction
I notice that you've changed the bundler version from 1.17.1 to 1.17.2 was this deliberate? Bundler releases of [1.17.2 and 1.17.3 were both very minor](https://github.com/rubygems/bundler/blob/master/CHANGELOG.md#1173-2018-12-27). I'm not bothered but if intentional it should be noted somewhere in the commit.

tlwr: Good spot, this was not intentional

notapatch:
Rails 4.2 releases stopped @ `Ruby 2.4.5` and since then a few bug fixes and, mainly, security fixes (all listed in the releases link). There is no online CI working for Rails 4.2 at all now. However, I have run Rails 4.2 and Ruby 2.4.9 locally on my machine and as you would expect, given the small changes, it passed all the tests.

Ruby 2.4.2 is too old to be running on. I think Planner should be running on Ruby 2.4.9 and not 2.4.2 👍 

I don't have any merge permissions so all I can wish you is good luck!

### Releases
[Ruby-2.4.6 - bugfixes and security](https://www.ruby-lang.org/en/news/2019/04/01/ruby-2-4-6-released/)
[Ruby-2.4.7 - security only](https://www.ruby-lang.org/en/news/2019/08/28/ruby-2-4-7-released/)
[Ruby-2.4.8 - security only](https://www.ruby-lang.org/en/news/2019/10/01/ruby-2-4-8-released/)
[Ruby-2.4.9 - security only](https://www.ruby-lang.org/en/news/2019/10/02/ruby-2-4-9-released/)

tlwr:
Thank you 👍 :)

The reason I raised this PR is because at work we're deprecating earlier versions of ruby and I imagine Heroku will do the same (although I haven't got any information about a date); and because I wanted to raise a feature but I didn't have 2.4.2 on my laptop :)

tlwr:
AFAICT the only salient changes between 2.4.5 and 2.4.9 are security vulnerabilities, I am not particularly concerned about rails and other dependencies breaking with changes from a point release

I also don't think that we will find many people who are using an old version of rails with a security patched version of ruby, because people who are likely to security patch ruby are also likely to upgrade rails before EOL (i.e. when 6 is released)

If there isn't appetite to upgrade past 2.4.5 then there is probably no point in upgrading (past 2.4.2) at all

notapatch:
First, I would also take this time to apologise to @despo as I hadn't realised that she was developing again - clearly the fact that you had been assigned to review should have been a good tip off. After this reply I'll do something fairly remarkable by my standards, I'll shut up.

---
Toby,

So this is a conversation between two people without merge powers about an unusual subject 
(how to handle a lack of security support for the current Ruby and Rails versions). If we want to get the commit through then we need to summarize the situation as best we can and leave it to the owners.

 >If there isn't appetite to upgrade past 2.4.5 then there is probably no point in upgrading (past 2.4.2) at all

I disagree and would say there are a number of beneficial bug fixes and security changes between Ruby 2.4.2 and 2.4.5 and being on a more common Ruby is a positive and I would then use 2.4.5 as a base to move to the next Ruby version 2.4.9 _but_ in the end that's your PR - so your view trumps mine and the owners views Trumps both of us.

So do you agree with the summary? You can cut and paste it and edit it as you wish - but there should be one summary to aid the owners decision or dialog with you - because it's an unusual subject. Then we have to leave the PR. It might be a few days it might be many weeks before anything happens and it's not a schedule that we can affect.

---

# Ruby Upgrade Summary

| Ruby version | Pros | Cons   |
|:------|:------|:------|
| 2.4.2  | Nothing to do | - Unusual Ruby <BR> - Missing bug fixes & security patches of later Rubies |
| 2.4.5 |  - Last Ruby tested with Rails 4.2 <br> - standard | Missing security patches of later Rubies |
| 2.4.9 | - Current Ruby version <BR> -  Verification of Rails 4.2 working with Ruby 2.4.9 by @notapatch on a local machine | - Very similar to 2.4.5 but [20 bugfixes](https://www.ruby-lang.org/en/news/2019/04/01/ruby-2-4-6-released/) and a few security fixes since 2.4.5  <BR> - No cloud CI of Rails 4.2 and Ruby 2.4.9 without work by codebar |
| 2.5.X |  - | Not ported to Rails 4.2 |

### Release Reference
[Ruby 2.4.3 bugfixes and security](https://www.ruby-lang.org/en/news/2017/12/14/ruby-2-4-3-released/)
[Ruby 2.4.4 bugfixes and security](https://www.ruby-lang.org/en/news/2018/03/28/ruby-2-4-4-released/)
[Ruby 2.4.5 bugfixes and security](https://www.ruby-lang.org/en/news/2018/10/17/ruby-2-4-5-released/)
[Ruby-2.4.6 bugfixes and security](https://www.ruby-lang.org/en/news/2019/04/01/ruby-2-4-6-released/)
[Ruby-2.4.7 security only](https://www.ruby-lang.org/en/news/2019/08/28/ruby-2-4-7-released/)
[Ruby-2.4.8 security only](https://www.ruby-lang.org/en/news/2019/10/01/ruby-2-4-8-released/)
[Ruby-2.4.9 security only](https://www.ruby-lang.org/en/news/2019/10/02/ruby-2-4-9-released/)

PR #1182, opened 2020-01-22
Accept CoC as Terms and Conditions

despo:
New and existing members must accept codebar's CoC before they can continue to the website to either complete their registration or get to any page that requires authentication.

This excludes information pages (everything under dashboard besides /dashboard and /, so they should be able to get to FAQ, Coach and Student guide, Coc, Coaches list)

### DIFF
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -11,8 +11,11 @@ class MembersController < ApplicationController
   end
 
   def step1
-    @suppress_notices = true
+    store_path and return redirect_to(terms_and_conditions_path) if current_user.accepted_toc_at.blank?
+
     @member = current_user
+    @suppress_notices = true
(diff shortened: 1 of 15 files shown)


app/controllers/members_controller.rb
```diff
@@ -11,8 +11,11 @@ def edit
   end
 
   def step1
-    @suppress_notices = true
+    accept_terms
+
     @member = current_user
+    @suppress_notices = true
```
matyikriszta: @despo was just wondering, why do we need this here?

app/controllers/members_controller.rb
```diff
@@ -11,8 +11,11 @@ def edit
...
+    flash[notice] = 'Thanks for signing up. Please fill in your details to complete the registration process.'
```
matyikriszta: @despo can this be moved into locales so it support translations etc?

PR #1188, opened 2020-02-04
/s/no dumb questions/no bad questions

alexwlchan:
The word "dumb" has ableist roots.  The phrase "bad question" expresses the same sentiment, but is less likely to make somebody feel uncomfortable.


KimberleyCook:
@alexwlchan thank you very much for this update.

Issue #1163, opened 2019-11-22
Chapter organisers should not see Review Jobs and View Feedback in menu


KimberleyCook:
## Description of the issue 📄
Chapter organisers should not see "View Feedback" and "Review Jobs" in the menu. Only admins should see this.

## Screenshots 📷

![image (1)](https://user-images.githubusercontent.com/2683270/69439941-d4a67180-0d3f-11ea-9d7a-3b43c0656570.png)

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

shen-sat:
Hi @KimberleyCook, I'd like to take on this issue if possible? I've contributed to open source once bfeore, and I've been learning to use Rails (for over 3 months) in my current job. I've not worked on on any authorisation features before, but keen to jump in :)

KimberleyCook:
@shen-sat please do :)

despo:
Resolved by attached PR

shen-sat:
Apologies for not working on this PR. I signed up to a couple of other OS projects and wasn't realistic with my time management. I should have given you a heads up :(

PR #1193, opened 2020-02-20
(feature) restrict access to admin/jobs and admin/feedback

despo:
to members with the admin role

As the admin/application_controller checks if a user is either an admin or an organiser for now add a super_admin/application_controller.

This could benefit from additional refactoring so there is better separation of what the organiser and admin roles can do. The tests should ideally also be refactored to reflect this.

Tackles issue #1163


app/controllers/admin/feedback_controller.rb
```diff
@@ -1,4 +1,4 @@
-class Admin::FeedbackController < Admin::ApplicationController
+class Admin::FeedbackController < SuperAdmin::ApplicationController
```
matyikriszta: @despo does this mean we don't want to give organizers access to feedbacks?

despo: They can still access the feedback that has been given for their chapter specifically.

matyikriszta: @despo ok, cool

app/controllers/coach/feedback_controller.rb
```diff
@@ -1,14 +0,0 @@
-class Coach::FeedbackController < Coach::ApplicationController
```
matyikriszta: @despo does this mean we also removed feedback access for coaches?

despo: It does. I think it's something we can clarify and revisit if necessary as we still get names posted and we don't just render feedback that was specifically left for the coach viewing the page.

despo: In addition to this the link was never explicitly exposed to coaches (it was part of the organiser menu) so not many people are/were aware and have ever access this.

matyikriszta: @despo ok, sure

PR #1194, opened 2020-02-20
Feature/add global mailing list

despo:


app/controllers/member/details_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  include MailingListConcerns
+
+  before_action :set_member
+  before_action :suppress_notices
+
+  def edit
+    accept_terms
+
+    flash[notice] = I18n.t('notifications.signing_up')
+    @member.newsletter ||= true
```
matyikriszta: @despo can you expand on what this line does? I'm not quite sure why this is here.

despo: It checks the newsletter checkbox by default. If they uncheck then they wont be added to the newsletter, if they leave it as is then they will be subscribed.

app/views/member/details/edit.html.haml
```diff
@@ -0,0 +1,37 @@
+%section#banner
```
matyikriszta: @despo can all the strings in this view be moved into translation files?

despo: Done

app/controllers/member/details_controller.rb
matyikriszta: @despo ok, it's what I thought but wanted to double check.

lib/tasks/mailing_list.rake
```diff
@@ -1,32 +1,15 @@
...
-    groups.each do |group|
-      mailing_list = MailingList.new(group.mailing_list_id)
+    members = Member.includes(:subscriptions).where.not('subscriptions.member_id' => nil).uniq
+    newsletter = MailingList.new(newsletter_id)
 
-      group.members.each do |member|
-        puts mailing_list.subscribe(member.email, member.name, member.surname)
-      end
+    members.each do |member|
+      newsletter.subscribe(member.email, member.name, member.surname)
```
despo: Todo before merge: 
I missed checking and setting the `opt_in_newsletter_at` timestamp. 
This will also enable members to unsubscribe from the subscription page..

PR #1218, opened 2020-03-12
(fix) flaky meeting_preseter_spec sql does not guarantee order

notapatch:
- "organisers" test expects that the array order is the same
   however SQL only guarantees the returned rows and not the
   order they are returned

Fixes this flaky test:
https://travis-ci.org/github/codebar/planner/builds/661635502?utm_source=github_status&utm_medium=notification


despo:
@notapatch thank you, we could do with a more stable built! Happy to merge when you rebase.

PR #1219, opened 2020-03-13
Rubocop fixes for Codeclimate/Rubocop integration

notapatch:
These are the Rubocop errors that were flagged up when I switched on codeclimate/Rubocop integration - [and have already been reviewed.](https://github.com/codebar/planner/pull/1046)  I do not know why these were over any other. However, they need to be fixed and they can be put in separately to the actual codeclimate/rubocop integration

Most were autocorrected the exceptions are the last two commits

865199a  and 233a0c5


notapatch:
Thanks @matyikriszta - I believe I have rebased

PR #1220, opened 2020-03-13
(feature) virtual workshops

despo:
Add support for virtual workshops

- [x] Add VirtualWorkshopPresenter to make handling workshop views easier
- [x] Display a more descriptive description for virtual workshop 
- [x] Add fields to store slack channel and slack channel link for virtual workshops (updates admin workshop forms, render information on admin workshop page and RSVPed workshop invitation)
- [x] Adds `student_spaces` and `coach_spaces` on workshop model and utilise those when the workshop is set to virtual instead of host spots
- [x] Update workshop admin page so it's clear to the organisers that this is a virtual workshop
- [x] Add virtual workshop specific emails: student & coach invitations, attending confirmation, getting spot from waiting list, attendance reminder


app/controllers/admin/workshops_controller.rb
```diff
@@ -39,32 +39,34 @@ def edit
...
 
-  def send_invites
-  end
+  def send_invites; end
 
   def attendees_checklist
     return render text: @workshop.attendees_checklist if request.format.text?
 
-    redirect_to admin_workshop_path(@workshop)
+    redirect_to admin_workshop_path(@workshop), notice: "The requested format is invalid: #{request.format}"
```
despo: The notice message should be extracted to the locale

app/views/admin/workshops/_invitation_management.html.haml
```diff
@@ -25,15 +25,15 @@
     .medium-6.columns
       %h4
         Students
-        %label.label.round.secondary #{@attending_students.count}#{(" of " + @workshop.host.seats.to_s) if @workshop.has_host?}
+        %label.label.round.secondary #{@attending_students.count} of #{@workshop.student_spaces}
```
despo: I'm not sure why we need a variable for `@attending_students`. Investigate and remove.

app/views/admin/workshops/_invitation_management.html.haml
```diff
@@ -25,15 +25,15 @@
...
         %label.label.round.signed-in #{ @attending_students.count(&:attended) }
       = render partial: 'admin/workshop/attendances', locals: { invitations: @attending_students }
       = render partial: 'admin/workshop/waiting_list', locals: { waiting_list: @student_waiting_list }
     .medium-6.columns
       .row
         %h4
           Coaches
-          %label.label.round.secondary #{@attending_coaches.count}#{(" of " + @workshop.host.coach_spots.to_s) if @workshop.has_host?}
+          %label.label.round.secondary #{@attending_coaches.count} of #{@workshop.coach_spaces}
```
despo: I'm not sure why we need a variable for `@attending_coaches`. Investigate and remove.

app/views/admin/workshops/_shared_form.html.haml
```diff
@@ -0,0 +1,34 @@
...
+    %p
+      %small= t("workshops.lead")
+.row
+  .large-12.columns
+    = f.input :description, as: :text
+%br
+.row
+  .large-12.columns
+    .panel.callout
+      %p Fill out these fields if you want RSVPs to open automatically at a future time. Leave blank if you'll handle it manually later.
```
despo: Extract text to locale

app/views/admin/workshops/show.html.haml
```diff
@@ -33,31 +33,50 @@
...
+          %strong.label.round.primary #{@workshop.student_spaces} student spots, #{@workshop.coach_spaces} coach spots
           - unless @workshop.spaces?
             %label.label.round.warning Full
-          %p.description
-            %br
-            = sanitize(@workshop.description)
+        - if !@workshop.invitable?
+          %br
+          %p
+            %label.label.info The workshop is not invitable
```
despo: Extract text to locale

app/views/virtual_workshops/_meta_tags.html.haml
```diff
@@ -0,0 +1,17 @@
+- content_for :meta_tags do
+  - title = "Virtual Workshop for #{workshop.chapter.name}"
```
despo: Extract title to locale (or reuse existing entry from locale as I think it may be identical)

app/views/admin/workshops/_invitation_management.html.haml
despo: This is a decorated collection and used elsewhere, so I'd rather not change it on this PR.

app/views/admin/workshops/_invitation_management.html.haml
despo: This is a decorated collection and used elsewhere, so I'd rather not change it on this PR.

app/controllers/admin/workshops_controller.rb
```diff
@@ -162,4 +170,20 @@ def workshop_created_within_specific_time_frame?
     Time.zone.now.between?(@workshop.created_at,
                            @workshop.created_at + WORKSHOP_DELETION_TIME_FRAME_SINCE_CREATION)
   end
+
+  def physical_workshop_and_no_host?(workshop)
+    !workshop.virtual && host_id.empty?
+  end
+
+  def no_host_error
```
matyikriszta: @despo this method is suuper confusing, especially in the way we are using it above. We are basically saying if there is a an error update the workshop and show a success message. I find that very confusing.

app/controllers/concerns/invitation_controller_concerns.rb
```diff
@@ -38,6 +37,7 @@ def accept
     end
 
     def reject
+      @workshop = WorkshopPresenter.decorate(@invitation.parent)
```
matyikriszta: @despo any reason why we are not using `@invitation.workshop` here?

app/controllers/workshops_controller.rb
```diff
@@ -32,4 +33,18 @@ def set_workshop
...
+    WorkshopInvitation.find_by(workshop: workshop, member: user, attending: true)
+  end
+
+  def find_or_create_invitation(workshop, user, role)
+    WorkshopInvitation.find_or_create_by(workshop: workshop,
+                                         member: user,
+                                         role: role)
+  end
+
+  def user_attending_or_waitinglisted?(workshop, user)
```
matyikriszta: @despo can we rename this method to `user_attending_or_waitlisted?`? Waitinglisted sounds a listen awkward.

app/presenters/workshop_presenter.rb
```diff
@@ -43,6 +64,16 @@ def time
     I18n.l(model.time, format: :time)
   end
 
+  def end_time
+    I18n.l(model.ends_at, format: :time)
+  end
+
+  def start_and_end_time
+    start_and_end_time = time
```
matyikriszta: @despo can we rename the `time` method to `start_time` to make it a bit more explicit what is happening here?

app/controllers/admin/workshops_controller.rb
despo: Agreed. It's meant to read no (host error) which is definitely not obvious.

app/controllers/concerns/invitation_controller_concerns.rb
despo: I *think* the idea when InvitationControllerConcerns was first implemented was for it to be reused amongst different types of events, but that no longer seems to be the case as the only other model that uses if (CourseInvitation) is no longer used in the app, so I will remove the use of `parent` here.

see here:. https://github.com/codebar/planner/blob/5bc08f09196b91b210c1899205424d2d5a791768/app/models 
https://github.com/codebar/planner/blob/master/app/models/workshop_invitation.rb#L30/course_invitation.rb#L9

app/controllers/workshops_controller.rb
despo: Ok!

app/presenters/workshop_presenter.rb
despo: After having a better look at this I'm not sure I want to make this change as part of this PR because
- `time` is used across the Event and WorkshopPresenter as it's the expected field in other event related models as well (Workshop, Event and Meeting to be precise), so to change it here I need to make a lot of additional changes to support this
- `time` should first be renamed in the DateTimeConcerns module, but as that is used across other models as well the work to verify that is behaves correctly everywhere is a lot
- the method is already used around a number of places app and also in shared event  related partials
- time is derived from the Workshop model's `date_and_time` so it would probably make sense to tackle from this level and rename the field to be `start_date_and_time`

As there are way too many changes to be made to resolve this are you happy with me opening an issue so this can be tackled separately?

matyikriszta: @despo I'm happy with you opening an issue for this to be tackled separately. Should have brought this up when `end_time` was introduced, but only noticed it here.

despo: @matyikriszta  New issue: https://github.com/codebar/planner/issues/1224

despo:
@matyikriszta that's the plan!

PR #1221, opened 2020-03-14
ApplicationHelper contact_email without instance variable

notapatch:
So relying on instance variables (in this case @workshop) to change a conditional makes code hard to follow and re-use - it "feels" like adding a global to the application.

If @workshop was defined as an instance variable the code does one thing and if it isn't it does another. So I added a workshop parameter to the method  and worked through the codebase passing @workshop wherever I found it defined in controller or view code.

---

This is part of the Rubocop errors that occur when [Rubocop / codeclimate integration setup is changed.  ](https://github.com/codebar/planner/pull/1046). However, this PR stands on it's own.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -40,8 +40,8 @@ module ApplicationHelper
     Planner::Application.config.twitter_id
   end
 
-  def contact_email
-    @contact_email ||= @workshop.present? ? @workshop.chapter.email : 'hello@codebar.io'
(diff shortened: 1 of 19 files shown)


notapatch:
@matyikriszta thank you for reviewing it! It's a chore bit of code - it required making a list and ticking each item off. The rebase brought in VirtualWorkshop code which increased the number of items.

I have also added tests for the none-workshop items.

notapatch:
:shipit:

PR #1225, opened 2020-03-17
(feature) add pairing csv

despo:
Relies on #1220 being merged (so only look at last 3 commits for now)

Attempting to make pairing a little bit easier by providing student and coach details on a CSV the organisers can download. As `Labels` returns a CSV with all attendees and their roles as dictated by the service we use to generate the name labels, I decided to expose a new file rather than to modify the existing one. 

It should also give an incentive to ask coaches to make sure their skills are up to date in their codebar profile.

pairing_csv

Example import into Google spreadsheets:
imported_csv

### DIFF
--- a/app/presenters/member_presenter.rb
+++ b/app/presenters/member_presenter.rb
@@ -18,4 +18,18 @@ class MemberPresenter < BasePresenter
   def subscribed_to_newsletter?
     opt_in_newsletter_at.present?
   end
+
+  def pairing_details_array(role, note)
+    role.eql?('Coach') ? coach_pairing_details : student_pairing_details(note)
+  end
+
+  private
+
+  def coach_pairing_details
+    [newbie?, full_name, 'Coach', 'N/A', skill_list.to_s]
+  end
+
+  def student_pairing_details(note)
+    [newbie?, full_name, 'Student', note, 'N/A']
+  end
 end
--- a/app/presenters/workshop_presenter.rb
+++ b/app/presenters/workshop_presenter.rb
@@ -3,6 +3,8 @@ class WorkshopPresenter < EventPresenter
   include ActionView::Context
   include ActionView::Helpers::DateHelper
 
+  PAIRING_HEADINGS = ['New attendee', 'Name', 'Role', 'Study note', 'Skills'].freeze
+
   def self.decorate(workshop)
     return VirtualWorkshopPresenter.new(workshop) if workshop.virtual?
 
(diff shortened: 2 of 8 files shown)


app/presenters/member_presenter.rb
```diff
@@ -18,4 +18,18 @@ def attending?(event)
...
   end
+
+  def pairing_details_array(role, note)
+    role.eql?('Coach') ? coach_pairing_details : student_pairing_details(note)
+  end
+
+  private
+
+  def coach_pairing_details
+    [newbie?, full_name, 'Coach', 'N/A', skill_list.to_s]
```
Rosa-Fox: guessing `full_name` and `skill_list` work due to being defined on `member`?

Should `skill_list.to_s` be `skill_list.join`?

app/presenters/workshop_presenter.rb
```diff
@@ -3,6 +3,8 @@ class WorkshopPresenter < EventPresenter
   include ActionView::Context
   include ActionView::Helpers::DateHelper
 
+  PAIRING_HEADINGS = ['New attendee', 'Name', 'Role', 'Study note', 'Skills'].freeze
```
Rosa-Fox: 👍

app/views/admin/workshops/show.html.haml
```diff
@@ -14,7 +14,10 @@
     = link_to '#', class: 'item disabled', title: t('admin.workshop.not_invitable.') do
       %i.fa.fa-send-o
       %label Invite
-  = link_to admin_workshop_path(format: 'csv'), class: 'item', title: 'CSV for labels' do
+  = link_to admin_workshop_path(format: 'csv'), class: 'item', title: 'CSV for pairing' do
```
Rosa-Fox: 👍

spec/features/admin/workshops_spec.rb
```diff
@@ -193,5 +193,29 @@
...
+
+    context 'attendee CSV' do
+      it 'returns a CSV with all workshop attendees' do
+        workshop = Fabricate(:workshop)
+        visit admin_workshop_path(workshop)
+        click_on 'Pairing CSV'
+
+        expect(page.current_path).to eq(admin_workshop_path(workshop, format: 'csv'))
+        expect(page).to have_content(WorkshopPresenter::PAIRING_HEADINGS.join(','))
+        expect(page).to_not have_content('ORGANISER')
```
Rosa-Fox: should this be `not_to` rather than `to_not`? Guessing it is fine but haven't run locally and usually use `not_to`!

app/presenters/member_presenter.rb
despo: Yes, they are delegated to Member.

`skill_list` is a taggable object so `to_s` will  return a comma separated list of all skills.

spec/features/admin/workshops_spec.rb
despo: That does read a bit better so happy to make the change. I'm just accustomed to using `to_not`!

despo:
@Rosa-Fox thank you for reviewing this 💯

Issue #1204, opened 2020-03-08
Remove `humanize_method` from ApplicationHelper and let presenters/locale handle date formatting


despo:
## Description of the issue 📄

Quoting from https://github.com/codebar/planner/pull/1075#issuecomment-554765580

> When I saw the original issue I did look over humanize_date and played around with it but I thought "scary method" and did not try to create a PR for the issue. Methods should have one purpose but it is trying to be a date, datetime, different formats within all these demands. Further, it hard codes an order of day and month which Americans, and no doubt others, will not like.
> 
> I think date variations should be handled directly by i18n.
> 
> i18n won't handle [ordinalizing numbers natively until Rails 6.0](https://github.com/rails/rails/pull/32168) - if you decide you don't need an ordinalized number in an admin list of dates then you have workshop format that already covers what you need (it isn't used anywhere at all at the moment and probably should be renamed after the data it is formatting not where it is used - the workshop). So the solution would be [something like this](https://github.com/notapatch/planner/pull/14/commits/82c894a9624e8d920349cf56804a11af48eb5322).
> 
> If we can't wait for Rails 6 and we want ordinalize now a [solution something like this](https://github.com/notapatch/planner/pull/14/commits/24d40e9aea16f9ae4431b1022034d49e83688e2a). I'm not sure how that works with the other en-GB, en-AU formats as I only ever got en?
> 
> Other than that, the date formatting and the test is at the wrong level - feature tests are slow and we always display a date. The formatting and testing should be at the presenter level which is fast - and if you want to test across different localisations you could end up with running 5 tests, one for each language.
> 
> ```ruby
> # untested
> class WorkshopPresenter
>   def day_and_date
>      I18n.l(model.date_and_time, format: :workshop)
>   end
> end
> ```

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

despo:
Some aspects of this have been unintentionally tackled by #1272 so this issue need sto be reviewed and some of the changes made in #1231 integrated after #1276 is merged.

PR #1231, opened 2020-03-22
[WIP] Pr format humanize date

notapatch:
Closes https://github.com/codebar/planner/issues/1204

Introduces an I18n date format to replace the helper method humanize_date

| Format | humanize_date call                                                                                                      |
|:--------|:---------------------------------------------------------------------------------|
| humanize_date | humanize_date(date, with_time: false, with_year: false)                              |
| humanize_date_with_year  |  humanize_date(date, with_time: false, with_year: true)             |
| humanize_date_with_time |  humanize_date(date, with_time: true, with_year: false)              |
| humanize_date_with_year_with_time  |  humanize_date(date, with_time: true, with_year: true) |

- Commits put together the code change with the associated test. 
- The format should match the humanize_date call date (and times)
  - with the exception of 2f751ce which I thought might be wrong
  - one of the tests got away with testing for date when it should have been date and time - the test was changed to date and time

462ff8d introduced `humanize_date_with_time_range` - which allowed me to replace and test the ternary operator used in many views for date time ranges (see 462ff8d for examples) . 

Testing put `format: :humanize_date` on both the specs and the production code. I'm not sure how much value testing a format is outside of branch logic - as long as you have literally used the right text it works. When ordinalization is supported natively in rails 6 there's less to go wrong. However, I've kept all the testing and added a few more for consistency.

d3a7316  - did not have direct tests but if you got the formatting wrong it would break the build. Writing a test to confirm the formatting did not make sense as it was no code logic and there wasn't a test for anything else in the view. 

c0bf1f9  - whatever I did in these files would not break the build. I did stay away from a few files where I couldn't work out where to test them / if they were tested. 🐔

### DIFF
--- a/app/controllers/admin/portal_controller.rb
+++ b/app/controllers/admin/portal_controller.rb
@@ -4,7 +4,7 @@ class Admin::PortalController < Admin::ApplicationController
 
     @jobs_pending_approval = Job.where(approved: false, submitted: true).count
     @chapters = Chapter.active.all.order(name: :asc)
-    @workshops = Workshop.upcoming
+    @workshops = Workshop.includes(:chapter).upcoming
     @groups = Group.joins(:chapter).merge(@chapters)
     @subscribers = Subscription.joins(:chapter).merge(@chapters)
                                .ordered.limit(20).includes(:member, :group)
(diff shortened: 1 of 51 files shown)


notapatch:
- WIP review codeclimate issues

notapatch:
The problem with refactoring mailers is that [Rails 5.1 parameterization](https://github.com/rails/rails/pull/27825) means that when you move to Rails 5.1 you should undo the refactor and then try and apply Rails parameterization (It looks interesting but until you have actually applied it you never know). In the mean time I've added refactors

notapatch:
[Requires this to be closed to fix the outstanding codeclimate issue](https://github.com/codebar/planner/pull/1238)

PR #1238, opened 2020-03-27
Fix Assignment Branch condition size with Admin portal presenter

notapatch:
> Assignment Branch Condition size for index is too high.

I was getting the above Rubocop error (above) when I made a modest change to one of the assignments in the PortalController#index method [as part of another PR](https://github.com/codebar/planner/pull/1231). These changes should fix this.

```ruby
  def index
    authorize :admin_portal

    @jobs_pending_approval = Job.where(approved: false, submitted: true).count
    @chapters = Chapter.active.all.order(name: :asc)
    @workshops = Workshop.upcoming
    @groups = Group.joins(:chapter).merge(@chapters)
    @subscribers = Subscription.joins(:chapter).merge(@chapters)
                               .ordered.limit(20).includes(:member, :group)
  end
```
1. [Minor fix to the Chapter Fabricator that was duplicating the config code](d0df65e )
2. [Support method to disable bullet on demand](https://github.com/codebar/planner/commit/de1436a1b6591c207a0135a26dd8fcde1b089121)
    - Needed during portal assignment testing (below) because I was not using group data returned in the subscribers method Bullet considered it was an eager loading issue.
3. [Add Presenter object which wraps up Portal assignments and tests each assignment.](https://github.com/codebar/planner/commit/fd0b9aa13b34c1318d564bff1882b4e6f3fcfa3c )
4. (fix) [Switched the code to the Portal object away from the multiple instance variables](https://github.com/codebar/planner/commit/7452945a65692df9b13547b4274b9cc8958791fc)

### DIFF
--- a/app/controllers/admin/portal_controller.rb
+++ b/app/controllers/admin/portal_controller.rb
@@ -2,12 +2,7 @@ class Admin::PortalController < Admin::ApplicationController
   def index
     authorize :admin_portal
 
-    @jobs_pending_approval = Job.where(approved: false, submitted: true).count
-    @chapters = Chapter.active.all.order(name: :asc)
-    @workshops = Workshop.upcoming
-    @groups = Group.joins(:chapter).merge(@chapters)
-    @subscribers = Subscription.joins(:chapter).merge(@chapters)
-                               .ordered.limit(20).includes(:member, :group)
+    @portal = AdminPortalPresenter.new
   end
 
   def guide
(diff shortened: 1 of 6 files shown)


despo:
@notapatch presenters should be used to format already retrieved data and not to make database calls. Moving the AR calls in the presenter masks instead of resolving the rubocop issue.

notapatch:
That's some very well tested but wrong code. Kill the PR?

despo:
@notapatch it is indeed and maybe that's the best for now. I wouldn't even know how to start with this and maybe what we need to try and figure out first with the other admins is if we do need all these information rendered on this page, as I think we only utilise a very small portion of it.

PR #1245, opened 2020-04-08
(Refactor) Remove @contact_email instance variable

notapatch:
- variable only used for memoization
 - only called once per request/response cycle and very light work
     - no obvious reason to memoize
---

This is another rubocop fix but you won't see the "error" unless you turn on `rubocop-rails` [which you can see in this PR.](https://github.com/codebar/planner/pull/1046)


notapatch:
I'm going to kill this. While it fixes the "issue" I don't get it because surely that means it's no longer memorizing it.

PR #1246, opened 2020-04-08
(Fix flaky) attendance_reminder match short text to avoid newline

notapatch:
- currently expect matches a long line of static text (there
   is no condition for the text to appear or not)
 - What happens is that it puts carriage returns =0D in the middle
   of the text stopping the match.
 - I have never got to the bottom of this problem (it only happens
   sometimes) and only on Travis.
 - It doesn't seem like a useful test, I think it is a way to make
   sure the expected template was used - so instead I will match
   to a shorter email heading.
--
[Example of the flaky test](https://travis-ci.org/github/codebar/planner/builds/672490520)

### DIFF
--- a/spec/mailers/meeting_invitation_mailer_spec.rb
+++ b/spec/mailers/meeting_invitation_mailer_spec.rb
@@ -63,7 +63,7 @@ RSpec.describe MeetingInvitationMailer, type: :mailer do
     end
 
     it 'renders the body' do
-      expect(mail.body.encoded).to match('This is a quick email to remind you that you have signed up for tomorrow\'s codebar Monthly')
+      expect(mail.body.encoded).to match('codebar Monthly Attendance Reminder')
       expect(mail.body.encoded).to match('hello@codebar.io')
     end
   end


notapatch:
Haha ... touch one thing and break another. However, in this case we have seen matches of long email strings result in these occasional breaks. This change will make the build more robust.

notapatch:
rebase!

despo:
@notapatch @travis-ci is misbehaving today 😠

notapatch:
Hey, there's always options @despo. I am sitting on a Github actions that works with codebar ... it does the build about 10s quicker. You would have to use Ruby `2.4.10` though. Also Circle CI is similar to Github actions.

despo:
@notapatch there is a long list of other things that have to be done and other than today I'm overall quite happy with them. But definitely something to consider if this becomes a repeat issue.
What frustrates me the most is that the issue is not reflected on their status page https://www.traviscistatus.com/ but we have 3 PRs pending a status update so there must be something causing a delay in their responses.

notapatch:
I've seen "Expected - Waiting for status to be reported" quite a few times. 🤷‍♂️

PR #1252, opened 2020-04-17
(chore) add hinttext to clarify use of `invitable`

despo:
on workshop admin page

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -603,6 +603,7 @@ en:
         description: 'Supports HTML'
         coach_spaces: Available coach spaces
         student_spaces: Available student spaces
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -603,6 +603,7 @@ en:
         description: 'Supports HTML'
         coach_spaces: Available coach spaces
         student_spaces: Available student spaces
+        invitable: When checked, it overrides the RSVP open local date and time.
```
matyikriszta, line 606: @despo I think we need to be more explicit here and say that this checkbox overrides the RSVP open local date and time and opens RSVP's. I think we need to mention that it actually opens RSVP's right away.

despo: updated text

Issue #1070, opened 2019-10-02
Admin member workshop history shows future events


biggianteye:
## Description of the issue 📄

In the admin interface I was looking at the workshop history for a member to see their "no shows". The screenshot below shows two problems:

- The wording about their most recent workshop says
> The last workshop they attended was on 08 Oct 2019

but this is inaccurate as this is a future event (it's October 2nd today) and therefore they haven't attended it.
- The future event is flagged as a "no show" which negatively misrepresents this person's attendance history.

## Screenshots 📷

![Screenshot of member workshop history](https://user-images.githubusercontent.com/1482649/66036216-98575180-e504-11e9-8d73-20c7c1efdbfc.png)

## Steps to fix 🛠

- Ensure the "last attended" text only uses past events.
- Either a) don't show future events in the workshop list or b) show it and omit the symbol indicating whether or not they attended.

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

notapatch:
Thanks for reporting the error it was really clear to understand with the screen print and I can confirm the bug. I've added some notes for myself, or the next programmer...

> Ensure the "last attended" text only uses past events.

The problems come from the [controller useing accepted_or_attended](https://github.com/codebar/planner/blob/master/app/controllers/admin/members_controller.rb#L10)

 ```
   def show
    ...
    @invitations = @member.workshop_invitations.accepted_or_attended  ... continues
    ...
    @last_attendance = @invitations.first.workshop if @invitations.any?
   end
```
accepted_or_attended returns members's workshop invitations where they are [attending or attended](https://github.com/codebar/planner/blob/master/app/models/concerns/invitation_concerns.rb#L14). The names are confusing and this has confused the programmer ( accepted  equals attending *scratches head* ).  So changing @last_attendance to just `accepted` would fix that problem.

Onto the table problem: 

>  a) don't show future events in the workshop

The easiest change would be to just use the attended scope as discussed in the previous problem. 

>  b) show it and omit the symbol indicating whether or not they attended

I would not recommend a missing symbol because no content can lead to formatting problems - instead a calendar icon next to attending. 

[...]

PR #1270, opened 2020-05-11
Ensure rsvp_time is always logged

### LINKED ISSUE #1156: Attendance list does not show all RSVP times
## Description of the issue 📄
The attendance list visible in the admin interface (`https://codebar.io/admin/workshops/<id>`) shows RSVP times for both students and coaches. The list is incomplete though. Only some entries have dates and it is unclear why.

I've put my hypothesis about why this might happen under the screenshot because it requires that context.

## Screenshots 📷
This screenshot shows the attendance list for a workshop that took place on 12th November and was set up on 5th November. It demonstrates that some students have RSVP times and some don't.

image

Not shown above is the fact that none of the waitlist entries have RSVP times listed. This observation combined with the fact that the only dates shown are on the day the event created gives me a hypothesis:

> The people who don't have an RSVP time are people who have been added from the waitlist.

Note that this hypothesis is disproven for coaches because the coach list for this event had a coach without a date and there was no waitlist for coaches in this instance. I don't know if coaches and students are handled differently, so the hypothesis may still be valid.

## Steps to fix 🛠
Without knowing the source of the problem, it's hard to say. Assuming my hypothesis is correct, keeping track of RSVP times for people in the waitlist would probably fix this.

## To do 📋

* [ ] Submit a pull request

despo:
- adds new field `automated_rsvp` and set to true when rsvp happens automatically when someone else drops out of workshop or admin moves up from waiting list
- display both `rsvp_time` and `reminded_at` on workshop admin page. This should enable us to figure out if there is an issue with the reminder emails, as at the moment they do seem to be sent out (checked for workshop https://codebar.io/workshops/1707 on our Sendgrid account and was able to confirm that reminders were going out)

### Updated view of member on admin/workshop page

rsvp_reminder_and_automated_rsvp

### Hover over top icon
rsvp_time

### Hover over bottom icon
reminded_at

## Hover over right hand side icon incidating admin or waitinglist automated addition to attendees
automated_rsvp

(data generated using Faker)
Resolves #1156

### DIFF
--- a/db/migrate/20200511151348_add_automated_rsvp_to_workshop_invitation.rb
+++ b/db/migrate/20200511151348_add_automated_rsvp_to_workshop_invitation.rb
@@ -1,0 +1,5 @@
+class AddAutomatedRsvpToWorkshopInvitation < ActiveRecord::Migration
+  def change
+    add_column :workshop_invitations, :automated_rsvp, :boolean
+  end
+end
--- a/app/controllers/admin/invitations_controller.rb
+++ b/app/controllers/admin/invitations_controller.rb
@@ -14,6 +14,7 @@ class Admin::InvitationsController < Admin::ApplicationController
         @invitation.update_attribute(:attending, true)
         @workshop.send_attending_email(@invitation) if @workshop.future?
 
+        @invitation.update(rsvp_time: Time.zone.now, automated_rsvp: true)
         message = "You have added #{@invitation.member.full_name} to the workshop as a #{@invitation.role}."
       else
         @invitation.update_attribute(:attending, false)
--- a/app/controllers/admin/invitations_controller.rb
+++ b/app/controllers/admin/invitations_controller.rb
@@ -11,10 +11,9 @@ class Admin::InvitationsController < Admin::ApplicationController
       attending = params[:attending]
 
       if attending.eql?('true')
-        @invitation.update_attribute(:attending, true)
+        @invitation.update(attending: true, rsvp_time: Time.zone.now, automated_rsvp: true)
         @workshop.send_attending_email(@invitation) if @workshop.future?
 
-        @invitation.update(rsvp_time: Time.zone.now, automated_rsvp: true)
         message = "You have added #{@invitation.member.full_name} to the workshop as a #{@invitation.role}."
       else
         @invitation.update_attribute(:attending, false)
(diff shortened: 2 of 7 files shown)


db/migrate/20200511151348_add_automated_rsvp_to_workshop_invitation.rb
```diff
@@ -0,0 +1,5 @@
+class AddAutomatedRsvpToWorkshopInvitation < ActiveRecord::Migration
+  def change
+    add_column :workshop_invitations, :automated_rsvp, :boolean
```
notapatch, line 3: ... several days later

Hi @despo

If there is any way it would help to have :automated_rsvp set to a boolean default at the database level? Currently, it is set to NULL/nil by the database by default and if the application code does not assign a boolean the nil will leak into the app. This is known as the three state boolean problem - booleans are only defined as true and false, the behaviour of a nil boolean .... well, is anyone's guess.

If you search the issues by whiny [something I wrote will pop to the top](https://github.com/codebar/planner/issues/1070#issuecomment-541415156) - I wrote about it in regard to `WorkshopInvitation[:attending]` but it applies to all the boolean data in the database including :automated_rsvp.

An example of the problems this leads to:
```
# inviation_concerns
scope :not_accepted, -> { where('attending is NULL or attending = false') }
                                               /\
                                               ||
   nuke it from orbit it's the only way to be sure 
```
[Link to code](https://github.com/codebar/planner/blob/master/app/models/concerns/invitation_concerns.rb#L15)

I believe the migration fix would look something like this (see references):
```
# a new migration
def up
    change_column_default :workshop_invitations, :automated_rsvp, false
end

def down
    change_column_default :workshop_invitations, :automated_rsvp, nil
end
```


### Reference
[Avoid the three state boolean problem](https://thoughtbot.com/blog/avoid-the-threestate-boolean-problem)
[Adding :default => true to boolean in existing Rails column](https://stackoverflow.com/questions/8627156/adding-default-true-to-boolean-in-existing-rails-column)
[Rails api - Change column default](https://api.rubyonrails.org/v5.1.7/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_default)

despo:
@notapatch yes good idea. I did consider it briefly but then forgot to amend the database script to default the value to false.

Issue #1259, opened 2020-05-02
Any event invitations should not be emailed to banned members


despo:
## Description of the issue 📄
Even though banned members are unable to RSVP to the monthlies, they still receive invitations. 

The process sending out monthlies invitations should be updated to not include banned members.

## Steps to fix 🛠
Have a look at `InvitationManager` as thsi is where all event invitations are trigerred from
`app/models/invitation_manager.rb`

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

notapatch:
Quick clarification @despo:  so emails generated by all of these methods?

```
  # app/models/invitation_manager.rb
  def send_event_emails(event, chapter)
  def send_monthly_attendance_reminder_emails(monthly)
  def send_course_emails(course)
```
Or just `send_monthly_attendance_reminder_emails` ?

---

If anyone reading this wants to have a go, please do... 

There seems to be a scope in member called `not_banned` that might be useful.

```
Class Member
  ...
  scope :not_banned, lambda {
                       joins('LEFT OUTER JOIN bans ON members.id = bans.member_id')
                         .where('bans.id is NULL or bans.expires_at < CURRENT_DATE')
                     }
 ...
```

despo:
By having a brief look I think only the monthlies emails would have to be updated as it seems like the `Member` [`in_group`](https://github.com/codebar/planner/blob/master/app/models/member.rb#L35) scope already takes `not_banned` into account so not all methods would have to be updated it seems.

```
  def chapter_students(chapter)
    Member.in_group(chapter.groups.students)
  end

  def chapter_coaches(chapter)
    Member.in_group(chapter.groups.coaches)
  end
end
```
https://github.com/codebar/planner/blob/master/app/models/invitation_manager.rb#L45

**However**, I am mostly interested in the tests being expanded to cover this scenario even for the methods that don't require any change, and this would include tests for methods included in https://github.com/codebar/planner/blob/master/app/models/concerns/workshop_invitation_manager_concerns.rb as well.

[...]

PR #1273, opened 2020-05-13
[WIP] Pr 1259 event invitation not emailed to banned

notapatch:
This is my work so far for issue [#1259](
https://github.com/codebar/planner/issues/1259) - I thought I'd keep you in the loop as it was high priority.

This work removes banned members at the scope level. Proving a negative is hard so for all cases I had a test that allowed a banned member and did another test with the minimum change to reject the banned member.

I was surprised that `in_group` filtered out the banned members rather than call it `allowed_in_group` or an `allowed` scope but I thought, well, I can't think of any reason to allow any banned member appearing in invitations so the above changes had the intention of making it impossible to get a banned member in an invitation scope.

- ea60cfa  - test to confirm `in_group` removes banned members
- 72763d3 - test to confirm `attending_meeting` removes banned members
- 48a8240 - invitable concern was missing any testing - this tested the current code with the shared_example
- b0924be - Changes invitable concern to remove banned members plus testing

These changes will remove all banned members from InvitationManager's lists (because they end up calling the above banned cleared scopes). The exception is `retrieve_and_notify_waitlisted(role:)` which I will look at if the work so far is OK.

When `retrieve_and_notify_waitlised` is complete then there can be no banned members getting an invitation. However, the issue expressly wanted tests for each of the methods in the WorshopInvitationManagerConcern and that would be the next step of the work.

### DIFF
--- a/app/presenters/workshop_presenter.rb
+++ b/app/presenters/workshop_presenter.rb
@@ -96,7 +96,7 @@ class WorkshopPresenter < EventPresenter
   private
 
   def students_checklist
-    model.attending_students.order('note asc').each_with_index.map do |a, pos|
+    model.attending_students.each_with_index.map do |a, pos|
       "#{member_info(a.member, pos)}\t\t\t#{note(a)}"
     end.join("\n\n")
   end
--- a/spec/models/event_spec.rb
+++ b/spec/models/event_spec.rb
@@ -2,6 +2,7 @@ require 'spec_helper'
 
 RSpec.describe Event, type: :model  do
   subject(:event) { Fabricate(:event) }
+  include_examples "Invitable", :invitation, :event
 
   it { should be_valid }
 
--- a/spec/models/meeting_spec.rb
+++ b/spec/models/meeting_spec.rb
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
 RSpec.describe Meeting, type: :model  do
+  include_examples "Invitable", :meeting_invitation, :meeting
+
   context 'validations' do
     subject { Meeting.new }
 
--- a/spec/support/shared_examples/behaves_like_invitable.rb
+++ b/spec/support/shared_examples/behaves_like_invitable.rb
@@ -1,0 +1,61 @@
+RSpec.shared_examples 'Invitable' do |invitation_type, invitable_type|
+  let(:invitable) { Fabricate(invitable_type) }
+
+  context '#attendances' do
+    it 'permits accepted' do
+      invitation = Fabricate(invitation_type,
+                             invitable_type => invitable,
+                             attending: true)
+
+      expect(invitable.reload.attendances).to include(invitation)
+    end
+
+    it 'rejects non-accepted' do
+      invitation = Fabricate(invitation_type,
+                             invitable_type => invitable,
+                             attending: false)
+
+      expect(invitable.reload.attendances).to_not include(invitation)
+    end
+  end
+
+  context '#attending_students' do
+    it 'accepts attending students' do
+      invitation_to_student = Fabricate(invitation_type,
+                                        role: 'Student',
+                                        invitable_type => invitable,
+                                        attending: true)
+
+      expect(invitable.reload.attending_students).to include(invitation_to_student)
+    end
+
+    it 'rejects non-attending students' do
+      invitation_to_student = Fabricate(invitation_type,
+                                        role: 'Student',
+                                        invitable_type => invitable,
+                                        attending: false)
+
+      expect(invitable.reload.attending_students).to_not include(invitation_to_student)
+    end
+  end
+
+  context '#attending_coaches' do
+    it 'accepts attending coaches' do
+      invitation_to_coach = Fabricate(invitation_type,
+                                      role: 'Coach',
+                                      invitable_type => invitable,
+                                      attending: true)
+
+      expect(invitable.reload.attending_coaches).to include(invitation_to_coach)
+    end
+
+    it 'rejects non-attending coaches' do
+      invitation_to_coach = Fabricate(invitation_type,
+                                      role: 'Coach',
+                                      invitable_type => invitable,
+                                      attending: false)
+
+      expect(invitable.reload.attending_coaches).to_not include(invitation_to_coach)
+    end
+  end
+end
--- a/spec/support/shared_examples/behaves_like_invitable.rb
+++ b/spec/support/shared_examples/behaves_like_invitable.rb
@@ -17,6 +17,15 @@ RSpec.shared_examples 'Invitable' do |invitation_type, invitable_type|
 
       expect(invitable.reload.attendances).to_not include(invitation)
     end
+
+    it 'rejects banned accepted' do
+      invitation = Fabricate(invitation_type,
+                             member: Fabricate(:banned_member),
+                             invitable_type => invitable,
+                             attending: true)
+
+      expect(invitable.reload.attendances).to_not include(invitation)
+    end
   end
 
   context '#attending_students' do
@@ -37,6 +46,17 @@ RSpec.shared_examples 'Invitable' do |invitation_type, invitable_type|
 
       expect(invitable.reload.attending_students).to_not include(invitation_to_student)
     end
+
+    it 'rejects banned attending students' do
+      invitation_to_banned_student = Fabricate(invitation_type,
+                                        member: Fabricate(:banned_member),
+                                        role: 'Student',
+                                        invitable_type => invitable,
+                                        attending: true)
+
+                                        # byebug
+      expect(invitable.reload.attending_students).to_not include(invitation_to_banned_student)
+    end
   end
 
   context '#attending_coaches' do
@@ -56,6 +76,16 @@ RSpec.shared_examples 'Invitable' do |invitation_type, invitable_type|
                                       attending: false)
 
       expect(invitable.reload.attending_coaches).to_not include(invitation_to_coach)
+    end
+
+    it 'rejects banned attending coaches' do
+      invitation_to_banned_student = Fabricate(invitation_type,
+                                        member: Fabricate(:banned_member),
+                                        role: 'Student',
+                                        invitable_type => invitable,
+                                        attending: true)
+
+      expect(invitable.reload.attending_students).to_not include(invitation_to_banned_student)
     end
   end
 end
--- a/app/models/concerns/invitable.rb
+++ b/app/models/concerns/invitable.rb
@@ -13,19 +13,11 @@ module Invitable
     end
 
     def attending_students
-      invitations.where(role: 'Student')
-                 .accepted
-                 .includes(:member)
-                 .joins(:member).merge(Member.not_banned)
-                 .order('members.name, members.surname')
+      attendances.where(role: 'Student').order('members.name, members.surname')
     end
 
     def attending_coaches
-      invitations.where(role: 'Coach')
-                 .accepted
-                 .includes(:member)
-                 .joins(:member).merge(Member.not_banned)
-                 .order('members.name, members.surname')
+      attendances.where(role: 'Coach').order('members.name, members.surname')
     end
   end
 end
--- a/app/models/concerns/invitable.rb
+++ b/app/models/concerns/invitable.rb
@@ -8,14 +8,24 @@ module Invitable
   module InstanceMethods
     def attendances
       invitations.accepted
+                 .includes(:member)
+                 .joins(:member).merge Member.not_banned
     end
 
     def attending_students
-      invitations.where(role: 'Student').accepted.includes(:member).order('members.name, members.surname')
+      invitations.where(role: 'Student')
+                 .accepted
+                 .includes(:member)
+                 .joins(:member).merge(Member.not_banned)
+                 .order('members.name, members.surname')
     end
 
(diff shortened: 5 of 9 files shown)


app/presenters/workshop_presenter.rb
```diff
@@ -96,7 +96,9 @@ def pairing_csv
   private
 
   def students_checklist
-    model.attending_students.order('note asc').each_with_index.map do |a, pos|
+    # TODO: review if it needs this removed scope or not
+    #       Not going to worry unless the rest of the code is OKd
```
despo, line 100: TODO can be removed as this does not require `not_banned` scope to be used. Ideally members should be removed from all RSVPed events as soon as they are banned, but even if this is missed, organisers would be able to see they are banned and they wouldnt be able to RSVP to any other future events so that is sufficient.

spec/models/event_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 RSpec.describe Event, type: :model  do
   subject(:event) { Fabricate(:event) }
+  include_examples "Invitable", :invitation, :event
```
despo, line 5: 👍

spec/models/meeting_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
 RSpec.describe Meeting, type: :model  do
+  include_examples "Invitable", :meeting_invitation, :meeting
+
```
despo, line 5: 👍

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
+# shared examples for Invitable
+# Example usage:
+#  include_examples "Invitable", :workship_invitation, :workshop
+#
```
despo, line 4: This looks great, but can we remove the comment please. I think documentation about shared examples is easily available and we don't need comments in the code.

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+RSpec.shared_examples 'Invitable' do |invitation_type, invitable_type|
+  let(:invitable) { Fabricate(invitable_type) }
+
+  context '#attendances' do
+    it 'permits accepted' do
+      invitation = Fabricate(invitation_type,
+                             invitable_type => invitable,
```
despo, line 11: can this be updated to `invitable_type: invitable` please

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+                             attending: true)
+
+      expect(invitable.reload.attendances).to include(invitation)
+    end
+
+    it 'rejects un-accepted' do
+      invitation = Fabricate(invitation_type,
+                             invitable_type => invitable,
```
despo, line 19: can this be updated to `invitable_type: invitable` please

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+  context '#attendances' do
+    it 'permits accepted' do
+      invitation = Fabricate(invitation_type,
+                             invitable_type => invitable,
+                             attending: true)
+
+      expect(invitable.reload.attendances).to include(invitation)
+    end
+
+    it 'rejects un-accepted' do
```
despo, line 17: Can we change this to read non accepted. I find un-accepted a bit confusing.

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+                             invitable_type => invitable,
+                             attending: false)
+
+      expect(invitable.reload.attendances).to_not include(invitation)
+    end
+
+    it 'rejects banned accepted' do
+      invitation = Fabricate(invitation_type,
+                             member: Fabricate(:banned_member),
+                             invitable_type => invitable,
```
despo, line 28: same as above

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+                                        invitable_type => invitable,
+                                        attending: true)
+
+      expect(invitable.reload.attending_students).to include(invitation_to_student)
+    end
+
+    it 'rejects un-attending students' do
+      invitation_to_student = Fabricate(invitation_type,
+                                        role: 'Student',
+                                        invitable_type => invitable,
```
despo, line 48: same as above

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+
+      expect(invitable.reload.attendances).to_not include(invitation)
+    end
+  end
+
+  context '#attending_students' do
+    it 'accepts attending students' do
+      invitation_to_student = Fabricate(invitation_type,
+                                        role: 'Student',
+                                        invitable_type => invitable,
```
despo, line 39: same as above

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+                                        attending: false)
+
+      expect(invitable.reload.attending_students).to_not include(invitation_to_student)
+    end
+
+    it 'rejects banned attending students' do
+      invitation_to_banned_student = Fabricate(invitation_type,
+                                        member: Fabricate(:banned_member),
+                                        role: 'Student',
+                                        invitable_type => invitable,
```
despo, line 58: same as above

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+                                        # byebug
+      expect(invitable.reload.attending_students).to_not include(invitation_to_banned_student)
+    end
+  end
+
+  context '#attending_coaches' do
+    it 'accepts attending coaches' do
+      invitation_to_coach = Fabricate(invitation_type,
+                                      role: 'Coach',
+                                      invitable_type => invitable,
```
despo, line 70: same as above

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+                                      attending: false)
+
+      expect(invitable.reload.attending_coaches).to_not include(invitation_to_coach)
+    end
+
+    it 'rejects banned attending coaches' do
+      invitation_to_banned_student = Fabricate(invitation_type,
+                                        member: Fabricate(:banned_member),
+                                        role: 'Student',
+                                        invitable_type => invitable,
```
despo, line 89: same as above

spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -0,0 +1,95 @@
...
+                                      invitable_type => invitable,
+                                      attending: true)
+
+      expect(invitable.reload.attending_coaches).to include(invitation_to_coach)
+    end
+
+    it 'rejects un-attending coaches' do
+      invitation_to_coach = Fabricate(invitation_type,
+                                      role: 'Coach',
+                                      invitable_type => invitable,
```
despo, line 79: same as above

app/presenters/workshop_presenter.rb
notapatch: `attending_students` along with all the other Invitable methods will remove banned Members. 

The TODO was to remind me that `.order('note asc')` was breaking the code. I don't actually know how it works now because there is already an ordering on attending_students [`.order('members.name, members.surname')`]  so a second in the chain order is ignored anyway. 

```
# current situation before I did anything was in effect:
.order('members.name, members.surname').order('note asc')
```


I will remove `order('note asc') `along with the TODO.

spec/support/shared_examples/behaves_like_invitable.rb
notapatch: done!

app/presenters/workshop_presenter.rb
despo: I saw the PR on Slack and completely missed that it was WIP when I added my comments last night.

If anything, I think it would be more beneficial to order the attendees by RSVP time so that it gives a better idea to organisers about when someone RSVPed without having to necessarily check all logged RSVP times.

spec/support/shared_examples/behaves_like_invitable.rb
notapatch: I did initially try JSON style/Ruby 1.9 hash style and I got the following error.

```ruby
  Failure/Error:
       invitation = Fabricate(invitation_type,
                              member: Fabricate(:banned_member),
                              invitable_type: invitable,
                              attending: true)
     
     ActiveRecord::UnknownAttributeError:
       unknown attribute 'invitable_type' for WorkshopInvitation.
     Shared Example Group: "Invitable" called from ./spec/models/workshop_spec.rb:5
     # ./spec/support/shared_examples/behaves_like_invitable.rb:22:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:80:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method `invitable_type=' for #<WorkshopInvitation:0x00007fa175f84eb8>
     #   ./spec/support/shared_examples/behaves_like_invitable.rb:22:in `block (3 levels) in <top (required)>'
```

I've done some reading around to try and understand this as best as I can (see references below). I haven't found anything authoritative [the following quote was the most interesting](https://stackoverflow.com/questions/8675206/is-there-any-difference-between-the-key-value-and-key-value-hash-no).

> The JavaScript style (key: value) is only useful if all of your Hash keys are "simple" symbols (more or less something that matches /\A[a-z_]\w*\z/i, AFAIK the parser uses its label pattern for these keys).

So, with the JSON style/Ruby 1.9 hash style is good for literal symbols - the colon at the end means it can convert it into symbol and use it as a hash key. In this case invitable_type is **not** a literal symbol it's a variable that holds a symbol, so I believe, it can't be converted into a symbol by the JSON style hash key implicitly - it thinks it a method and gives the above error.

There _may_ be ways around this and I've tried weird conversions that would look way worse the than hash rocket without success. 

Anyway, this is what I've tried. Do you know of a solution or would it be acceptable to leave it as a hash rocket?

### References
[Ruby docs Hash](https://ruby-doc.org/core-2.7.0/Hash.html)
[How to access a symbol hash key using a variable in Ruby](https://stackoverflow.com/questions/25537934/how-to-access-a-symbol-hash-key-using-a-variable-in-ruby)
[Is there any difference between the `:key => “value”` and `key: “value”` hash notations?](https://stackoverflow.com/questions/8675206/is-there-any-difference-between-the-key-value-and-key-value-hash-no)

app/presenters/workshop_presenter.rb
notapatch: It sounds like slack gives you the PR name and not the Github title - not sure what to do about that. 

It's definitely fixable. The ordering could be removed from the invitable#attending_students and invitable#attending_coaches and the order explicitly applied by having a `by_name` `by_rsvp_time` It looks like it would affect about 10 places. 🤷‍♂️

spec/support/shared_examples/behaves_like_invitable.rb
notapatch: @despo I pressed the wrong button. I pressed resolved when I meant to press comment. (Github messaging isn't my strong point as I'm sure matyikriszta would attest to).

despo: I don't see it as resolved :)

However I'm happy with this explanation and will set all `key: value` comments I made to resolved.

app/presenters/workshop_presenter.rb
despo: @notapatch ah ok. Sounds like something for a new PR then so let's ignore my comment about changing the ordering here and I'll open a new issue.

despo: https://github.com/codebar/planner/issues/1275

spec/support/shared_examples/behaves_like_invitable.rb
notapatch: done! changed all un- to non-  I think the test pairings looked better 👍

despo:
@notapatch 💯

notapatch:
Haha, there's a couple more changes I was going to do. Shall I open another PR or leave it here?

The situation is that Invitable will filter out banned members - so that should mean InvitationManager will have removed banned members with the exception of `retrieve_and_notify_waitlisted`.

If I had continued the TODO was:
1. Move `attendee?` from Workshop into Invitable as it depended on Invitable methods
   -  I thought it would help the focus of Workshop.

2. Look at Waitlist and make sure no banned members could get into it.

Not pushing - as long as you know Waitlist hasn't been secured. I'm happy.

despo:
@notapatch 🤦  you are welcome to do 1 on a new PR. As for 2, I wouldn't. Banned members should not be able to perform any actions on invitations unless they have RSVPed and I think a better way to tackle this would be to update `Admin/BanController` to instead remove a banned member from any upcoming events by checking if they have any upcoming attending invitations (for all types of events).

PR #1280, opened 2020-05-17
(Fix) invitation_management members not_accepted n+1

notapatch:


app/views/admin/workshops/_invitation_management.html.haml
```diff
@@ -15,7 +15,9 @@
       .row
         = simple_form_for :workshop, url: admin_workshop_invitations_path(@workshop, attending: true), remote: true, method: :put do |f|
           .medium-6.columns
-            = f.select :invitations, @workshop.invitations.not_accepted.all.map { |u| [ "#{u.member.full_name} (#{u.role})", u.token] }, { include_blank: true }, { class: 'chosen-select', required: true, data: { placeholder: 'Select a member to RSVP' } }
+            = f.select :invitations,
+                       @workshop.invitations.includes(:member).not_accepted.all.map { |u| [ "#{u.member.full_name} (#{u.role})", u.token] },
+                       { include_blank: true }, { class: 'chosen-select', required: true, data: { placeholder: 'Select a member to RSVP' } }
```
despo, line 20: Thank you. Is it possible to also move the 'Select a member to RSVP` to the locale file please

notapatch: I did it for all of the select_a_member to RSVP. I was thinking that localization might be better taking the "invitation" namespace higher. So instead of:
`messages.invitations.select_a_member_to_rsvp`
Have
`invitations.select_a_member_to_rsvp`

_if_ you agree with that it could be given as a beginner issue. (I can always do it later if there's no takers)

notapatch:
Ran for 2 min 54 sec 👍

app/views/admin/workshops/_invitation_management.html.haml
despo: Yes that is better. I prefer using the `message` namespace specifically for notice or warning messages as well as it's clearer what the message is intended for, just completely missed it when reviewing.

Our localisation is always WIP so improving it and using shorthands would be handy. We do use them a little bit (eg some forms mostly) but it is hard to enforce in PRs and it's not as if we get many anyway. Happy for an issue to be created about it though.

PR #1281, opened 2020-05-17
travis cache bundler

notapatch:
Half of the build time is installing gems and this tells Travis to cache them.
What you should see is that the first build takes around 5min+ as normal but after that it takes approx 2.50. 

We can also save 10s if we use Ruby 2.4.5 (which is installed on Travis by default).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 dist: xenial
 language: ruby
+cache: bundler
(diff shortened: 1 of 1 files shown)


despo:
@notapatch thanks. As we are using ruby 2.4.2 in our production environment we should stick with that when running tests as well.

PR #1283, opened 2020-05-17
(fix) Upgrade rails to resolve vulnerability

despo:
[CVE-2020-8163] Potential remote code execution of user-provided local names in Rails < 5.0.1


despo:
Taken care of by dependabot in #1287

Issue #1151, opened 2019-11-03
Timestamps deprecation warnings


jhsu802701:
Running the database migrations leads to multiple instances of the following deprecation warning message:
```
WARN -- : DEPRECATION WARNING: `#timestamps` was called without specifying an option for `null`. In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to prevent the behavior of your existing migrations from changing.
```

notapatch:
Good spot 👍 

What we have:

```
class CreateInvitations < ActiveRecord::Migration
  def change
    create_table :invitations do |t|
      t.references :sessions, index: true
      ...
      t.timestamps
    end
  end
end
```

Rails 5 is changing the default[1]: 
> Changed the default null value for timestamps to false. (commit)

We should want to stay with the default. I don't think we are doing anything clever with the timestamps? In that case what we want is:

```
class CreateInvitations < ActiveRecord::Migration
  def change
    create_table :invitations do |t|
      t.references :sessions, index: true
      ...
      t.timestamps null: false        <= this gives no nulls in column
    end
  end
end
```

[The way to do it is to use `change_column_null`](https://api.rubyonrails.org/v6.0.0/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null) 

The main worry is that the table could have a null timestamp (always possible). In that case you can use a fourth parameter which sets the default. For our example the migration should look something like this (untested!):


```ruby
def change
  change_column_null :invitations, :timestamps, false, Time.now
end

```

Where **false means no nul**l.

The verification would be:

```sql
planner=# \d invitations
                                          Table "public.invitations"
     Column     |            Type             | Collation | Nullable |       Default
----------------+-----------------------------+-----------+-------------------------
 id             | integer                     |           | not null | next...
 created_at     | timestamp without time zone |           | not null |  <=== this
 updated_at     | timestamp without time zone |           | not null |   <=== this

```

Clearly, we can knock out these migrations in a few minutes - the main issue is time updating production. 


### References
[1] [Rails 5.0 Release notes - active-record-notable changes](https://guides.rubyonrails.org/5_0_release_notes.html#active-record-notable-changes)
[How to change a nullable column to not nullable in a Rails migration?](https://stackoverflow.com/questions/5966840/how-to-change-a-nullable-column-to-not-nullable-in-a-rails-migration)

PR #1285, opened 2020-05-18
Rails 5 upgrade prep

despo:
Upgrading to Rails 5 is a massive task, so I thought to approach it by first resolving non breaking issues that we can integrate in our codebase now. This brings down the failures from **227** to **~123** when running rspec against **rails v5.2.4.2** 

1. Update references to images
**Why**: Rails 5.2 changed the `image_tag` behaviour which now only accepts a url. Up to Rails 5.1 is was possible to pass the uploader object and rails could guess.
2. Wrap SQL in Arel
**Why**: To remove future depreciation warning due to non-attribute arguments not allowed in Rails 6. Known-safe values can be passed by wrapping them in Arel.sql().
3. Ensure belongs_to association is specified before utilising it in :through association as not doing so causes a failure

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -14,7 +14,7 @@ class DashboardController < ApplicationController
       hash
     end
 
-    @testimonials = Testimonial.order('RANDOM() ').limit(5).includes(:member)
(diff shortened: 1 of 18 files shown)


notapatch:
LGTM 👍 

There's this [deprecation issue on timestamps](https://github.com/codebar/planner/issues/1151) which is a Rails 5 blocker.

PR #1294, opened 2020-05-18
Remove gem rspec-its

notapatch:
["its" left the core of rspec a while a go.](https://gist.github.com/myronmarston/4503509) I think I possibly even removed some of them but the gem had been left in being loaded up each time the tests are run for no good reason.


despo:
@notapatch looks good. Happy to merge if you can rebase.

PR #1300, opened 2020-05-18
Rename shared examples to begin with stem of behaves_like_

notapatch:
- want to have consistency for naming of shared examples.
 - the other option is to remove the behaves_link_ stem

----
A commit that's been sitting around. Not looking at validations till tomorrow.


notapatch:
@despo rebased and built!

PR #1301, opened 2020-05-19
Workshop validates ends_at

notapatch:
- Chapter required (instead of chapter_id) because timezone method
   requires chapter.time_zone

---
There _might_ be issues with VirtualWorkshops and date_and_time as the :if, callbacks, super, and concerns make this non-trivial. It's a basic pass through.

### DIFF
--- a/spec/models/workshop_spec.rb
+++ b/spec/models/workshop_spec.rb
@@ -7,7 +7,7 @@ RSpec.describe Workshop, type: :model do
   context 'validates' do
     it { is_expected.to validate_presence_of(:chapter_id) }
 
-    context "date_and_time" do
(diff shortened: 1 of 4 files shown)


spec/models/workshop_spec.rb
```diff
@@ -16,13 +16,29 @@
       end
 
       it 'validate if chapter_id present' do
-        workshop.chapter_id = 1
+        workshop.chapter = Fabricate(:chapter)
```
despo, line 19: I believe the reason we use the chapter id is to simulate conrtoller behavior which sets the chapter_id on the workshop model before ti attempts to save.

notapatch:
@despo rebased and rebuit

Issue #1237, opened 2020-03-25
[Feature] Utilise workshop#rsvp_closes_at


despo:
## Description of the issue 📄

The workshop model has an `rsvp_closes_at` field that is not currently in use as it cannot be set when the workshop is created or edited.

- [ ] Enable organiser/admin to set `rsvp_closes_at` via admin workshop new and edit pages
- [ ] Ensure that workshop#rsvp_available? check is utilised 
  - [ ] in invitation page
  - [ ] in invitation controller e.g. https://github.com/codebar/planner/blob/867d3e9117863ba6a7c298cf6085652c0157d5ff/app/controllers/concerns/invitation_controller_concerns.rb#L41
  - in waiting list controller 

The list above may not be entirely accurate so please add relevant checks and tests where needed.

### Existing methods using `rsvp_closes_at`

https://github.com/codebar/planner/blob/867d3e9117863ba6a7c298cf6085652c0157d5ff/app/models/workshop.rb#L61

https://github.com/codebar/planner/blob/867d3e9117863ba6a7c298cf6085652c0157d5ff/spec/models/workshop_spec.rb#L56

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

PR #1302, opened 2020-05-19
Refactor workshop fabricator

notapatch:


spec/fabricators/workshop_fabricator.rb
```diff
@@ -1,86 +1,51 @@
...
+  ends_at Time.zone.now + 2.days + 2.hours
   chapter
+  virtual true
+  student_spaces 10
+  coach_spaces 10
+  slack_channel 'a-channel'
+  slack_channel_link 'https://codebar.slack.link'
   after_build do |workshop|
-    Fabricate(:workshop_sponsor, workshop: workshop, sponsor: Fabricate(:sponsor, number_of_coaches: 0), host: true)
+    Fabricate(:workshop_sponsor, workshop: workshop, sponsor: Fabricate(:sponsor), host: false)
```
despo, line 49: Since this is being refactored, as a virtual workshop does not have to have any sponsors, can we competely remove the after_build callback so that by default tests use the more common virtual workshop scenario of not having any sposors.

notapatch: I introduced a virtual_workshop_sponsored for a couple of tests that needed it.
On the tests that needed change I had the choice of running all the tests as virtual_workshop_sponsored or creating an extra context around the tests that needed the sponsor and making only them virtual_workshop_sponsored. I choose the former to avoid additional nesting which can get confusing for readers of the code - however this is an opinion and I'm happy to do either.

notapatch:
@despo 

One question which will help my domain knowledge. 

### Workshop attributes
 - #Invitable - I assume means people can be invited onto the course?
 - #rsvp_opens_at/rsvp_closes_at - I'm not sure  - something like the date range where people can ask to attend the course? So when rsvp_closes_at is nil it is still open to join?

So:
For the following Fabricators invitable has been set to false. I can understand not having false when rsvp opens in the future. But, if I understand the domain model, I am guessing that :`workshop_auto_rsvp_in_past` should, in most cases, have `invitable true`?

```ruby
Fabricator(:workshop_auto_rsvp_in_past, from: :workshop) do
  rsvp_opens_at Time.zone.now - 1.day
  invitable false  # <============= true maybe??
end

Fabricator(:workshop_auto_rsvp_in_future, from: :workshop) do
  rsvp_opens_at Time.zone.now + 1.day
  invitable false
end
```

Changing to true does not break the build.

despo:
@notapatch 
- `invitable` is an event (or specifically a workshop) that is available for members to RSVP and be invited to
- If we don't want an event to be `invitable` immediately, we can specify the`rsvp_opens_at` and  that takes over as the opening time, and `invitable` is ignored.
- `rsvp_closes_at` is not actually utilised at the moment, but it's meant to be the time that invitations can be `rejected` until (it's meant to prevent last minute cancellations as it does not give us sufficient time to allocate people from the waiting list etc this is also indicated in our emails) and it should take over the time that members can `reject` their invitations up to (eg [instead of 3.5hours before the workshop](https://github.com/codebar/planner/blob/master/app/controllers/concerns/invitation_controller_concerns.rb#L41)). (related issue https://github.com/codebar/planner/issues/1237)
- After looking at the workshop model, [`invitable` is ignored when `rsvp_opens_at` is set](https://github.com/codebar/planner/blob/master/app/models/workshop.rb#L71.) so we can just remove it from the fabricator completely. Some of the views also seem to be wrongly referring to `invitable` instead of `invitable_yet?` in error, I think making the switch was missed when `invitable_yet?` was introduced to the codebase ([eg here](https://github.com/codebar/planner/blob/9e2403a6aa3252f4236a56f485b5c4163efb8e91/app/views/admin/workshops/show.html.haml#L9), [here](https://github.com/codebar/planner/blob/867d3e9117863ba6a7c298cf6085652c0157d5ff/app/models/invitation_manager.rb#L5), [and here](https://github.com/codebar/planner/blob/867d3e9117863ba6a7c298cf6085652c0157d5ff/app/views/workshops/_actions.html.haml#L16))

notapatch:
Thanks very much. That does makes sense. I'll update the fabricators and `invitable` => `invitable_yet?`- so you'll soon know if I understood. I was going to look at `invitable_yet?` anyway because I thought it made the `_actions` partials read a bit clearer.

despo:
@notapatch that sounds great, thank you!

PR #1311, opened 2020-05-20
Changes around testing Ban

notapatch:
Using Ban and wanted to understand it better so I've added tests.

I made 4 changes to the actual code

| Comment  | Sha |  Note |
| ------------- | ------------- | ------------ |
| 1. Added ActiveSupport TimeHelpers  | 84f0dd3  | it's a way of testing time without 3rd party libraries. |
| 2.  Ban#expired_at required  | 0334f05  |[bans/new - expires_at is required](https://github.com/codebar/planner/blob/master/app/views/admin/bans/new.html.haml#L15) |
|  3. Custom validation refactored to use Ruby safe navigation  |  0ff139a   |  -  |
|  4. Prefer Time.zone.now over Date.current |  46b23fc   |  Not sure if it's important for Time.zone aware code on banning?  |

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -24,6 +24,7 @@ RSpec.configure do |config|
   config.include ApplicationHelper
   config.include LoginHelpers
   config.include JobrefHelpers
+  config.include ActiveSupport::Testing::TimeHelpers
   config.use_transactional_fixtures = false
   config.infer_base_class_for_anonymous_controllers = false
   config.order = 'random'
--- a/app/models/ban.rb
+++ b/app/models/ban.rb
@@ -6,7 +6,7 @@ class Ban < ActiveRecord::Base
 
   validate :valid_expiry_date?
 
-  scope :active, -> { where('expires_at > ?', Date.current) }
+  scope :active, -> { where('expires_at > ?', Time.zone.now) }
   scope :permanent, -> { where(permanent: true) }
 
   def active?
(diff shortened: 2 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -24,6 +24,7 @@
   config.include ApplicationHelper
   config.include LoginHelpers
   config.include JobrefHelpers
+  config.include ActiveSupport::Testing::TimeHelpers
```
despo, line 27: happy to add this, however just wondering, what value does it add to this PR or the existing tests?

notapatch: Allows you to use [TimeHelpers](https://api.rubyonrails.org/v5.2.4.2/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel) which are useful when testing TimeZone aware code. [I used `travel` in the code here - I like the block form which only changes the time within the block](https://github.com/codebar/planner/pull/1311/commits/46b23fc2be3f7b67fcca60d19dc35eab43da5ae7#diff-0852c341dc1026ff5fd36822b4b82433R33)

[My idea would be to remove any reference to TimeCop](https://andycroll.com/ruby/replace-timecop-with-rails-time-helpers-in-rspec/)

despo: Nice, I didn't notice the reference to `travel`. Removing TimeCop if there is an ActiveSupport alternative sounds great. I couldn't even remember that we were using it :)

PR #1312, opened 2020-05-20
Replace Timecop with ActiveSupport::Testing::TimeHelpers

notapatch:
Controller tests will need to be looked at during upgrade but for an example ...


notapatch:
rebased!

PR #1315, opened 2020-05-21
Remove ApplicationController#upcoming_workshops  

notapatch:
- available to all views but not used
 - similar but not the same:
   - @upcoming_workshops used in
       - chapter_controller and show view
       - dashboard_controller and show view
   - ChapterPresenter an upcoming_workshops

---
Also it's private method - but I don't remember if Ruby enforces that 🤔

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -159,12 +159,6 @@ class ApplicationController < ActionController::Base
 
   helper_method :jobs_pending_approval, :chapters
 
-  def upcoming_workshops
-    Workshop.upcoming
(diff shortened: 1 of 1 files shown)


despo:
@notapatch  Good one! thank you. It used to be called from [`layouts/_aside`](https://github.com/codebar/planner/blame/25d623df750ac01854c192482a300f9e1ff44c2b/app/views/layouts/_aside.html.haml) but that was quite a while ago 😆

notapatch:
Great! `chapters` just above it is also worth reviewing.

PR #1321, opened 2020-05-21
Refactor sponsor validations

notapatch:
### (refactor) Sponsor validations using shoulda  a4180ec 

 - added validation for enum level
 - kind_of? => is_a? - no change in the code as they are
   synonymous. Change because Rubocop has annointed is_a?
 - remove respond_to because it's an incomplete way of finding
   the attributes of a class - they get out of date and it's
   just easier to use the schema.
 - prefer the more expressive and compact shoulda matchers over
   the hand written validations where possible
 - remove rspec-collection_matchers that are no longer part of
   rspec core

--- 

### (feature) Sponsor validate presence of level 4bbf8d3 

I've added a separate more speculative commit 4bbf8d3 as I wasn't sure what the effect of making the level required. If you press create by default you get standard level. I imagine there aren't many that don't have a level? I assume, if you tried to edit one without a level it would then ask you for a level.

---

#### Unknown
Address is required in the model but the system doesn't care if it's blank. Not sure if that's important or not.

### DIFF
--- a/app/models/sponsor.rb
+++ b/app/models/sponsor.rb
@@ -17,7 +17,7 @@ class Sponsor < ActiveRecord::Base
   has_many :contacts, through: :member_contacts, class_name: 'Member', foreign_key: 'member_id'
 
   validates :level, inclusion: { in: Sponsor.levels.keys }
-  validates :name, :address, :avatar, :website, :seats, presence: true
+  validates :name, :address, :avatar, :website, :seats, :level, presence: true
   validate :website_is_url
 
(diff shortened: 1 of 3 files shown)


despo:
@notapatch all sponsors should have a level. I don't know if old sponsors were added a level after it was put in place, but it's easy enough to set one if their details are being manually updated anyway.

About the address, I think we would like to have it for all sponsors but would just leave it as is for now as it's something we are currently discussing.

Issue #206, opened 2015-02-12
Coach feedback


despo:
Currently all attending students receive an email the day after a workshop asking them for feedback.

The student feedback is anonymous, however I'm not sure if this should apply to the coaches. Maybe optionally anonymous.

-----
Refactor and utilise existing feedback model to achieve this.

notapatch:
A rake task runs everyday which checks for workshops that were completed the day before. The rake task puts the `Workshop` and `Member` information into a FeebackRequest object and calls `create`.

```ruby
# lib/tasks/feedback.rake - with logging removed
task request: :environment do
  workshops = Workshop.completed_since_yesterday
  workshops.each do |workshop|
    WorkshopInvitation.accepted.where(workshop: workshop, role: 'Student').map do |invitation|
      FeedbackRequest.create(member: invitation.member, workshop: workshop, submited: false)
    end
  end
end
```

`FeedbackRequest` adds a unique token to the the object which it saves along with the Workshop and Member information to the database. After the save an email, using `FeedbackRequestMailer.request_feeback`, is sent out which asks the Member to give feedback.

```ruby
# app/models/feedback_request.rb
class FeedbackRequest < ActiveRecord::Base
  ...
  validates :token, uniqueness: { case_sensitive: false }
   ...
  before_validation :set_token
  after_create :email

  private

  def set_token
    self.token = loop do
      random_token = SecureRandom.urlsafe_base64(nil, false)
      break random_token unless self.class.where(token: random_token).exists?
    end
  end

  def email
    FeedbackRequestMailer.request_feedback(workshop, member, self).deliver_now
  end
end
```

The request_feedback email is a standard email with the exception of a link feedback controller using the token created in the FeedbackRequest.
```haml
# app/views/feedback_request_mailer/request_feedback.html.haml
.content
%table{ bgcolor: "#FFFFFF" }
  %tr
    %td
      %p
        =link_to "Submit feedback", full_url_for(feedback_url(@feedback_request.token)), class: 'btn'
```

feedback_url is routed to `FeedbackController#show` and uses the `FeedbackRequest` token to link the request to the Feedback. The show has a feedback form on it and behaves more like a `new` action than a show. 

```ruby
class FeedbackController < ApplicationController
  def show
    feedback_request = FeedbackRequest.find_by(token: params[:id], submited: false)

    if feedback_request.nil?
      flash[:notice] = I18n.t('messages.feedback_not_found')
      return redirect_to root_path
    end

[...]

PR #1324, opened 2020-05-21
(refactor) Feedback model validations

notapatch:
- prefer the more expressive and compact shoulda matchers
   over the hand written validations where possible
 - removes rspec-collection_matchers that have been removed
   from rspec core
 - remove respond_to because it's an incomplete way of finding
   the attributes of a class - they get out of date and it's
   just easier to use the schema.
---
My stack was blown by method name with coach in twice. I've suggested a different name for the method.

### DIFF
--- a/spec/models/feedback_spec.rb
+++ b/spec/models/feedback_spec.rb
@@ -1,103 +1,39 @@
 require 'spec_helper'
 
-RSpec.describe Feedback, type: :model  do
+RSpec.describe Feedback, type: :model do
   subject(:feedback) { Fabricate.build(:feedback) }
-
-  let(:valid_feedback_token) { 'valid_feedback_token' }
-  let(:invalid_feedback_token) { 'invalid_feedback_token' }
-  let(:chapter) { Fabriate(:chapter) }
-
-  it { should respond_to(:request) }
-  it { should respond_to(:rating) }
-  it { should respond_to(:suggestions) }
-  it { should respond_to(:coach) }
-  it { should respond_to(:tutorial) }
 
   context 'validations' do
     context '#rating' do
-      it 'should not be blank' do
-        feedback = Fabricate.build(:feedback, rating: nil)
-
-        expect(feedback).to_not be_valid
-        expect(feedback).to have(1).error_on(:rating)
-      end
-
-      it 'should accept numbers from 1 to 5' do
-        (1..5).each do |rating|
-          feedback = Fabricate.build(:feedback, rating: rating)
-
-          expect(feedback).to be_valid
-        end
-      end
-
-      it 'should be invalid with number higher than 5' do
-        feedback = Fabricate.build(:feedback, rating: 6)
-
-        expect(feedback).to_not be_valid
-        expect(feedback).to have(1).error_on(:rating)
-      end
-
-      it 'should be invalid with number lower than 1' do
-        feedback = Fabricate.build(:feedback, rating: 0)
-
-        expect(feedback).to_not be_valid
-        expect(feedback).to have(1).error_on(:rating)
-      end
-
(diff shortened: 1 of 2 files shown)


spec/models/feedback_spec.rb
```diff
@@ -3,75 +3,27 @@
...
-      end
-
-      it "should display correct error message when member with 'student' role given" do
+      it 'rejects none-coach members' do
         feedback = Fabricate.build(:feedback, coach: Fabricate(:student))
 
         feedback.valid?
 
-        expect(feedback.errors.messages[:coach]).to include("Coach member doesn't have 'coach' role.")
+        expect(feedback.errors[:coach]).to include("Coach member doesn't have 'coach' role.")
```
despo, line 26: This is wrong. A legacy from how things used to work. 
Anyo member can be a coach at a workshop and they do not be subscribed to any groups.

It's not related to your PR though so I can either merge or be happy to accept a change removing this validation entirely.

spec/models/feedback_spec.rb
```diff
@@ -3,75 +3,27 @@
...
 
-      it "is invalid with member with 'student' role" do
-        feedback = Fabricate.build(:feedback, coach: Fabricate(:student))
-
-        expect(feedback).to_not be_valid
-        expect(feedback).to have(1).error_on(:coach)
-      end
-
-      it "should display correct error message when member with 'student' role given" do
+      it 'rejects none-coach members' do
```
despo, line 21: small typo, non-coach

spec/models/feedback_spec.rb
notapatch: I removed the validation completely....

spec/models/feedback_spec.rb
notapatch: section is gone

notapatch:
I stuffed the whole file through rubocop which picked up a couple of formatting errors.

despo:
@notapatch thank you 💯

spec/models/feedback_spec.rb
```diff
@@ -1,103 +1,37 @@
...
+    it { is_expected.to validate_presence_of(:tutorial) }
   end
 
   context '#submit_feedback' do
     let(:feedback_request) { Fabricate(:feedback_request) }
 
-    let (:params) do
-      { rating: 4, coach: Fabricate(:coach), tutorial: Fabricate(:tutorial) }
-    end
+    let(:params) { rating: 4, coach: Fabricate(:coach), tutorial: Fabricate(:tutorial) }
```
despo, line 18: @notapatch I think it needs to contain a nested hash
`let(:params) {  { rating: 4, coach: Fabricate(:coach), tutorial: Fabricate(:tutorial) } }`

notapatch: That did indeed confuse me - I would say that nearly all of the submit_feedback code did leave me scratching my head. I put the Rubocop change in and backed away from the code slowly. 💣

despo: I think it's barely been touched since it was put in place as it just works. It could potentially do with a refactoring but then again there are so many aspects of the app that could be improved! 

Maybe we'll get to it when we put coach feeedback in place as it can probably be reused for that purpose.

notapatch: I would add it as a separate issue. Refactoring validations is relaxation between issues and I've been all too successful at not working at the cancelling issue. In that regards, I'm looking at refactoring workshop/_actions so that I can put cancelling in _without_ making the partial any more complex - I feel there's a slimmer, friendlier partial waiting to come out - it could also be delusion.

despo: @notapatch https://github.com/codebar/planner/issues/206

PR #1326, opened 2020-05-22
Remove redundant self

notapatch:
- prefer removing redundant self
 - auto-corrected
   rubocop -a --only=Style/RedundantSelf

---

If any of these Rubocops aren't liked I can update the rubocop.yml appropriately.

### DIFF
--- a/app/models/chapter.rb
+++ b/app/models/chapter.rb
@@ -50,6 +50,6 @@ class Chapter < ActiveRecord::Base
   end
 
   def set_slug
-    self.slug ||= self.name.parameterize
+    self.slug ||= name.parameterize
   end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -38,6 +38,6 @@ class Course < ActiveRecord::Base
   private
 
   def set_slug
-    self.slug = title.parameterize if self.slug.nil?
+    self.slug = title.parameterize if slug.nil?
   end
 end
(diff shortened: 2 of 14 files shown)


notapatch:
@despo if at all possible I would keep Rubocop commits as fixing the one class of problem.

```
# 1 issue to fix
Avoid using rescue in its modifier form. New

    WorkshopSponsor.hosts.for_workshop(id).first.sponsor rescue nil

Severity: Minor
Found in app/models/workshop.rb by rubocop
```

despo:
@notapatch can we have a separate commit for the other fix then? 
I know it's a bit frustrating when new issues come up just because you've modified other bits of code on the line but even though in an ideal world we would tackle each issue separately, the reality is that we are more likely to attempt to resolve them as part of other work.

notapatch:
Needed to make a change to `Workshop#host ` but I didn't understand the code (never used rescue in a one-liner). The code relied on WorkshopSponsor so I tested that before finally looking at Workshop#host.

I didn't use the Rubocop 'fix' because I wanted to avoid using exceptions in the normal flow of the code, if possible, and safe navigation offered that solution.

despo:
@notapatch looks good. The code just mean that t if an exception was thrown at any point in that call, so for example if there are no sponsors set as hosts on the workshop at all and thus `[].first` responds with `nil` when `[].first.sponsor` fails and throws an exception it catches it and returns `nil` instead.

It was also put in place long before the safe navigation operator was introduced (or maybe before it started being common in Ruby and reccommended for use).

PR #1329, opened 2020-05-22
Remove coffee script

koddsson:
As far as I can tell there was only one file written in coffeescript so I converted it to javascript and removed a whole dependency we were keeping around just for that one file 😊


despo:
@koddsson staging deployment is failing and to sort it out we need this fix https://github.com/lautis/uglifier/issues/127#issuecomment-352224986

Any chance you can open a PR with it please as I can't approve and merge my PRs.

koddsson:
Doh! My bad. Glad you got it sorted.

koddsson:
@despo Do we have the staging commands in a script somewhere? I could add a test script that would make sure something like this doesn't happen again?

despo:
@koddsson our deployment is a heroku push so no, we don't. But that's why we have staging anyway, to catch any issues before they make it to production so I think it's okay. 
We don't have automated deployment between staging and production because we do want to manually verify that everything is working first, just in case.

PR #1334, opened 2020-05-23
Include the error image on instead of hotlinking

koddsson:
Right now we are serving up the image from GitHub instead just serving it from our own of server.

This change includes the `codebar-girl.gif` image in the project and links to it on the error and not found pages.broken 

You can see the image being broken in production by going to a non-existing url such as https://codebar.io/foobar.

image


koddsson:
Here's how it looks fixed

image

despo:
@koddsson can you please rebase.

Issue #1278, opened 2020-05-15
Don't show not invitable label on workshop page unless members cannot RSVP


despo:
## Description of the issue 📄
At the moment workshops who have an rsvp opening time set always get the non invitable text on the admin page as only the [`invitable`](https://github.com/codebar/planner/blob/master/db/schema.rb#L562) attribute is checked instead of both that the rsvp opening time is in the past **or** invitable is checked (look at `invitable_yet?`, defined in the `workshop_model`.

## Screenshots 📷

<img width="586" alt="Screen Shot 2020-05-15 at 15 52 29 1" src="https://user-images.githubusercontent.com/159200/82064222-60cc9c80-96c4-11ea-8e99-34afe2471a82.png">

## Steps to fix 🛠
- update check to be `rsvp opening times || invitable` for non invitable label
- update text when rsvps are available to specify since when `RSVPs opened up at XX:XX on XX/XX/XXXX` 

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

AFed29:
Got the docker setup working! Would quite like to take this up if that's ok! I'll check what was fixed in #1335 and try and resolve this issue 😄

matyikriszta:
@AFed29 yes, go ahead

Indrajitnaiya09:
@AFed29 Are you working on this? Else I'd like to pick it up.

AFed29:
Go for it! I've been swamped recently so haven't looked much at it!

PR #1335, opened 2020-05-24
Fix relative RSVP date text

koddsson:
Half of #1278. Just makes sure that we say `RSVPs opened at [date]` when the date was in the past and `RSVPs will open at [date]` when the date is in the future.

I had trouble with the rest of the issue. Finding where `invitable?` is defined is not apparent to me but I might just not understand how this codebase is structured. I've also never seen HAML before and I got to admit that I think I prefer plain old HTML or ERB instead of it 😅

### DIFF
--- a/app/views/admin/workshops/show.html.haml
+++ b/app/views/admin/workshops/show.html.haml
@@ -42,7 +42,10 @@
         %h3
           %small #{humanize_date(@workshop.date_and_time, @workshop.ends_at, with_time: true)}
         - if @workshop.rsvp_opens_at
-          RSVPs will open at
+          - if @workshop.open_for_rsvp?
+            RSVPs opened at
+          - else
+            RSVPs will open at
           = @workshop.rsvp_opens_at.strftime('%H:%M on %d/%m/%Y.')
           %br
           %br
--- a/app/views/admin/workshops/show.html.haml
+++ b/app/views/admin/workshops/show.html.haml
@@ -42,13 +42,12 @@
         %h3
           %small #{humanize_date(@workshop.date_and_time, @workshop.ends_at, with_time: true)}
         - if @workshop.rsvp_opens_at
-          - if @workshop.open_for_rsvp?
-            RSVPs opened at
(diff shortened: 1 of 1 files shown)


despo:
@koddsson thanks for the PR!

`invitable` is an [attribute of the workshop model](https://github.com/codebar/planner/blob/master/db/schema.rb#L562). 
We should however be using `invitable_yet?`, defined in the `workshop_model` to ensure all criteria is correclty checked as described in the issue.

app/views/admin/workshops/show.html.haml
```diff
@@ -42,10 +42,12 @@
         %h3
           %small #{humanize_date(@workshop.date_and_time, @workshop.ends_at, with_time: true)}
         - if @workshop.rsvp_opens_at
-          RSVPs will open at
-          = @workshop.rsvp_opens_at.strftime('%H:%M on %d/%m/%Y.')
-          %br
-          %br
+          %div{style: "margin-bottom: 2em"}
```
despo, line 45: This looks great but can the css be moved in a class in one of the sass files instead - if one does not exist with such a property. This is so we can both reuse it but also not litter the views with inline styling.

koddsson: Sure! I wasn't sure what to do since I thought foundation would have utility classes but I couldn't find them. Should CSS always be one-offs or do we want utility classes such as `margin-bottom-2`?

koddsson:
> `invitable` is an [attribute of the workshop model](https://github.com/codebar/planner/blob/master/db/schema.rb#L562).

Oooh I was looking in the model and controller. I'll make sure to look in the schema next time.

app/views/admin/workshops/show.html.haml
despo: I think a utility class would be better in this instance.

koddsson:
```suggestion
          %div.mb2
```

PR #1336, opened 2020-05-24
[WIP] listable testing

notapatch:
### Definitely WIP
I like the code because it is testing all of Listable objects but in so doing it's given me a few questions. Note the previous tests was under `spec/models/concerns/listable_spec `and it created a Workshop to test it - which led to small issues in the remaining Listable objects  (`Course`, `Meeting` and `Event`) as they got out of step.

It isn't ready for trying to merge without finding the answers to a couple of questions.

#### Remove Listable recent scope
- I couldn't find it being used so I removed #recent

#### Add Listable shared example
- the main commit this tests Listable on as many of the Listable types it can.

#### Remove from most_recent includes
- `include(:sponsors)` while it works for Workshop it doesn't work for other classes Meeting and Course has a 'sponsor' singular.  So I removed it.
-  `load` broke the build when most_recent is called in the `ChapterController#show`.

#### Remove course_spec tests which duplicate behaves_like_listable
- testing replaced by Listable shared example

#### Listable order desc on past and completed_since_yesterday
- This is where it gets interesting. The only previous testing was on a concern that instantiated Workshop. Workshop has a default_scope which is set to `:desc` so `past` and 'completed_since_yesterday' appear to have descending sort. However, all the other Objects (course, meeting, event) don't have a default_sort they all get :ascending sorts.  `past` is the only one that is used on an object other than Workshop. All the previous tests are counting the number of objects the scope finds and only one actually tests the order (completed_since_yesterday expects to be descending so I decided to make them descending).

#### Event is broken because the before_save isn't Time zone aware
- Event is broken the sorting is a problem (but will be fine with I know for sure what the sorting is meant to be) but also `before_save` is not time zone aware. The other classes that work use the `DateTimeConcern`.

---

## Questions
1. Listable#past && Listable#completed_since_yesterday - should they have a `desc` sort order?
2. Event#before_save - should that be using `DateTimeConcerns` and is it code that you would like to be fixed for TimeZones? (No idea how difficult that is)


despo:
@notapatch
1. Yes I think they should be ordered by date and time descending
2. We should probably timezone event. I think it was missed as the main priority and what we mostly use our app for is workshops. But as events don't need to belong to a particular chapter it would make more sense to store the timezone directly on the event model, so maybe for now we can just inherit and utilise the default app's timezone - so **I think** we can achieve that with a `time_zone` method on the Event model that calls `Time.zone`. That should enable using DateTimeConcerns within the model for setting everything else up. (FYI I am not sure about this as this is from just briefly inspecting the class)

Also if workshop does need `include(:sponsors)` for `most_recent` to be optimised, as I can see that the previous commit referenced a Bullet warning, I would leave it changed in `Listable` so it can be reused by the other models, but would probably overwrite `most_recent` within the actual Workshop model class with the previous definition.

As for `load` I really don't remember its purpose so maybe leave it out for now if it doesn't appear to be causing any issues.

notapatch:
Sorry I forgot an issue...

`upcoming` sorts `date_and_time` ascending for `Course`, `Meeting` and `Event`  but because of the `default_scope` it's different on `Workshop`.  For `Workshop` it sorts `descending`.  You can use `unscoped` to get around this for the test but that then introduces other sorting issues in the code base. In the short term I was going to remove the sort test on `upcoming`. If it's an important issue the only way I can see is to use `unscope` and`upcoming` in some manner or remove the `default_scope`.

`most_recent` is only used by `Workshop` to avoid overriding I thought I could just move `most_recent` in the original form into `Workshop`?

notapatch:
@despo - I have a treatment for Event with DateTimeConcerns with hard coded time_zone (like meeting)... which can be put through once other PR's clear.

PR #1338, opened 2020-05-25
Margin utilities

koddsson:
I was gonna use a margin utility in #1335 but found the margins utilities already declared to be in a bit of disarray. This is my attempt to create margin utilities that are consistent and intuitive to use. They follow the same naming scheme as was already in place and start from 0 and go up to 5em.

Not sure if this is helpful but here's a diagram explaining the system:

```
.m[tbrl]*[0-5]
^^ ^^^^  ^^^^^
|| |     |
|| |     + size from 0 to 5em.
|| + denote a direction, top, bottom, right or left. omit for all directions.
|+ m for margin.
+ class selector.
```

### DIFF
--- a/app/assets/stylesheets/partials/_margins.scss
+++ b/app/assets/stylesheets/partials/_margins.scss
@@ -9,3 +9,22 @@
 .mt1 {
   margin-top: 50px;
 }
+
+/* Margin utility classes */
+@for $i from 0 through 5 {
+  .margin-top-#{$i} {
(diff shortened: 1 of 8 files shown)


despo:
@koddsson thank you, I tried out a couple of the changes and it looks good.

And indeed our style is a mess and we are also using quite an old version of foundation as we need to first upgrade everything else before we can get to it, which doesn't help.

PR #1339, opened 2020-05-25
[WIP] (Fix) Allow original exception to be seen outside of production

notapatch:
Note: Changes creates an error in development on my local machine. I need to find out why ...
---
 - currently the exception block is giving the error page in every
   environment. This is bad because during development and testing
   it is helpful to see the stack trace.
 - fix allows you to see the stack trace in not production.
 - Note: if you have a staging environment so:
   Rails.env.staging?
   => true
   You would want to change this code. The alternative form is
   raise if Rails.env.production? || Rails.env.test?

### Before / Bad 👎 

```ruby
  1) viewing a course i can view a course's information
     Failure/Error: expect(page).to have_content course.title
       expected to find text "Et amet soluta repudiandae." in "Something went wrong\nIf the problem persists please get in touch with us on Slack.\nIn the meantime why don't you join our community on Slack?"
     # ./spec/features/viewing_a_course_spec.rb:10:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:80:in `block (2 levels) in <top (required)>'
```

### After / Good  👍 
```ruby
  1) viewing a course i can view a course's information
     Failure/Error: - if @course.past?
     
     ActionView::Template::Error:
       undefined method `past?' for #<CoursePresenter:0x00007fd162477b98>
     # ./app/views/courses/show.html.haml:8:in `_app_views_courses_show_html_haml__615509449984628271_70268647366560'
     # ./spec/features/viewing_a_course_spec.rb:8:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:80:in `block (2 levels) in <top (required)>'
     # ------------------
     # --- Caused by: ---
     # NoMethodError:
     #   undefined method `past?' for #<CoursePresenter:0x00007fd162477b98>
     #   ./app/views/courses/show.html.haml:8:in `_app_views_courses_show_html_haml__615509449984628271_70268647366560'
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,8 @@ class ApplicationController < ActionController::Base
   rescue_from Exception do |ex|
     Rollbar.error(ex)
     Rails.logger.fatal(ex)
+    raise unless Rails.env.production?
+
     respond_to do |format|
       format.html { render 'errors/error', layout: false, status: :internal_server_error }
       format.all  { render nothing: true, status: :internal_server_error }


app/controllers/application_controller.rb
```diff
@@ -4,6 +4,8 @@ class ApplicationController < ActionController::Base
   rescue_from Exception do |ex|
     Rollbar.error(ex)
     Rails.logger.fatal(ex)
+    raise unless Rails.env.production?
```
despo, line 7: This is helpful, however I think it's best to not condition code per environment and a better approach would be to rethink and rewrite this as we also have a problem with not being able to see the full trace of errors when they take place in production which is a much bigger issue than other environments.

notapatch: That's interesting. I guessed that in production you could see the full stack error in Rollbar log. I am wrong in this guess?

despo: I need to look this up more as it's been a while since coming to the conclusion that we should change this.

You can indeed see what's going on in rollbar, however as this is an open source project and we want anyone to be able to work on the project, we want better visibility of ongoing issues. We have rollbar configured to post on an public Slack channel but the details rendered there are not sufficient .

Eg screenshots below
Screen Shot 2020-05-25 at 19 12 45
Screen Shot 2020-05-25 at 19 12 38

notapatch: So some sort of ENV variable that you could set with Heroku so you can see a full crash on staging/production server instead of 404/500 error page? 🤔

PR #1340, opened 2020-05-25
DateTimeConcerns Review

notapatch:
- (Refactor) DateTimeConcern move code to private
  - improve readability
- DateTimeConcerns tested
   - main work
    - I use `travel_to` because it's easier to follow. If I use `Time.zone.now` the tests are constantly changing.
- (refactor) Prefer safe navigation operator over try

### DIFF
--- a/app/models/concerns/date_time_concerns.rb
+++ b/app/models/concerns/date_time_concerns.rb
@@ -23,7 +23,7 @@ module DateTimeConcerns
     end
 
     def time
-      date_and_time.try(:time)
+      date_and_time&.time
     end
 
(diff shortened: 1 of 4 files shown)


notapatch:
@despo  - I've done a re-treatment of the review which is everything except the `#past? `issue. With this in It will allow me to get `Event` to use the `DateTimeConcern` - which is what I was really interested in doing.

I can do the `#past?` commits as a separate PR later - the large effects of the changes mean it should be scrutinized separately.

notapatch:
rebased!  :shipit:

PR #1341, opened 2020-05-26
Meeting validations review

notapatch:


spec/models/meeting_spec.rb
```diff
@@ -4,23 +4,31 @@
   include_examples "Invitable", :meeting_invitation, :meeting
 
   context 'validations' do
-    subject { Meeting.new }
+    subject(:meeting) { Fabricate(:meeting) }
+    it {is_expected.to validate_presence_of(:date_and_time) }
+    it {is_expected.to validate_presence_of(:venue) }
+
+    context '#slug' do
+      it 'does not validate if slug present' do
```
despo, line 12: the wording here is a bit confusing.  Something like `fail when slug not present` would read better in this instance

spec/models/meeting_spec.rb
```diff
@@ -4,23 +4,31 @@
...
+        new_meeting.slug = ''
+        new_meeting.valid?
+
+        expect(new_meeting.errors[:slug]).to be_empty
+      end
 
-    it '#date_and_time' do
-      should have(1).error_on(:date_and_time)
-    end
+      it 'validates if slug present' do
```
despo, line 21: Maybe pass if slug is present?

notapatch:
@despo - I made the changes you requested and pressed the icon with two arrows making a circle. My hope is this is the correct way of you rechecking it.

PR #1343, opened 2020-05-27
Member remove unused methods

notapatch:
I've used a tool that checks for unused methods (rails_best_practices). Member was the most interesting because it's complicated.


notapatch:
rebased!
[The build failed the first time](https://travis-ci.org/github/codebar/planner/builds/692250919) whatever caused the Timezone issue is wasn't removing some unused methods from Member. I have run it a couple of times since and it worked.

I'll keep thinking about it but for me ... it remains a mystery build break 👻

despo:
@notapatch no idea!

PR #1345, opened 2020-05-28
Refactoring admin controllers with time attributes

notapatch:
I think the overview of these commits is to try and get these controllers and forms looking as similar as possible so that when the Event comes in it will join a more consistent style. 

The event changes will come after this PR.

[I've only seen the word Flocking in Sandi Metz's 99 bottles of OOPs I have no idea if it's a general term or not... basically make things look the same and see where that takes you with the refactoring.](https://www.sandimetz.com/99bottles).

Note: "Admin controllers with time" is arbitrary - and really I think all the controllers should try to flock - say by having Scaffold order of actions.

---
### Workshop update tests
- modernise tests to a 'house style'

### MeetingController strong_params remove date_and_time
- date_and_time did not need to return to MeetingController

### Admin forms with time use safe navigation
- safe navigation is the Ruby language standard and looks "neater"

### Admin controllers with time ordered params by form appearance
- @despo - would you prefer it that they were all put into a few lines to appease Rubocop?
                    The cop is about avoiding confusingly long code blocks but in this case it
                     is a list of params. My preference would be to add it as an exception to 
                     Rubocop.yml.
                     **Examples** - it's hard to find controllers with many params I don't know why but when they do they put them in one per row:
                     [Gitlab](https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/controllers/groups_controller.rb#L206)
                     [Discourse](https://github.com/discourse/discourse/blob/master/app/controllers/groups_controller.rb#L519)
                     [Diaspora](https://github.com/diaspora/diaspora/blob/develop/app/controllers/users_controller.rb#L141)

### Admin controllers with time have actions in scaffold order
- there is a lot of "variation" between different controllers and a lot of that comes from the order of appearance of actions. I didn't feel the different orders helped my understand the controller and while I thought having a consistent order would help.

## Edit:
Reviewed code in other well known projects to see how they handled large lists of params

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -2,6 +2,17 @@ class Admin::EventsController < Admin::ApplicationController
   before_action :set_event, only: [:show]
   before_action :find_event, only: %i[edit update]
 
+  def show
+    authorize @original_event
+
+    @address = AddressPresenter.new(@event.venue.address) if @event.venue.present?
+    @attending_students = InvitationPresenter.decorate_collection(@original_event.attending_students)
+    @attending_coaches = InvitationPresenter.decorate_collection(@original_event.attending_coaches)
+    @host_address = AddressPresenter.new(@event.venue.address)
+
+    return render text: @event.attendees_csv if request.format.csv?
+  end
+
   def new
     @event = Event.new
   end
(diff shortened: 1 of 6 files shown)


notapatch:
- Coverage remained the same at 94.286% 
Correct to 3 decimal places, huh.

PR #1348, opened 2020-05-30
Basic Rubocop over spec

notapatch:
I would like to treat the spec directory with the same care as the app directory. Currently, there is no linting over the spec directory at all. These are a few Rubocop defaults over the spec directory.

When the basics are fixed my aim would be to apply `rubocop-rspec`.

All this would improve consistency in the testing code.

### DIFF
--- a/spec/support/shared_examples/behaves_like_invitable.rb
+++ b/spec/support/shared_examples/behaves_like_invitable.rb
@@ -49,12 +49,11 @@ RSpec.shared_examples 'Invitable' do |invitation_type, invitable_type|
 
     it 'rejects banned attending students' do
       invitation_to_banned_student = Fabricate(invitation_type,
-                                        member: Fabricate(:banned_member),
-                                        role: 'Student',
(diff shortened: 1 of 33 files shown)


spec/support/shared_examples/behaves_like_invitable.rb
```diff
@@ -49,12 +49,12 @@
...
-                                        role: 'Student',
-                                        invitable_type => invitable,
-                                        attending: true)
+                                               member: Fabricate(:banned_member),
+                                               role: 'Student',
+                                               invitable_type => invitable,
+                                               attending: true)
 
-                                        # byebug
+      # byebug
```
matyikriszta, line 57: @notapatch I think this could be deleted, I doubt we need this here.

notapatch: great... removed

notapatch:
@matyikriszta - fixed and rebased

PR #1351, opened 2020-06-04
(enhancement) Fabricator ends_at calculated from start time

notapatch:
- date_and_time (the starting time) and ends_at is a time range
 - most of the time you want ends at to be a few hours after
   starts at. Most of the tests are setup without the programmer
   worrying about this. By having the ends_at calculate from the
   start time it will be a consistently sensible default.
 - You can now Fabricate a workshop with a starting time and not
   worry about explicitly setting an ending time.

 - the wrinkle with the code is that Fabricator breaks if
   date_and_time is set to nil. This happened in one test and I
   would prefer to fix that test rather than complicate the
   Fabricator (which are untested) with a nil check


despo (approved):
👍

PR #1362, opened 2020-07-03
Some wording tweaks on the job board donation page

KimberleyCook:


app/views/donations/new.html.haml
```diff
@@ -7,17 +7,16 @@
 .stripe.reverse
   .row
     .large-12.columns
-      %h2 Donations
+      %h2 Job board donation
```
matyikriszta, line 10: @KimberleyCook I think we should frame this as a payment now as opposed to a donation. We are providing a service people are paying for.

KimberleyCook:
@matyikriszta finally got round to updating the PR ha

PR #1376, opened 2020-08-05
Improve layout of _venue partial if venue has no address 

matyikriszta:
## Description
This PR improves the layout and the content shown in the _venue partial when the workshop's/event's etc. venue has no address (aka all address fields are empty). It also fixes a small bug in the` AddressPresenter` `to_html` method wherein a dangling comma would show for venues with no address.

## Status
Ready for Review

## Screenshots
**With address**
Screenshot_2020-08-04 codebar(1)

**Without address**
Screenshot_2020-08-04 codebar

### DIFF
--- a/app/views/shared/_venue.html.haml
+++ b/app/views/shared/_venue.html.haml
@@ -1,19 +1,21 @@
+- formatted_address = address.to_html
+- section_title = formatted_address.empty? ? "Host" : "Venue"
+
 #venue
-  .small-12.column
-    %h3 Venue
   .medium-6.columns
+    %h3= section_title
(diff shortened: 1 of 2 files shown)


app/views/shared/_venue.html.haml
```diff
@@ -1,19 +1,21 @@
+- formatted_address = address.to_html
+- section_title = formatted_address.empty? ? "Host" : "Venue"
```
matyikriszta, line 2: Dirty way to change  the section title to host if the venue has no address. Should remove this once we add proper virtual support to Events.

PR #1377, opened 2020-08-06
add levels to sponsorships

jabawack81:
This PR is adding the possibility to have four levels of sponsors for an event `standard` (similar to the old level), `bronze`, `silver` and `gold`.

To accomplish that I added a column in the `sponsorship` table called used to connect events and sponsors and it's used to assign the level of the sponsor based on the content [see](https://github.com/jabawack81/planner/blob/9b2a235b49bb9634a3193b15dce2d6a8b86babe7/app/models/event.rb#L10), and I edited four views to edit and display this new feature:
add the fields in the new and edit form:
edit/new event
add the new levels in the show event page in the admin section:
show page in admin section
and the show event page for the user
show event page for the user

I added a new validation to the event model to prevent having a sponsor as more than one level with this [sponsors_uniqueness](https://github.com/jabawack81/planner/blob/9b2a235b49bb9634a3193b15dce2d6a8b86babe7/app/models/event.rb#L29), first, it detects any duplicated sponsors and then checks if any of the four levels include it and add an error to the level that does, I realise this implementation is quite intensive but considering that it will be invoked only on the creation or edit of events, that I would imagine will happen once or twice a week but is giving a level of details for the error.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -25,7 +25,7 @@ class Event < ActiveRecord::Base
   validate :gold_sponsors_uniqueness
 
   before_save do
-    begins_at = Time.zone.parse(self.begins_at)
+    begins_at = Time.parse(self.begins_at)
     self.date_and_time = date_and_time.change(hour: begins_at.hour, min: begins_at.min)
   end
 
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -7,10 +7,10 @@ class Event < ActiveRecord::Base
 
   belongs_to :venue, class_name: 'Sponsor'
   has_many :sponsorships
-  has_many :sponsors, -> { where('sponsorships.level'=> nil) }, through: :sponsorships, source: :sponsor
-  has_many :bronze_sponsors, -> { where('sponsorships.level'=> 'bronze') }, through: :sponsorships, source: :sponsor
-  has_many :silver_sponsors, -> { where('sponsorships.level'=> 'silver') }, through: :sponsorships, source: :sponsor
-  has_many :gold_sponsors, -> { where('sponsorships.level'=> 'gold') }, through: :sponsorships, source: :sponsor
+  has_many :sponsors, -> { where('sponsorships.level' => nil) }, through: :sponsorships, source: :sponsor
+  has_many :bronze_sponsors, -> { where('sponsorships.level' => 'bronze') }, through: :sponsorships, source: :sponsor
+  has_many :silver_sponsors, -> { where('sponsorships.level' => 'silver') }, through: :sponsorships, source: :sponsor
+  has_many :gold_sponsors, -> { where('sponsorships.level' => 'gold') }, through: :sponsorships, source: :sponsor
   has_many :organisers, -> { where('permissions.name' => 'organiser') }, through: :permissions, source: :members
   has_and_belongs_to_many :chapters
   has_many :invitations
@@ -22,21 +22,17 @@ class Event < ActiveRecord::Base
   validate :sponsors_uniqueness
 
   before_save do
-    begins_at = Time.parse(self.begins_at)
+    begins_at = Time.zone.parse(self.begins_at)
     self.date_and_time = date_and_time.change(hour: begins_at.hour, min: begins_at.min)
   end
 
   def sponsors_uniqueness
     ids = sponsorships.reject(&:marked_for_destruction?).map(&:sponsor_id)
     duplicated = ids.select { |e| ids.count(e) > 1 }
-    standards_ids = sponsors.map(&:id)
-    bronze_ids = bronze_sponsors.map(&:id)
-    silver_ids = silver_sponsors.map(&:id)
-    gold_ids = gold_sponsors.map(&:id)
-    errors.add(:sponsors, :duplicated_sponsor) unless (standards_ids & duplicated).empty?
-    errors.add(:bronze_sponsors, :duplicated_sponsor) unless (bronze_ids & duplicated).empty?
-    errors.add(:silver_sponsors, :duplicated_sponsor) unless (silver_ids & duplicated).empty?
-    errors.add(:gold_sponsors, :duplicated_sponsor) unless (gold_ids & duplicated).empty?
+    errors.add(:sponsors, :duplicated_sponsor) unless (sponsors.map(&:id) & duplicated).empty?
+    errors.add(:bronze_sponsors, :duplicated_sponsor) unless (bronze_sponsors.map(&:id) & duplicated).empty?
+    errors.add(:silver_sponsors, :duplicated_sponsor) unless (silver_sponsors.map(&:id) & duplicated).empty?
+    errors.add(:gold_sponsors, :duplicated_sponsor) unless (gold_sponsors.map(&:id) & duplicated).empty?
   end
 
   def to_s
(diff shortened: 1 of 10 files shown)


KimberleyCook:
@matyikriszta would you be able to review this PR today please?

app/models/event.rb
```diff
@@ -7,7 +7,10 @@ class Event < ActiveRecord::Base
 
   belongs_to :venue, class_name: 'Sponsor'
   has_many :sponsorships
-  has_many :sponsors, through: :sponsorships
+  has_many :sponsors, -> { where('sponsorships.level' => nil) }, through: :sponsorships, source: :sponsor
+  has_many :bronze_sponsors, -> { where('sponsorships.level' => 'bronze') }, through: :sponsorships, source: :sponsor
+  has_many :silver_sponsors, -> { where('sponsorships.level' => 'silver') }, through: :sponsorships, source: :sponsor
+  has_many :gold_sponsors, -> { where('sponsorships.level' => 'gold') }, through: :sponsorships, source: :sponsor
```
codeincontext, line 13: This is exactly how I would have done it 👍

app/models/event.rb
```diff
@@ -16,9 +19,13 @@ class Event < ActiveRecord::Base
   validates :slug, uniqueness: true
   validate :invitability, if: :invitable?
   validates :coach_spaces, :student_spaces, numericality: true
+  validate :sponsors_uniqueness
+  validate :bronze_sponsors_uniqueness
+  validate :silver_sponsors_uniqueness
+  validate :gold_sponsors_uniqueness
```
codeincontext, line 25: It seems to me like a sponsor is probably unique across all levels. Maybe @KimberleyCook can clarify is someone can be a gold and a bronze sponsor, for example.

I think that could reduce the amount of code needed as well

app/models/event.rb
```diff
@@ -16,9 +19,13 @@ class Event < ActiveRecord::Base
...
 
   before_save do
-    begins_at = Time.parse(self.begins_at)
+    begins_at = Time.zone.parse(self.begins_at)
```
codeincontext, line 28: I think this should be tested and critiqued in a different PR, just in case it's that say for a reason

app/models/event.rb
```diff
@@ -76,4 +83,41 @@ def coach_emails
...
+    when :bronze then bronze_sponsors.any?
+    when :standard then sponsors.any?
+    else sponsorships.any?
+    end
+  end
+
+  private
+
+  def duplicated_sponsors
+    @duplicated_sponsors ||= fetch_duplicated_sponsors
```
codeincontext, line 100: I think changing the validations to no longer be per-level would mean this no longer needs to be memoized

app/views/events/show.html.haml
```diff
@@ -53,18 +53,20 @@
...
+            %h3.text-center Gold 
+            = render partial: "sponsors", object: @event.gold_sponsors
+          - if @event.sponsors?(:silver)
+            %h3.text-center Silver
+            = render partial: "sponsors", object: @event.silver_sponsors
+          - if @event.sponsors?(:bronze)
+            %h3.text-center Bronze
+            = render partial: "sponsors", object: @event.bronze_sponsors
+          - if @event.sponsors?(:standard)
+            %h3.text-center Standard
```
codeincontext, line 68: If the event only has standard sponsors, should it still show the title? Question for @KimberleyCook

app/views/admin/events/edit.html.haml
```diff
@@ -36,9 +36,18 @@
...
+          .large-3.medium-6.small-12.columns
+            %br
+            = f.association :bronze_sponsors, input_html: { data: { placeholder: 'Select bronze sponsors' }}, collection: Sponsor.all
+          .large-3.medium-6.small-12.columns
+            %br
+            = f.association :silver_sponsors, input_html: { data: { placeholder: 'Select silver sponsors' }}, collection: Sponsor.all
+          .large-3.medium-6.small-12.columns
             %br
-            = f.association :sponsors, input_html: { data: { placeholder: 'Select sponsors' }}
+            = f.association :gold_sponsors, input_html: { data: { placeholder: 'Select sponsors' }}, collection: Sponsor.all
```
codeincontext, line 50: Is it as easy to have a list of sponsors that each have a dropdown for a level? I think that's more intuitive for me, though it's absolutely not a blocker

app/models/event.rb
codeincontext: Update: I just saw `sponsors_uniqueness` above which does what I was expecting. In that case I don't think validating inside levels hurts

app/models/event.rb
jabawack81: sorry I have no idea how this ended up in there, I imagine I corrected it with `rubocop` by mistake, let me restore it as it was.

app/models/event.rb
jabawack81: the validation doesn't discriminate between levels, so it will fail either if the duplication is at the same level or in different levels

app/models/event.rb
```diff
@@ -7,7 +7,10 @@ class Event < ActiveRecord::Base
 
   belongs_to :venue, class_name: 'Sponsor'
   has_many :sponsorships
-  has_many :sponsors, through: :sponsorships
+  has_many :sponsors, -> { where('sponsorships.level' => nil) }, through: :sponsorships, source: :sponsor
```
despo, line 10: Is there any value in keeping 'standard' sponsors as I see this causing confusion moving forward.
Adding a migration moving all the existing sponsorships to the lowest available level would enable us to display sponsors in past events and not keep functionality that we don't want to be utilised any longer.

jabawack81: In my first version of this PR the new column in the connection table had the default to `bronze`.
I kept the standard level of sponsorship for retro compatibility and to allow future events without levels but if all the future events will have them I will simply remove the relationship and reinstate the default value if instead, the problem is more the cleanness of the form we could add a switch to hide the standard or levelled input.

despo: Can we then split this up so that
1. In the form either a 'standard' sponsor can be set, or 'gold, bronze or silver' - **add some information text** letting admin/organisers know that they should either select `standard' sponsors or leveled sponsors
2. Condition the views so that either standard sponsors are rendered (displayed a level here would also be unecessary so drop the standard title in the member view, maybe keep it in the admin view) or different sponsorship tiers and the relevant sponsors

It would have been nice to have this association to be implemented [using nested attributes](https://www.pluralsight.com/guides/ruby-on-rails-nested-attributes) which would simplify the views as well as the relationships and model associations, but it would be a bit of work and I think we need this in ASAP so it's maybe something to consider for future optimisation of this functionality.

jabawack81: that makes sense, I tried to keep the view as similar as it is currently but yes, with nested attributes could make everything cleaner.

for the new sponsor section, I'm thinking to replace the four dropdowns with: 
on top a checkbox to select levelled or standard sponsors then under it a three columns grid with a line for each sponsor and on the first column a dropdown menu for the sponsors and a series of three radio button for the level if selected or hide it if not needed then a button to remove the sponsors, and after the grid a button to add a new sponsor.

Because this functionality is needed immediately I think we could merge this PR as it is and I'll open a new one for the new form.

despo: @jabawack81 ok but can we add some information text informing editors that they should either select standard or any of the other levels so that regardless of any future work being done it is clear why those four levels are set.

Also can you please rebase! Thanks.

jabawack81: done.

Screenshot_2020-08-11_11-24-33

Sorry I resolved the conversation by mistake.

despo:
@jabawack81 Thanks 💯

app/models/event.rb
despo: No worries. Thanks for making the change.

despo:
:shipit: https://codebar.io/events/uncodebar-6

PR #1386, opened 2020-08-20
Refactor event form

jabawack81:
# **WORK IN PROGRESS DO NOT MERGE**

Like discussed with @despo on the comment section of the pr #1377 I started working on a refactor of the form for events.

I'm at a good point visually and I could use some feedback:

Screenshot_2020-08-20 codebar

@despo was this what you were talking about with using nested attributes?

at the moment the `Add Sponsor` and `Remove Sponsor` are not working but the type of sponsorship, sponsors, and sponsor level are working.

* the radio buttons for the sponsors level as visible and enabled only when `Levelled` is active:

  Screenshot_2020-08-20 codebar(1)
  Screenshot_2020-08-20 codebar(2)

  I had to use radio buttons instead of a select because in my first attempt I couldn't find a good way to disable it when the standard type is selected.

For the moment each of the row for a sponsor has three sections with only the class `large-3` and so they'll expand to fill all the row for any other media query making the form a little hard to read I was thinking to maybe add an `HR` element visible only in smaller screens between each sponsor.


jabawack81:
@despo I've updated this PR and the functionality is ready to be tested:
Screenshot_2020-12-20_15-45-39

there are still a couple of problems with the CSS:

when in a smaller screen the elements are all on top of each other, I would like to space them a little but I'm not sure what would be the best approach, I was thinking to add a div with a `<br>` between the elements hidden in the large screens or maybe add a media query to add some padding to the divs.
Screenshot_2020-12-20_15-46-02

when resizing the window the select inputs don't resize, not sure if there is a solution for this out of having an event listener for the resize of the window but I'm worried this is overkill.
Screenshot_2020-12-20_15-46-19

in the meantime, I'll work on the failing test ( CI / Tests (pull_request)  / CodeFactor / codeclimate )

PR #1401, opened 2020-09-23
Change donations to payments

despo:
- Renames donations to payments and removes sponsors from page as they are no longer relevant
- Links new job to payments instead of donations
- Only logged in users can get to the payments page


KimberleyCook (approved):
This looks good to me

app/views/member/jobs/new.html.haml
```diff
@@ -31,7 +31,7 @@
               %label{for: 'short'}= t('member.jobs.new.short')
             %li
               %input#donation{type: 'checkbox'}
-              %label{for: 'donation'}= t('member.jobs.new.pay.html', link: new_donation_path)
+              %label{for: 'donation'}= t('member.jobs.new.pay.html', link: new_payment_path)
```
matyikriszta, line 34: @despo do we have to update the for attribute here?

despo: @matyikriszta it doesnt break anything as it maps to the #donation id on the line above, but I will update before I merge for clarity as it's now a payment and not a donation. Good catch, I didnt even notice.

PR #1413, opened 2020-10-20
Add option to unpublish job post

matyikriszta:
## Description
This PR adds the option to unpublish job posts once they've been published. Unpublishing a job post resets the `published_on` and `approved_by` fields to nil and sets the job status to pending.

## Status
Ready for Review

## Related Trello ticket
https://trello.com/c/VpGr7xPY

## Screenshots
Screenshot 2020-10-20 at 11 03 29 am

### DIFF
--- a/app/policies/job_policy.rb
+++ b/app/policies/job_policy.rb
@@ -12,6 +12,6 @@ class JobPolicy < ApplicationPolicy
   end
 
   def unpublish?
-    user.has_role?(:admin) || Chapter.find_roles(:organiser, user).any?
+    user.has_role?(:admin)
   end
 end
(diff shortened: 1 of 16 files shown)


KimberleyCook (approved):
Nice 🎉 🎉

app/policies/job_policy.rb
```diff
@@ -10,4 +10,8 @@ def show?
   def approve?
     user.has_role?(:admin) || Chapter.find_roles(:organiser, user).any?
   end
+
+  def unpublish?
+    user.has_role?(:admin) || Chapter.find_roles(:organiser, user).any?
```
despo, line 15: Organisers should not be able to manage jobs so even though we have not fixed the rest of the policies to reflect that I think we should make sure to remove it from any new functionality.

despo (changes requested):

app/policies/job_policy.rb
matyikriszta: @despo ok, sure, can update the new policy.

despo (approved):
LGTM 👍

PR #1416, opened 2020-10-26
Little PR adding a link to the codebar stories blog in our footer

KimberleyCook:
I noticed we were not linking to this blog anywhere on our website, so I've added it to the footer.


despo (approved):
LGTM 👍

PR #1418, opened 2020-11-05
Update homepage, main events page and single workshop page to use Bootstrap 5

KimberleyCook:
Even though there are lots of files it isn't that bad I promise.

There are notes about deleting CSS in some of the files that I'm unsure if I can delete just yet as I'm not 100% sure whether deleting it will break another page, so notes have been added as a very last task to delete all the unused CSS (that I don't delete as I go).

Updated homepage (with new dropdown menu)
screencapture-localhost-3000-2020-11-05-17_12_27

Updated single workshop page

screencapture-localhost-3000-workshops-4-2020-11-05-17_13_11

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -452,6 +452,7 @@ DEPENDENCIES
   sass-rails (~> 5.0.1)
   shoulda-matchers (~> 4.3)
   simple_form
+  sprockets-rails (= 3.2.1)
   stripe
   turbolinks
   tzinfo-data
--- a/app/views/events/_events.html.haml
+++ b/app/views/events/_events.html.haml
@@ -1,4 +1,4 @@
-.container
+.container-fluid
   .event
     - grouped_events.each do |date, workshops|
       .row
--- a/app/views/events/index.html.haml
+++ b/app/views/events/index.html.haml
@@ -2,15 +2,14 @@
 
 .container-fluid
   .row
-    .col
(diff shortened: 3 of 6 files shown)


Gemfile
```diff
@@ -6,11 +6,12 @@ gem 'rails', '4.2.11.3'
 gem 'acts-as-taggable-on', '~> 5.0'
 gem 'carrierwave'
 gem 'carrierwave-ftp', require: 'carrierwave/storage/sftp'
-gem 'compass-rails', git: 'https://github.com/Compass/compass-rails'
+gem 'compass-rails'
```
KimberleyCook, line 9: I had to update this to get the new Bootstrap working, let me know @despo if you think it'll be an issue.

despo: No idea! I thought it had to do with foundation but that does seem to be the case (related commit: 
https://github.com/codebar/planner/commit/7b57dc5fc9d6c7baef5ebb3f58b305d14be9a94b)

We should probably try to push this fix on its own to staging and see if it causes any issues with deployment. If not we should be fine.

KimberleyCook: Sounds good, I'll do a separate PR for this to test staging 👍

KimberleyCook: Separate PR here https://github.com/codebar/planner/pull/1421

KimberleyCook:
Screwed up a rebase so closing this PR, and opening a new one!

PR #1419, opened 2020-11-05
Remove github-markdown

gnclmorais:
I’ve noticed we have two gems responsible for rendering Markdown into HTML. Since `github-markdown` has been [deprecated][deprecated] and we have [`redcarpet`][redcarpet], I thought it would be a good idea to rely on the latter and remove the former.

[deprecated]: https://rubygems.org/gems/github-markdown
[redcarpet]: https://rubygems.org/gems/redcarpet


config/initializers/redcarpet.rb
```diff
@@ -0,0 +1,3 @@
+require 'redcarpet'
+
+Rails.configuration.redcarpet = Redcarpet::Markdown.new(Redcarpet::Render::HTML.new)
```
gnclmorais, line 3: This is a simple way of creating a _single_ instance of Redcarpet and reuse it everywhere else, instead of creating new objects every time we want to render Markdown into HTML. Happy to hear alternatives! 🙂

despo:
@gnclmorais this looks great, good catch.  👍 

But maybe we can should just remove redcarpet entirely and switch to [commonmarker](https://github.com/gjtorikian/commonmarker) which supports Github Flavored Markdown.

despo:
@gnclmorais still happy to approve as it is, as we can do that in a separate PR! So if you just rebase I'm happy to merge and it may even make today's release :)

despo (approved):
👍

gnclmorais:
Hey @despo! I’m glad you mentioned that, I had the same thought but, like you also suggested, I think it’s something for a follow-up PR. There are two or three places where we render Markdown which have different settings of how/what to render, so that is something to address on itself. I’m happy to follow this pull request with another one switching to [`gjtorikian/commonmarker`](https://github.com/gjtorikian/commonmarker). 😄

gnclmorais:
Hey @despo, I took a look at it and it was surprisingly smooth, so here it is: https://github.com/codebar/planner/pull/1423 🚀

despo:
@gnclmorais awesome, thank you 💯

PR #1423, opened 2020-11-07
Render Markdown with commonmarker

gnclmorais:
### ~~Review https://github.com/codebar/planner/pull/1419 first~~ ✔

As suggested by @despo, I took a stab at replacing `redcarpet` gem with [`commonmarker`](https://github.com/gjtorikian/commonmarker) to see how troublesome would be, and… end up being pretty simple! After #1419 being merged into `master`, I’ll rebase this PR and it will be ready for review. 😄 

**Note:** I see some refactoring opportunities in this code, not sure if we’re interested in taking them. For example, homogenising the way we render Markdown straight into a view. We could move `CommonMarker.render_html(…)` into a helper so we could simply have `render_markdown(…)` in the views, abstracting use from the specific Markdown engine. I also wonder if we can get rid of those `raw` and `.html_safe`… 🤔

### DIFF
--- a/app/views/admin/portal/guide.html.haml
+++ b/app/views/admin/portal/guide.html.haml
@@ -9,28 +9,28 @@
 
       %p Chapter organisers can create and modify sponsors and workshops only for chapters that they are permissioned. To make a member into an organiser of a chapter,
 
-      = raw @markdown.render("`Member.find(:id).add_role(:organiser, Chapter.find(:id))`")
+      = raw CommonMarker.render_html("`Member.find(:id).add_role(:organiser, Chapter.find(:id))`")
 
       %p Meeting organisers can create and modify monthlies. To make a member into a meeting organiser,
 
-      = raw @markdown.render("`Member.find(:id).add_role(:organiser, Meeting`")
+      = raw CommonMarker.render_html("`Member.find(:id).add_role(:organiser, Meeting`")
 
       %p Admins can create and modify sponsors, workshops, meetings and events across all chapters. Admins can also create new chapters. To permission a member as an admin,
 
-      = raw @markdown.render("`Member.find(:id).add_role(:admin)`")
+      = raw CommonMarker.render_html("`Member.find(:id).add_role(:admin)`")
 
       %p Similarly, to remove privileges,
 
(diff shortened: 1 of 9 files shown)


app/views/admin/portal/guide.html.haml
```diff
@@ -9,28 +9,28 @@
 
       %p Chapter organisers can create and modify sponsors and workshops only for chapters that they are permissioned. To make a member into an organiser of a chapter,
 
-      = raw @markdown.render("`Member.find(:id).add_role(:organiser, Chapter.find(:id))`")
+      = raw CommonMarker.render_html("`Member.find(:id).add_role(:organiser, Chapter.find(:id))`")
```
despo, line 12: can we not just use `dot_markdown(text)` in the entire guide as well? Then we would have only one place referencing common marker.

despo (changes requested):
Looks good but why not just call `dot_markdown()` instead of `raw(CommonMarket.html())` from the guide as well then? Not sure I'm missing something here?

despo:
@gnclmorais Thanks for your PR, it looks great. I have left some minor feedback.

As for not using `html_safe` or `raw` I'm not sure how we could do this. I understand that it can be classed as not safe, however the places that we allow for this to be called either provide static content, eg the guide, or the content is provided by authorised users e.g. organisers or admin of the app, which we trust.

We can however configure `commonmarker` to use the tagfilter extension which should filter unsafe tags like <script> 

This can be done by updating the call to common marker to be something like this:

`CommonMarker.render_html(text, :DEFAUT, %i[tagfilter]).html_safe).html_safe`

gnclmorais:
Hey @despo! I took a look at your (great!) suggestion and it seems that `CommonMarker.render_html(text, :DEFAUT)` is enough to turn…
```html
<scrip>alert("Hi");</script>
```
… into this…
```html
<p><!-- raw HTML omitted -->alert(&quot;Hi&quot;);<!-- raw HTML omitted --></p>
```
… making `%i[tagfilter]` irrelevant. Seems like the default rendered is already safe enough.

despo:
@gnclmorais sounds good to me!

despo (approved):
LGTM 💯

PR #1435, opened 2020-11-16
Update shared venue partial to Booststrap

KimberleyCook:
Updating the shared venue partial to Bootstrap.

I promised future PRs would be smaller @despo and @matyikriszta :D

### DIFF
--- a/app/views/shared/_venue.html.haml
+++ b/app/views/shared/_venue.html.haml
@@ -1,8 +1,8 @@
 - formatted_address = address.to_html
 - section_title = formatted_address.empty? ? "Host" : "Venue"
 
-#venue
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -129,8 +129,8 @@
     t.string   "name"
     t.string   "surname"
     t.string   "email"
-    t.boolean  "mailing_list_consent",        default: false
-    t.string   "token",                                       null: false
+    t.boolean  "mailing_list_consent", default: false
+    t.string   "token", null: false
```
KimberleyCook, line 133: This was failing the rake tests, so had to update.

matyikriszta (approved):
👍

PR #1436, opened 2020-11-16
Bootstrap organisers and avatar shared partials

KimberleyCook:
Bootstrap organisers and avatar shared partials.

### DIFF
--- a/app/views/members/_avatar_listing.html.haml
+++ b/app/views/members/_avatar_listing.html.haml
@@ -1,10 +1,5 @@
-%ul.small-block-grid-2.medium-block-grid-4
-  - member.each do |member|
-    %li.text-center
-      =link_to twitter_url_for(member.twitter), class: 'user-link' do
-        =image_tag(member.avatar(56), class: 'th radius', title: member.full_name, alt: member.full_name)
-      %br
-      %br
-      = member.full_name
-      - if show_info
-        = member.mobile
+- member.each do |member|
+  .col-sm-4.col-md-3.mb-4
+    =link_to twitter_url_for(member.twitter), class: 'user-link' do
+      =image_tag(member.avatar(56), class: 'rounded-circle', title: member.full_name, alt: member.full_name)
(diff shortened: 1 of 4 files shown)


app/views/members/_avatar_listing.html.haml
```diff
@@ -1,10 +1,5 @@
...
-      %br
-      %br
-      = member.full_name
-      - if show_info
-        = member.mobile
+- member.each do |member|
+  .col-sm-4.col-md-3.mb-4
+    =link_to twitter_url_for(member.twitter), class: 'user-link' do
+      =image_tag(member.avatar(56), class: 'rounded-circle', title: member.full_name, alt: member.full_name)
+    %p.mt-3.mb-1= member.full_name
```
KimberleyCook, line 5: I removed the telephone number, as I'm not sure we should be sharing this with everyone on the workshop page publically, only in the admin area for the page once someone has RSVP'd.

db/schema.rb
```diff
@@ -129,8 +129,8 @@
     t.string   "name"
     t.string   "surname"
     t.string   "email"
-    t.boolean  "mailing_list_consent",        default: false
-    t.string   "token",                                       null: false
+    t.boolean  "mailing_list_consent", default: false
+    t.string   "token", null: false
```
KimberleyCook, line 133: This was failing the rake tests again, so had to update.

Once this PR or [this one](https://github.com/codebar/planner/pull/1435) are merged in, I will no longer need to update this.

app/views/members/_avatar_listing.html.haml
gnclmorais: Sounds good 👍

despo: @KimberleyCook could we apply this separately in a new PR and not as part of the boostrap changes though.

matyikriszta (approved):
I think this looks good now 👍

PR #1439, opened 2020-11-18
Change Ban button to Suspend

despo:
- the requirement comes from organiser feedback, as the word Ban is scary
- this PR also adds an admin locale file so that we can begin moving admin section translations there


matyikriszta (approved):
👍

KimberleyCook (approved):
Nice!

PR #1441, opened 2020-11-20
Remove sponsor model contact details and members

despo:
- remove unused columns and association table
- remove member_contact fabricator
- remove member migration rake task that has already run in production


matyikriszta (approved):
👍

PR #1442, opened 2020-11-20
Bootstrap workshop page actions partial

KimberleyCook:
FYI the new buttons look like this

Screenshot 2020-11-20 at 13 50 53

### DIFF
--- a/app/views/workshops/_actions.html.haml
+++ b/app/views/workshops/_actions.html.haml
@@ -1,6 +1,5 @@
 - if !@workshop.invitable_yet?
-  %p
-    %label.label.info= t('workshops.not_open_for_rsvp')
+  %p.badge.bg-primary= t('workshops.not_open_for_rsvp')
 - else
   - if @workshop.past? and !logged_in?
     %p= t('workshops.sign_up_to_be_invited')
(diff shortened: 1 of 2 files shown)


matyikriszta (approved):
👍

app/views/workshops/_actions.html.haml
```diff
@@ -15,19 +14,20 @@
 
 - if logged_in? && @workshop.future? && (@workshop.invitable || @workshop.open_for_rsvp?)
   - if @workshop.attendee?(current_user)
-    = button_to "Manage your invitation", rsvp_workshop_path(@workshop), method: :post, class: 'button round '
+    = button_to "Manage your invitation", rsvp_workshop_path(@workshop), method: :post, class: 'btn btn-primary'
   - else
     - if current_user.student? && current_user.coach?
-      = button_to "Attend as a student", rsvp_workshop_path(@workshop, role: "Student"), method: :post,class: 'button round '
-      = button_to "Attend as a coach", rsvp_workshop_path(@workshop, role: "Coach"), method: :post,class: 'button round '
+      %div.d-flex
```
despo, line 20: Looks good. We can skip `%div` though as we don't tend to use it across our views unless there is no css class specified as `.class-name` translates by default to `<div class="class-name"></div>`

despo (approved):
👍

app/views/workshops/_actions.html.haml
KimberleyCook: Ah yes of course, nice catch. Will update.

Issue #1317, opened 2020-05-21
When meeting invitation emails are sent, 'MeetingInvitations' should be created


despo:
## Description of the issue 📄

Currently we track meeting invitations emailed out by [updating a value on the meeting model](https://github.com/codebar/planner/blob/master/app/controllers/admin/meetings_controller.rb#L54). There are a couple of issues with that, such as:
- Emails are all emailed out from the controller and don't utilise a delayed job
- This process often fails due to the amount of emails and thus the variable is not set
- It makes the process prone to errors and we've had issues with repeat emails multiple times!

## Steps to fix 🛠

In order to be able to both track and manage meeting invitations the same way we do for other events, 
- The meeting invitation sent process should be moved to the `InvitationManager` model and an invitation should be sent out **only if no MeetingInvitation associated with the Meeting and Member does not exist**, similar to WorkshopInvitations etc
- A `MeetingInvitation` entry should be created and be associated with each email
- The appropriate changes need to happen in the relevant Meeting views and controllers so that on RSVP the model is updated rather than created
- The MeetingInvitation mailed needs to be updated to accept a model
- Any other required changes to the `MeetingInvitation` model to support this functionality

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

gnclmorais:
I’m happy to take this, @despo! 🙋

despo:
@gnclmorais assigned!

despo:
Resolved by #1454

PR #1454, opened 2020-11-27
Create `MeetingInvitations` when meeting invitation emails are sent

gnclmorais:
Details on https://github.com/codebar/planner/issues/1317

I believe this is the core functionality, but please let me know if I missed something. `MeetingInvitation` are now created when the email is sent (through a delayed job) and retrieved by a token (if present).

**Update:** one test started failing with my last changes, I’ll address it this weekend. 💪🏻

### DIFF
--- a/app/controllers/admin/meetings_controller.rb
+++ b/app/controllers/admin/meetings_controller.rb
@@ -46,11 +46,6 @@ class Admin::MeetingsController < Admin::ApplicationController
   end
 
   def invite
-    if @meeting.invites_sent
-      return redirect_to admin_meeting_path(@meeting),
-                         notice: t('admin.messages.meeting.invitations_already_sent')
-    end
-
     InvitationManager.new.send_meeting_emails(@meeting)
 
     redirect_to admin_meeting_path(@meeting), notice: t('admin.messages.meeting.sending_invitations')
--- a/app/controllers/admin/meetings_controller.rb
+++ b/app/controllers/admin/meetings_controller.rb
@@ -52,7 +52,9 @@ class Admin::MeetingsController < Admin::ApplicationController
     end
 
     @meeting.invitees.not_banned.each do |invitee|
-      MeetingInvitationMailer.invite(@meeting, invitee).deliver_now
+      invitation = MeetingInvitation
+        .create(meeting: @meeting, member: invitee, role: 'Participant')
+      MeetingInvitationMailer.invite(@meeting, invitee, invitation).deliver_now
     end
     @meeting.update_attribute(:invites_sent, true)
 
--- a/app/controllers/admin/meetings_controller.rb
+++ b/app/controllers/admin/meetings_controller.rb
@@ -51,12 +51,7 @@ class Admin::MeetingsController < Admin::ApplicationController
                          notice: t('admin.messages.meeting.invitations_already_sent')
     end
 
-    @meeting.invitees.not_banned.each do |invitee|
-      invitation = MeetingInvitation
-        .create(meeting: @meeting, member: invitee, role: 'Participant')
-      MeetingInvitationMailer.invite(@meeting, invitee, invitation).deliver_now
-    end
-    @meeting.update_attribute(:invites_sent, true)
+    InvitationManager.new.send_meeting_emails(@meeting)
 
     redirect_to admin_meeting_path(@meeting), notice: t('admin.messages.meeting.sending_invitations')
   end
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -62,7 +62,7 @@ class InvitationsController < ApplicationController
     invitation = if params[:token].present?
       MeetingInvitation.find_by(token: params[:token], member: current_user)
     else
-      MeetingInvitation.build(
+      MeetingInvitation.new(
         meeting: meeting, member: current_user, role: 'Participant'
       )
     end
@@ -71,7 +71,8 @@ class InvitationsController < ApplicationController
 
     if invitation.save
       MeetingInvitationMailer.attending(meeting, current_user).deliver_now
-      redirect_to meeting_path(meeting), notice: t('messages.invitations.meeting.rsvp')
+      redirect_to meeting_path(meeting, token: invitation.token),
+                  notice: t('messages.invitations.meeting.rsvp')
     else
       redirect_to :back, notice: 'Sorry, something went wrong'
     end
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -59,7 +59,13 @@ class InvitationsController < ApplicationController
 
     meeting = Meeting.find_by(slug: params[:meeting_id])
 
-    invitation = MeetingInvitation.find_or_create_by(meeting: meeting, member: current_user, role: 'Participant')
+    invitation = if params[:token].present?
+      MeetingInvitation.find_by(token: params[:token])
+    else
+      MeetingInvitation.build(
+        meeting: meeting, member: current_user, role: 'Participant'
+      )
+    end
 
     invitation.update_attribute(:attending, true)
 
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -60,7 +60,7 @@ class InvitationsController < ApplicationController
     meeting = Meeting.find_by(slug: params[:meeting_id])
 
     invitation = if params[:token].present?
-      MeetingInvitation.find_by(token: params[:token])
+      MeetingInvitation.find_by(token: params[:token], member: current_user)
     else
       MeetingInvitation.build(
         meeting: meeting, member: current_user, role: 'Participant'
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -57,19 +57,10 @@ class InvitationsController < ApplicationController
   def rsvp_meeting
     return redirect_to :back, notice: 'Please login first' unless logged_in?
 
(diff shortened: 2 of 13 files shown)


app/models/invitation_manager.rb
```diff
@@ -26,6 +26,15 @@ def send_course_emails(course)
   end
   handle_asynchronously :send_course_emails
 
+  def send_meeting_emails(meeting)
+    meeting.invitees.not_banned.each do |invitee|
+      invitation = MeetingInvitation.new(meeting: meeting, member: invitee, role: 'Participant')
+      MeetingInvitationMailer.invite(meeting, invitee, invitation).deliver_now if invitation.save
+    end
+    meeting.update_attribute(:invites_sent, true)
```
despo, line 34: We should not be able to ignore the `invites_sent` attribute  as it was put in place to prevent organisers from sending out multiple meeting invitations to members, so can we remove this line please. 

There is a little bit more work to be done to remove it entirely but as long as we don't set it to true it should be okay for now.

app/views/meetings/_meeting_actions.html.haml
```diff
@@ -28,8 +28,8 @@
         = link_to "Can't make it anymore? Click here to cancel your spot.", meeting_cancel_path(meeting_id: @meeting.id, token: @invitation.token)
 
     - elsif @meeting.invitable && @meeting.not_full
-      = link_to meeting_invitation_path(@meeting), class: 'button round' do
-        = "Attend"
+      = link_to "Attend", meeting_invitation_path(@meeting, token: @invitation&.token),
+        class: "button round"
```
despo, line 32: nitpicking a bit, but can we please move this to the line above as this can earily be confused for nested content.

app/controllers/admin/meetings_controller.rb
```diff
@@ -51,10 +51,7 @@ def invite
                          notice: t('admin.messages.meeting.invitations_already_sent')
     end
 
-    @meeting.invitees.not_banned.each do |invitee|
-      MeetingInvitationMailer.invite(@meeting, invitee).deliver_now
-    end
-    @meeting.update_attribute(:invites_sent, true)
+    InvitationManager.new.send_meeting_emails(@meeting)
```
despo, line 54: 💯

despo (changes requested):
Thank you for the PR @gnclmorais - it looks really good!

I left some minor feedback and it would also be great if the codeclimate warnings are resolved.

app/views/meetings/_meeting_actions.html.haml
gnclmorais: Happy to, this is my Haml habit of breaking things up 😄

app/models/invitation_manager.rb
gnclmorais: Right, that makes sense. The reason I left it as it is was mostly because of this:
```rb
  # …
  # app/controllers/admin/meetings_controller.rb

  def invite
    if @meeting.invites_sent
      return redirect_to admin_meeting_path(@meeting),
                         notice: t('admin.messages.meeting.invitations_already_sent')
    end

    InvitationManager.new.send_meeting_emails(@meeting)

    redirect_to admin_meeting_path(@meeting), notice: t('admin.messages.meeting.sending_invitations')
  end

  # …
```

What would be the new behaviour once `invites_sent` is gone? Do we send duplicated invitations? Do we (add code to) skip people already invited? Do we don’t send anything since they have already been sent? And how do we identify that we have sent something, if needed? By the presence of at least one invite?

Sorry if there are too many questions to answer, but I would like to hear your take on this before I remove `invites_sent` — I’d be happy to. 😃

despo: #### TLDR; no extra code needs to be written as conditions to ensure that no duplicate MeetingInvitations are created and emailed out are already in place.

We don't send duplicate invitations. We [have a uniqueness validation]( https://github.com/codebar/planner/blob/master/app/models/meeting_invitation.rb#L8) on `MeetingInvitation` which will prevent the creation of more than one invitation for a member/meeting combination. This will cause line 32 on the code above to not save a new meeting invitation (nor send an email) and move on to the next meeting_invitation. It may not be optimal but it does the job.

I did briefly consider that we could maybe add a unique index of meeting_id and member_id on the meeting_invitation table, however, that will cause `.save` to throw an exception that we will have to handle (and I think more work than what is necessary to get this to work, given we use the same pattern for `WorkshopInvitation`).

app/models/meeting.rb
```diff
@@ -3,6 +3,8 @@ class Meeting < ActiveRecord::Base
   include Listable
   include Invitable
 
+  self.ignored_columns = %w(invites_sent)
```
despo, line 6: @gnclmorais this was introduced in rails 5.1.7 - we are not quite there yet :)

gnclmorais: Good catch! For a moment I was wondering if I got the syntax wrong. 😀 Removing then… It was just a safeguard if we could.

app/controllers/invitations_controller.rb
```diff
@@ -57,15 +57,13 @@ def reject
...
-    meeting = Meeting.find_by(slug: params[:meeting_id])
+    invitation = load_invitation
+    meeting = invitation.meeting
 
-    invitation = MeetingInvitation.find_or_create_by(meeting: meeting, member: current_user, role: 'Participant')
-
-    invitation.update_attribute(:attending, true)
-
-    if invitation.save
+    if invitation.update(attending: true)
```
despo, line 63:
```suggestion
    invitation.assign_attributes(attending: true)
    
    if invitation.save
```
As `update` returns the resulting object back whether that is updated or not, the `else` clause of this statement will never be reached

gnclmorais: > The resulting object is returned whether the object was saved successfully to the database or not. ([source](https://apidock.com/rails/ActiveRecord/Base/update/class))

You are right, I was certain it would return `true/false`… I’ll revert this particular change now. 👍 

P.S.: Thank you so much for the code reviews! ⚡

gnclmorais: Actually, @despo, are you sure? I was double-checking on my console and I seem to get a `true/false` instead:
```rb
# successful update
irb(main):009:0> Member.last.update(surname: "Morais")
D, [2020-12-04T09:37:14.202862 #96295] DEBUG -- :   Member Load (1.2ms)  SELECT  "members".* FROM "members"  ORDER BY "members"."id" DESC LIMIT 1
D, [2020-12-04T09:37:14.204774 #96295] DEBUG -- :    (0.4ms)  BEGIN
D, [2020-12-04T09:37:14.211939 #96295] DEBUG -- :   AuthService Load (0.8ms)  SELECT "auth_services".* FROM "auth_services" WHERE "auth_services"."member_id" = $1  [["member_id", 574]]
D, [2020-12-04T09:37:14.216121 #96295] DEBUG -- :   Member Exists (0.7ms)  SELECT  1 AS one FROM "members" WHERE ("members"."email" = 'gnclmorais@gmail.com' AND "members"."id" != 574) LIMIT 1
D, [2020-12-04T09:37:14.222830 #96295] DEBUG -- :   SQL (0.5ms)  UPDATE "members" SET "surname" = $1, "updated_at" = $2 WHERE "members"."id" = $3  [["surname", "Morais"], ["updated_at", "2020-12-04 09:37:14.217152"], ["id", 574]]
D, [2020-12-04T09:37:14.224083 #96295] DEBUG -- :   ActsAsTaggableOn::Tagging Load (0.4ms)  SELECT "taggings".* FROM "taggings" WHERE "taggings"."taggable_id" = $1 AND "taggings"."taggable_type" = $2  [["taggable_id", 574], ["taggable_type", "Member"]]
D, [2020-12-04T09:37:14.225752 #96295] DEBUG -- :    (0.9ms)  COMMIT
=> true

# unsuccessful update
irb(main):010:0> Member.last.update(email: nil)
D, [2020-12-04T09:38:09.206675 #96295] DEBUG -- :   Member Load (1.4ms)  SELECT  "members".* FROM "members"  ORDER BY "members"."id" DESC LIMIT 1
D, [2020-12-04T09:38:09.208022 #96295] DEBUG -- :    (0.3ms)  BEGIN
D, [2020-12-04T09:38:09.211060 #96295] DEBUG -- :   AuthService Load (0.5ms)  SELECT "auth_services".* FROM "auth_services" WHERE "auth_services"."member_id" = $1  [["member_id", 574]]
D, [2020-12-04T09:38:09.216430 #96295] DEBUG -- :   Member Exists (0.6ms)  SELECT  1 AS one FROM "members" WHERE ("members"."email" IS NULL AND "members"."id" != 574) LIMIT 1
D, [2020-12-04T09:38:09.217502 #96295] DEBUG -- :    (0.4ms)  ROLLBACK
=> false
```

despo: @gnclmorais haha, it seems like I'm a couple of years out of date on the documentation of update. It does indeed do that now, I was just used to working around it and google bringing back the old version result first didn't help!

gnclmorais:
Hey @despo, is there anything else I’m missing here or that you would like me to tackle differently? Let me know 😃

despo:
@gnclmorais if you can resolve this  https://github.com/codebar/planner/pull/1454#discussion_r532989367 it would be great. I won't be able to have a look on staging until tomorrow anyway so my plan is to merge this tomorrow evening!

despo (approved):
👍

app/views/meetings/_meeting_actions.html.haml
gnclmorais: OK, I just addressed this — sorry, I had forgotten about it 🙇

despo:
Thanks @gnclmorais 💯

PR #1455, opened 2020-12-02
Improve admin member view

despo:
Work to improve admin/member view and enable admins and organisers to make better decisions about events that need to take place, such as suspensions, attendance warnings or eligibility inquiries etc.

This is still work in progress and will likely change.

member-with-no-rsvps

member-with-upcoming-rsvps

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -270,3 +270,6 @@ RSpec/SubjectStub:
   Enabled: true
   Exclude:
     - 'spec/models/job_spec.rb'
+
+RSpec/NestedGroups:
+  Max: 5
--- a/app/assets/stylesheets/foundation_and_overrides.scss
+++ b/app/assets/stylesheets/foundation_and_overrides.scss
@@ -1395,6 +1395,10 @@ body.dashboard-faq dl dt {
   color: $warning-color;
 }
 
+.info {
+  color: $info-color;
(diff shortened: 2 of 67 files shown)


brunogirin:
This looks great! Will the "attendance summary" include the number of missed workshops over the past 6 months? It would be good for this screen to have some guidance too, such as:
- If someone is marked has a no show marker and no attendance warning, admins should send one
- If they have more no-shows after an attendance warning, it might time to consider a short suspension
- If they keep not showing up, consider a long suspension

This should always be at the discretion of the chapter admin so just some lightweight guidance would do.

When dealing with no-shows, is it important to identify whether the chapters they've no-showed at are administed by the admin? For example, if you have someone who attends Cambridge and Oxford and only has no-shows with Cambridge, would it be useful for both chapters to know that? Cambridge may want to take action but not Oxford.

KimberleyCook (approved):
From my glance, this looks good and gives organisers much more information which I really like.

despo:
https://github.com/codebar/planner/pull/1497 must be merged first to resolve spec failure.

I will look into integrating @brunogirin's feedback in the next couple of days. If I don't get around it, I plan to merge this by Monday and tackle the feedback on a separate PR given that this will be a major improvement even when merged as is.

PR #1458, opened 2020-12-02
Use GitHub actions

despo:


matyikriszta:
@despo if we are moving over to Github Action can we remove the travis config?

despo:
@matyikriszta I don't think we are quite ready to remove travis just yet. I actually expect this checks to fail because of the rubocop checks and want to see how it behaves when a new PR is opened.

matyikriszta (approved):
👍

PR #1464, opened 2020-12-04
Fix circular reference

despo:
ERROR: /home/runner/work/planner/planner/spec/support/omniauth.rb:4: warning: circular argument reference - uid


matyikriszta (approved):
👍

PR #1468, opened 2020-12-08
Resolve omniauth security warning (CVE 2015 9284)

despo:
The request phase of the OmniAuth Ruby gem is vulnerable to Cross-Site Request Forgery when used as part of the Ruby on Rails framework, allowing accounts to be connected without user intent, user interaction, or feedback to the user. This permits a secondary account to be able to sign into the web application as the primary account.

Details about how to work around this issue:
https://github.com/omniauth/omniauth/wiki/Resolving-CVE-2015-9284


matyikriszta (approved):
👍

PR #1472, opened 2021-01-11
Staging remove foundation replace with bootstrap

KimberleyCook:


matyikriszta:
@KimberleyCook do you have a list of pages that have been migrated over to Bootstrap? I think this would be easiest to test by clicking around the new pages in the browser.

KimberleyCook:
@matyikriszta there is a Trello ticket here https://trello.com/c/Oevjg5ae with a checklist of all the ones I've done so far

matyikriszta:
@KimberleyCook thanks, will review this in the next few days

matyikriszta:
@KimberleyCook I've started this spreadsheet with small layout issues I've noticed that I think would be good to fix before we go live with this.

despo:
Just an update on this. I spend sometime creating a new layout to isolate foundation and bootstrap but then realised that there is a mix use of components using either one or the other in pages, which means that that can't be done. :(

matyikriszta (approved):

PR #1473, opened 2021-01-12
Remove double negative in registration form

gavinhenderson:
The sentence next to the newsletter checkbox took me quite a while to understand due to its use of a double negative.

I have removed both negatives so that it says the same thing just using simpler language.

It would now read: 

> 'If you want to receive the codebar newsletter check this box'

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -684,7 +684,7 @@ en:
       member:
         email: 'Email address'
         name: 'First name'
-        newsletter: 'If you want to receive the codebar newsletter check this box'
+        newsletter: 'Sign up for our newsletter'
         pronouns: 'What pronouns do you use?'
     errors:
       models:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -684,7 +684,7 @@ en:
       member:
         email: 'Email address'
         name: 'First name'
-        newsletter: 'If you do not want to receive the codebar newsletter uncheck this box'
+        newsletter: 'If you want to receive the codebar newsletter check this box'
         pronouns: 'What pronouns do you use?'
     errors:
       models:


config/locales/en.yml
```diff
@@ -684,7 +684,7 @@ en:
       member:
         email: 'Email address'
         name: 'First name'
-        newsletter: 'If you do not want to receive the codebar newsletter uncheck this box'
+        newsletter: 'If you want to receive the codebar newsletter check this box'
```
despo, line 687: The issue with this is that the checkbox is already checked, so this text is not representative of the state of the checkbox.IIRC we decided that we wanted the checkbox to be checked by default, so changing that is not an option.

Definitely open to any other sugestions about how to reword this text.

despo (changes requested):
@gavinhenderson thanks for your PR. I left some feedback about why we can't change the text to the suggested one. 

However, we are definitely open to discussing this and listening to other suggestions about how this can be reworded.

config/locales/en.yml
gavinhenderson: Maybe the text shouldn't be a Call To Action in either form as it is right now? Is there an alternative that works no matter the default state?

Some suggestions:

> - [x] - Sign up for our newsletter

> - [x] - Receive the codebar newsletter

> Would you like to receive the codebar newsletter?
> - [x] - Yes
> - [ ]  - No 

Either way having a double negative like this seems designed to confuse people into getting onto the mailing list.

Disclaimer, not a UX expert

despo: Happy to go ahead with option one (as the third option requires to change the component to a radio button and handle that correctly at the controller), as it would read well and wouldn't require any additional changes to the component.

If you amend the local text to reflect that I'm happy to merge!

gavinhenderson:
@despo Changes made 👍

despo (approved):

despo:
@gavinhenderson thank you 💯

Issue #1474, opened 2021-01-20
Adding people from the waiting list manually can make them disappear


Drescoding:
## Description of the issue 📄

On an event, I tried to add someone manually (+), they confirmed that they got an email but their name doesn't appear anywhere (CSV, email list or the attendees). I then tried to add them from "Select a member to RSVP", it says it worked (see screenshots) but then they still don't appear.

It happened with Jean Francois (see screenshot) for the Oxford chapter and Toni Georgieva on Wednesday 20/01/2021 for the London chapter.

## Screenshots 📷
![image](https://user-images.githubusercontent.com/37814413/105472043-5ae77c00-5c93-11eb-8343-4e002c02a429.png)
![image](https://user-images.githubusercontent.com/37814413/105472056-5de26c80-5c93-11eb-8d54-4644b8cbd0cd.png)

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

gnclmorais:
I’ve address this on https://github.com/codebar/planner/pull/1475, thanks to @despo’s insight 🕵️🕵️‍♀️

PR #1475, opened 2021-01-23
Display error message when invitation update fails

gnclmorais:
Alex documented on #1474 the issue of adding someone manually from the waitlist, sometimes the member gets the correct email but they disappear from the list they were supposed to be added to.

Despo pointed me to this particular case there the invitation update could fail and we would have no way to know it. When this happens, their spot on the waiting list would be deleted.

This pull request addresses that issue now, paying attention to the result of the update, notifying the admin if it failed and only removing the WaitList entity if no error occurred.

Since every PR also gets some notes from Code Climate and CodeFactor, the diff grew because I had to deal with their comments. Let me know if you want to pair on reviewing this. 😃

### DIFF
--- a/app/controllers/admin/invitations_controller.rb
+++ b/app/controllers/admin/invitations_controller.rb
@@ -9,20 +9,26 @@ class Admin::InvitationsController < Admin::ApplicationController
 
     if params.key?(:attending)
       attending = params[:attending]
-
       if attending.eql?('true')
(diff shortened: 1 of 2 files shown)


matyikriszta (approved):
LGTM 👍

KimberleyCook:
@gnclmorais merged :D

PR #1476, opened 2021-01-27
Quick design update to make the sponsors section on event pages look better

KimberleyCook:
The sponsors section on the events pages currently looks a bit crap, I've put in a quick fix to make them look better.

- They will now stack next to each other rather than above
- The logo itself will link out rather than the bit of text next to it

This change has been done on my bigger Bootstrap PR but I'm not sure when this will be merged in, this is a quick fix.

| Before | After |
|-------|------|
|Screenshot 2021-01-27 at 11 37 09|Screenshot 2021-01-27 at 11 35 24|

### DIFF
--- a/app/views/events/_sponsors.html.haml
+++ b/app/views/events/_sponsors.html.haml
@@ -1,10 +1,3 @@
 - sponsors.each do |sponsor|
-  %li
-    .row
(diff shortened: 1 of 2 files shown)


matyikriszta (approved):
👍

PR #1480, opened 2021-02-11
Improve member checkin freeze

gnclmorais:
This pull request tries to help with the slowdown felt when admins try to check members in for a workshop. It does that in two ways:
- Tries to be a bit more specific with the HTML that is being replaced, targeting a smaller DOM surface area and only re-instantiating Foundation’s features that it needs (I think we only need the _tooltips_);
- When clicking on a check-in link, it gets replaced with a spinning icon until it gets replaced with the new HTML:
  codebar--loading

---

image
This freeze happens because a lot of JavaScript gets triggered with we call `$(document).foundation()`. In order to make this much faster than it is, I think we need to change the technique we follow to update this page.

### DIFF
--- a/app/views/admin/workshop/_attendances.html.haml
+++ b/app/views/admin/workshop/_attendances.html.haml
@@ -6,7 +6,9 @@
           %td.small-1.columns
             - if (invitation.parent.date_and_time - 30.minutes).past?
               - if invitation.attended.nil?
-                = link_to admin_workshop_invitation_path(@workshop, invitation, attended: true), class: 'verify_attendance', title: 'Verify attendance', method: :put, remote: true do
+                = link_to admin_workshop_invitation_path(@workshop, invitation, attended: true),
+                  class: 'verify_attendance', title: 'Verify attendance', method: :put, remote: true,
+                  data: { disable_with: "<i class='fa fa-spinner spin'></i>" } do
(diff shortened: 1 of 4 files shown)


app/views/admin/workshop/_attendances.html.haml
```diff
@@ -6,7 +6,9 @@
           %td.small-1.columns
             - if (invitation.parent.date_and_time - 30.minutes).past?
               - if invitation.attended.nil?
-                = link_to admin_workshop_invitation_path(@workshop, invitation, attended: true), class: 'verify_attendance', title: 'Verify attendance', method: :put, remote: true do
+                = link_to admin_workshop_invitation_path(@workshop, invitation, attended: true),
+                  class: 'verify_attendance', title: 'Verify attendance', method: :put, remote: true,
+                  data: { disable_with: "<i class='fa fa-spinner spin'></i>" } do
```
gnclmorais, line 11: This `data-disable-with` trick is a nice things a lot of Rails app have, [rails-ujs (Unobtrusive JavaScript)](https://m.patrikonrails.com/a-definitive-guide-to-railss-unobtrusive-javascript-adapter-ef13bd047fff#dfbe).

matyikriszta (approved):
This looks good, thanks @gnclmorais

PR #1485, opened 2021-02-21
Don't allow rsvp-ing a workshop that is in the past #1484

### LINKED ISSUE #1484: I should not be able to register for a past event
## Description of the issue 📄

As a coach I was browsing events for today during lunch break. I found one, opened it in a new tab, and went about my business. It was in a different timezone, which I didn't realise. Half an hour before I thought the event started, I came back to my open tab, and clicked the button to attend as coach. This worked, and I ended up on the invitation page, which correctly said it had already taken place. I did not notice this at the time. So I went on Discord, noticed people were paired up, I wasn't needed, facepalmed and felt embarrassed for not being able to remember that Oslo and London are in different timezones.

I suppose my attendance record in the database shows that I signed up after the event took place. I'm happy to share my ticket ID (or the link to it) privately to make finding it easier.

## Screenshots

Screenshot showing today's date and yesterday's workshop with the button still there

I think there should be validation when signing up for an event on whether this event is in the past. I believe that right now, I could retroactively add me to any event with just GETting the URL, and as long as there are free slots, that would work. This would inflate my participation record greatly, making me more revered but also nobody would ever remember me.

## Steps to fix 🛠

I believe this would have to be done in https://github.com/codebar/planner/blob/master/app/controllers/events_controller.rb#L49 or https://github.com/codebar/planner/blob/master/app/controllers/concerns/invitation_controller_concerns.rb#L10, but I don't know enough Ruby to suggest a fix. The template already does the right thing if it is rendered once an event has taken place, but the backend does not.

## To do 📋

* [ ] Submit a pull request

simbabque:
The backend did not do validation on the workshop's time and date when
a user sent an rsvp. There is validation in the view when the workshop
page is rendered, so the button to attend does not show up on past
workshops. However if the page was old and the button was there, a user
could still sign up and get an invitation created. The workshop was then
displayed as in the past.

For virtual workshops it is possible to attend anyway, so a confirmation
email would be sent. This change stops the system from ever creating
invitations for past workshops.

I think the new test file should also contain tests for the other validation criteria in the `workshops_controller#rsvp action`. I decided not to include them here to keep the PR to the scope of the issue.

Closes #1484

### DIFF
--- a/app/controllers/workshops_controller.rb
+++ b/app/controllers/workshops_controller.rb
@@ -9,7 +9,7 @@ class WorkshopsController < ApplicationController
   end
 
   def rsvp
-    redirect_to :back, notice: t('workshops.registration_not_open') unless @workshop.invitable_yet?
+    return redirect_to :back, notice: t('workshops.registration_not_open') unless @workshop.available_for_rsvp?
 
     if role_params.nil?
       @invitation = find_attending_invitation(@workshop, current_user)
--- a/spec/support/shared_examples/behaves_like_managing_workshop_attendance.rb
+++ b/spec/support/shared_examples/behaves_like_managing_workshop_attendance.rb
@@ -163,8 +163,8 @@ RSpec.shared_examples 'managing workshop attendance' do
             visit workshop_path(workshop)
 
             expect(page).to have_button('Manage your invitation')
-            expect(page).not_to have_link('Attend as a student')
-            expect(page).not_to have_link('Attend as a coach')
+            expect(page).not_to have_button('Attend as a student')
+            expect(page).not_to have_button('Attend as a coach')
           end
         end
 
--- a/spec/support/shared_examples/behaves_like_managing_workshop_attendance.rb
+++ b/spec/support/shared_examples/behaves_like_managing_workshop_attendance.rb
@@ -167,6 +167,17 @@ RSpec.shared_examples 'managing workshop attendance' do
             expect(page).not_to have_link('Attend as a coach')
           end
         end
+
+        context 'when the workshop has already happened' do
+          it 'can not rsvp to workshop that is now in the past' do
+            travel_into_the_future = Time.zone.now + 3.days
+            allow(Time).to receive(:now).and_return(travel_into_the_future)
+      
+            click_on 'Attend as a coach'
+            expect(page).to have_content('This event has already taken place')
+            expect(page).not_to have_button('Attend as a coach')
+          end    
+        end
       end
     end
 
--- a/app/models/workshop.rb
+++ b/app/models/workshop.rb
@@ -72,6 +72,10 @@ class Workshop < ActiveRecord::Base
     open_for_rsvp? || invitable
   end
 
+  def available_for_rsvp?
+    invitable_yet? && date_and_time.future?
+  end
+
   # Is this person attending this event?
   def attendee?(person)
     return false if person.nil?


simbabque:
I'm not sure how to address the two issues that CodeFactor and codeclimate have found. This is my first Ruby code, and I'm not familiar with the tools.

I believe I could have make the `rsvp` method shorter by switching existing code to post-fix conditions, but that feels like it would reduce readability, thus defeating what I think the point of these tools is. I don't know what to do about too many expectations. I've always struggled when I was told my expectations are too high. :wink:

app/controllers/workshops_controller.rb
```diff
@@ -11,6 +11,10 @@ def show
   def rsvp
     redirect_to :back, notice: t('workshops.registration_not_open') unless @workshop.invitable_yet?
 
+    unless @workshop.date_and_time.future?
+      return redirect_to :back, notice: t('messages.already_taken_place')
```
despo, line 15:
```suggestion
      return redirect_to :back, notice: t('messages.already_taken_place')  unless @workshop.date_and_time.future?
```

despo: @simbabque Maybe a better way around this that is also easier to read (and should resolve the codeclimate issue) is to extend the workshop model to perform the check about the workshop being available for rsvp, and calling that through here. I think adding a more specific message is a nice to have, but not essential as I _think_ the refreshed page might already have some text indicating that the event has already taken place.


e.g.

```ruby
def available_for_rsvp?
   invitable_yet? && date_and_time.future?
end

File: spec/features/rsvping_workshop.rb [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+      travel_into_the_future = Time.zone.now + 3.days
+      allow(Time).to receive(:now).and_return(travel_into_the_future)
+
+      click_on 'Attend as a coach'
+      expect(page).to have_current_path(workshop_path(future_workshop), ignore_query: true)
+      expect(page).to have_content('This event has already taken place')
+      expect(page).not_to have_button('Attend as a coach')
+    end
+  end
+end
```
despo, line 37: We already have a spec testing this functionality https://github.com/codebar/planner/blob/master/spec/features/managing_workshop_attendance_spec.rb so I we only need the last scenario. It should probably be added to the [shared examples](https://github.com/codebar/planner/blob/master/spec/support/shared_examples/behaves_like_managing_workshop_attendance.rb) file.

despo (changes requested):
Thank you for your PR. I left some feedback about how the codefactor and codeclimate issues can be addressed and some additional comments about the specs.

simbabque:
Thank you @despo. I've moved the check into the model and moved the test into existing shared examples file. While there I noticed that another context that checks for the attendance buttons with `have_link` could never pass, so I included a fix for that.

spec/support/shared_examples/behaves_like_managing_workshop_attendance.rb
```diff
@@ -163,10 +163,21 @@
             visit workshop_path(workshop)
 
             expect(page).to have_button('Manage your invitation')
-            expect(page).not_to have_link('Attend as a student')
-            expect(page).not_to have_link('Attend as a coach')
+            expect(page).not_to have_button('Attend as a student')
+            expect(page).not_to have_button('Attend as a coach')
```
despo, line 167: Good catch! Thank you.

despo (approved):
@simbabque this looks great. Thank you so much! 💯

PR #1498, opened 2021-04-01
Remove skills controller

despo:
- Coach data should only be available through /coaches, for the specified year and depending on attendances.


matyikriszta (approved):
👍

Issue #1490, opened 2021-03-01
Add a Flag Icon before someone's name if a note has been added to their profile recently


KimberleyCook:
## Description of the issue 📄
"Notes" can be added to someone's profile by an organiser, but it isn't 100% clear and obvious when one has been added. To let other organisers know one was added recently, we would like to add a little flag icon before the person's name in the admin area. (show icon for the last 5 workshops RSVP'd for). 

Further discussion can be found here from the Chapter Org Meeting https://github.com/codebar/chapter-organiser-meetings/issues/53

## Screenshots 📷

![Screenshot 2021-03-01 at 14 17 23](https://user-images.githubusercontent.com/2683270/109502475-f2a96880-7a90-11eb-861e-c6a4fcb62e94.png)

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

PR #1499, opened 2021-04-05
Display latest notes for a member

gnclmorais:
When looking through the admin list of students and coaches, this commit will display a speech balloon icon if the user has new notes from the past 5 workshops. When hovered, you’ll see a tooltip with the content of these notes.

Fixes https://github.com/codebar/planner/issues/1490

codebar--hover-with-notes


app/assets/stylesheets/partials/_member_note.scss
```diff
@@ -1,11 +1,8 @@
...
-  padding-bottom: 1em;
-
-  border-bottom: 1px solid #ddd;
-
-  &:last-child {
-    border-bottom: none;
+  & + & {
+    margin-top: 1em;
+    padding-top: 1em;
+    border-top: 1px solid #ddd;
```
gnclmorais, line 5: This CSS change allows us to show stacked notes as previously but it will fix a small issue with putting the notes HTML inside the tooltip, where  the tooltip notch (that triangle) was the last element so we were getting an extra border without any need. This makes it a bit more specific (i.e. only a note followed by another note will have this applied).

app/controllers/concerns/admin/workshop_concerns.rb
```diff
@@ -7,14 +7,20 @@ module Admin::WorkshopConcerns
...
       @coach_waiting_list = WaitingListPresenter.new(WaitingList.by_workshop(@workshop)
                                                 .where_role('Coach')
-                                                .order(:created_at))
+                                                .order(:created_at)
+                                                .includes(member: { member_notes: :author }))
       @student_waiting_list = WaitingListPresenter.new(WaitingList.by_workshop(@workshop)
                                                   .where_role('Student')
-                                                  .order(:created_at))
+                                                  .order(:created_at)
+                                                  .includes(member: { member_notes: :author }))
```
gnclmorais, line 23: These changes address a few N+1 queries Bullet warned me about.

app/controllers/concerns/admin/workshop_concerns.rb
```diff
@@ -7,14 +7,19 @@ module Admin::WorkshopConcerns
...
+      @attending_coaches = InvitationPresenter.decorate_collection(
+        @workshop.attending_coaches.all.with_notes_and_their_authors
+      )
+
+      @coach_waiting_list = WaitingListPresenter.new(
+        WaitingList.by_workshop(@workshop).where_role('Coach').order(:created_at).with_notes_and_their_authors
+      )
+      @student_waiting_list = WaitingListPresenter.new(
+        WaitingList.by_workshop(@workshop).where_role('Student').order(:created_at).with_notes_and_their_authors
+      )
```
gnclmorais, line 22: The additional `with_notes_and_their_authors` scope preloads notes and their authors to address some N+1 queries Bullet was complaining about.

gnclmorais:
About the issues found by CodeFactor and CodeClimate, I must say the rules that are being flagged are quite restrictive and I don’t believe they provide that much value. Could we at bump the limits a bit, perhaps?

KimberleyCook (approved):
Looks good to me, but worth having another pair of eyes too.

despo:
Because of the changes that have been merged in admin/member I think some of these changes are not applicable anymore. Can you rebase and review?

gnclmorais:
@despo, rebased 👌

app/assets/stylesheets/partials/_member_note.scss
despo: Does this apply to the new view?

app/views/admin/workshop/_attendances.html.haml
```diff
@@ -23,6 +23,11 @@
               %span.has-tip{ 'data-tooltip': '', 'aria-haspopup': 'true', title: 'Multiple no shows and attendance warnings in the last six months' }
                 %i.fa.fa-exclamation-circle
 
+            - recent_notes = invitation.member.recent_notes
```
despo, line 26: Can we maybe move this to the member presenter. We should avoid setting up variables in views if possible as they are easy to miss and add unecessary noise..

app/assets/stylesheets/partials/_member_note.scss
gnclmorais: You are right; since the view changed, this is probably unnecessary now.

app/assets/stylesheets/main.scss
```diff
@@ -7,7 +7,6 @@
 @import 'partials/events';
 @import 'partials/social_media';
 @import 'partials/chapter';
-@import 'partials/member_note';
```
gnclmorais, line 10: https://github.com/codebar/planner/pull/1455 changed an admin’s view of a user. Since then, I believe this CSS hasn’t been used (as well as the `app/assets/stylesheets/partials/_member_note.scss` partial below).

gnclmorais:
Hey @despo and @matyikriszta, whenever you have a minute to review this, I would appreciate… 🙇‍♂️ It would address https://github.com/codebar/planner/issues/1490, a suggestion from one of our organiser meetings.

Regarding the issues found by CodeFactor and CodeClimate, I must say the rules that are being flagged are quite restrictive and I don’t believe they provide that much value. Could we at bump the limits a bit, perhaps?

Thank you!

matyikriszta (commented):
At a first glance this looks good to me.

app/controllers/concerns/admin/workshop_concerns.rb
```diff
@@ -7,14 +7,19 @@ module Admin::WorkshopConcerns
...
-                                                  .order(:created_at))
+      @attending_students = InvitationPresenter.decorate_collection(
+        @workshop.attending_students.all.with_notes_and_their_authors
+      )
+      @attending_coaches = InvitationPresenter.decorate_collection(
+        @workshop.attending_coaches.all.with_notes_and_their_authors
+      )
+
+      @coach_waiting_list = WaitingListPresenter.new(
+        WaitingList.by_workshop(@workshop).where_role('Coach').order(:created_at).with_notes_and_their_authors
```
matyikriszta, line 18: @gnclmorais not blocking but I'm wondering if it'd be worth moving the code block that retrieves waitlisted coaches and students (on line 21) into their own methods on member, similarly to `attending_coaches` and `attending_students`. I think that would be make it a bit easier to digest this method.

gnclmorais: I agree, I can move part of it to a new method. Good suggestion, thanks!

matyikriszta (approved):
Looks great 👍

gnclmorais:
image
Seems like I can’t merged this, so… if any reviewer that can do it would please go ahead and merge it, I would be thankful! 🙇‍♂️

PR #1500, opened 2021-04-05
Cache Ruby gems on GitHub Actions

gnclmorais:
While working on , I’ve noticed that GitHub Actions were taking a surprising amount of time:
image

Looking into the most recent one, the overall time spent on dependencies alone was surprising (notice the times on the right):
image

This seems like missing cache on GitHub Actions, so I [researched a bit](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically) and I was able to largely improve this time (notice the times on the right):
image

I moved the config options we had on the `Install dependencies` step into `.bundle/config` (example below) but also to the GitHub Action:
```diff
---
BUNDLE_WITHOUT: "production"
+BUNDLE_JOBS: 4
+BUNDLE_RETRY: 3
```

#### Example

[Here’s the first run](https://github.com/codebar/planner/actions/runs/719512688) of this pull request and [here’s the second one](https://github.com/codebar/planner/actions/runs/719528470).

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -26,6 +26,9 @@ jobs:
       with:
         ruby-version: 2.4.2
         bundler-cache: true
+      env:
+        BUNDLE_JOBS: 4
+        BUNDLE_RETRY: 3
 
     - name: Setup test database
       env:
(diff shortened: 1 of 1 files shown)


KimberleyCook (approved):
🎉 💃 🎉 💃

despo:
Thanks @gnclmorais :shipit:

PR #1503, opened 2021-04-15
Update calls to deleted methods in _admin_menu partial & in coach/feedback/index page

matyikriszta:
## Description
This PR fixes a small bug in the admin left hand side menu. Some Member method names got update is #1455 which meant that the New Workshop and Add Sponsor options in the menu were permanently hidden, as the `is_admin_or_organiser?` method always evaluated to false.

## Status
Ready for Review

## Related Github issue
Bug introduced in #1455

## Screenshots
Screenshot 2021-04-15 at 9 02 10 am

### DIFF
--- a/app/views/layouts/_admin_menu.html.haml
+++ b/app/views/layouts/_admin_menu.html.haml
@@ -3,14 +3,14 @@
 %li= link_to 'Admin Dashboard', admin_root_path
 %li= link_to 'View Feedback', admin_feedback_index_path
 %li= link_to 'New Announcement', new_admin_announcement_path
-- if current_user.is_admin_or_organiser?
-  %li= link_to 'New workshop', new_admin_workshop_path
-  %li= link_to 'New sponsor', new_admin_sponsor_path
+%li= link_to 'New workshop', new_admin_workshop_path
+%li= link_to 'New sponsor', new_admin_sponsor_path
 %li= link_to 'New event', new_admin_event_path
 %li= link_to 'New monthly', new_admin_meeting_path
 %li= link_to 'New chapter', new_admin_chapter_path
 %li= link_to 'New group', new_admin_group_path
 %li= link_to "Review jobs (#{jobs_pending_approval})", admin_jobs_path
+
 - if current_user.has_role?(:organiser, :any)
   %li
     %label My Chapters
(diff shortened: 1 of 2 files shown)


despo (approved):
Good catch. Thanks for ficing.

app/views/layouts/_admin_menu.html.haml
```diff
@@ -3,15 +3,15 @@
 %li= link_to 'Admin Dashboard', admin_root_path
 %li= link_to 'View Feedback', admin_feedback_index_path
 %li= link_to 'New Announcement', new_admin_announcement_path
-- if current_user.is_admin_or_organiser?
+- if current_user.admin_or_organiser?
   %li= link_to 'New workshop', new_admin_workshop_path
   %li= link_to 'New sponsor', new_admin_sponsor_path
```
matyikriszta, line 8: @despo I'm wondering what is the point of this check here. I don't see it being necessary, can we remove it? Once you are an admin you should be able to do these actions and we are checking whether the user is an admin when we are including the partial in `app/views/layouts/_aside.html.haml`.

KimberleyCook (approved):

app/views/layouts/_admin_menu.html.haml
despo: Interesting. I wonder if it's a copy paste from the initial partial that was shared between admins and organisers. Yes, let's remove.

app/views/layouts/_admin_menu.html.haml
```diff
@@ -3,15 +3,15 @@
...
 %li= link_to 'New event', new_admin_event_path
 %li= link_to 'New monthly', new_admin_meeting_path
 %li= link_to 'New chapter', new_admin_chapter_path
 %li= link_to 'New group', new_admin_group_path
 %li= link_to "Review jobs (#{jobs_pending_approval})", admin_jobs_path
-- if current_user.has_role?(:organiser, :any)
+- if current_user.organised_chapters.any?
```
despo, line 14: Can we revert this? `organised_chapter` will retrieve all chapters that someone is an organiser on and apply `any?` whilst checking for the role will just check for a match in the permissions table, which is faster.

matyikriszta: @despo sure, makes sense

PR #1531, opened 2021-05-10
Update es.yml

103percent:
Some more string translations


KimberleyCook (approved):
Thank you so much for these translations!!

103percent:
Motivated by the fact that we can get all our Spanish speaking friends in on the codebar fun!

PR #1540, opened 2021-05-14
Update es.yml

103percent:
Some more strings


KimberleyCook (approved):
🎉

config/locales/es.yml
```diff
@@ -80,15 +80,15 @@ es:
...
-    not_attending_already: "You have already confirmed you can't make it!"
-    no_available_seats: "Unfortunately there are no more spaces left :(. If you would like to join our waiting list send us an email at %{email}"
-    feedback_saved: "Thank you for taking the time to give us feedback!"
-    feedback_not_found: "You have already submitted feedback for this event."
+    rejected_invitation: "Estamos tan tristes que no puedes asistir, pero gracias por hacérnoslo saber, %{name}."
+    already_rsvped: "¡Ya has confirmado tu asistencia!"
+    not_attending_already: "¡Ya has confirmado que no puedes asistir!"
+    no_available_seats: "Desafortunadamente no quedan más espacios :(. Si desea unirse a nuestra lista de espera envíenos un correo electrónico a %{email}"
+    feedback_saved: "¡Gracias por tomarse el tiempo para darnos su opinión!"
+    feedback_not_found: "Ya ha enviado comentarios para este evento".
```
despo, line 91: There seems to be an issue with the double quote character used. I think this should resolve it.

```suggestion
    feedback_not_found: "Ya ha enviado comentarios para este evento".
```

matyikriszta: @despo I think the issue is that the period is outside the double quotes, so the correct string would be:

```suggestion
    feedback_not_found: "Ya ha enviado comentarios para este evento."
```

PR #1552, opened 2021-05-26
Update es.yml

103percent:
Some more strings!


KimberleyCook (approved):

KimberleyCook:
I've approved but can't merge because some tests are failing 🤔

103percent:
> I've approved but can't merge because some tests are failing 🤔

Hmm. I'm a little lost as to how I can help this one along. I'm a bit lost with the interface here :'(

matyikriszta:
@103percent I think the issue was with one of the services we use (coveralls) versus your code. I've re-run the jobs and they all passed. There is now however a merge conflict, could you rebase this PR so we can merge?

PR #1559, opened 2021-06-08
Replace 'minority groups' language

timwis:
Hey all 👋🏻 no pressure to merge this PR, as I know codebar is very deliberate about language. But I just wanted to offer this alternative, since "minority group members" wouldn't cover women.

Note that we'd want to replicate this in `en_US` and `en_AU` etc. But I thought I'd just change this one in case you want to tweak the language first. Happy to change the others if helpful.

### DIFF
--- a/config/locales/en_GB.yml
+++ b/config/locales/en_GB.yml
@@ -48,7 +48,7 @@ en-GB:
         format: '%u%n'
   homepage:
     intro: "<strong>codebar</strong> is a charity that facilitates the growth of a diverse tech community by running free regular programming workshops for minority groups in tech."
-    explanation: "Our goal is to enable minority group members to learn programming in a safe and collaborative environment and expand their career opportunities. To achieve this we run free regular workshops, regular one-off events and try to create opportunities for our students making technology and coding more accessible."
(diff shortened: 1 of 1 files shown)


matyikriszta:
Hey @timwis, we are always happy to consider improvements to our language. I am however wondering why you think the 'minority group members' phrase does not cover women? Women are part of minority groups are they not? I am happy to change but currently I don't see how your suggestion is an improvement to the current language.

timwis:
Absolutely! I'm no expert on the subject, but my understanding is that because women as a whole make up 50% of the population, the term 'minority' doesn't fit (except when combined with other characteristics/intersectionality). Rather, they are _underrepresented_ in the technology industry.

Just to be clear, my intention isn't to split hairs, but to avoid unintentionally excluding anyone. But please feel free to get more eyes on this if you're not sure!

matyikriszta:
Hey @timwis, sorry for not getting back to you sooner and thanks for the explanation, I now see what you meant more clearly. On first look I wasn't sure how this improves the meaning of the text but now I agree that your wording is perhaps better (also English is not my first language so nuances in meaning sometimes get lost on me). @KimberleyCook what do you think?

PR #1562, opened 2021-06-14
Update es.yml

103percent:
More Spanish Strings


matyikriszta (approved):
Looks good @103percent

PR #1575, opened 2021-07-13
Add development environment to Skylight

gnclmorais:
I’m looking into using Skylight to perf some local changes before deploying them, and I believe this is how you do it, according to their documentation: https://www.skylight.io/support/environments
I’m happy to revert this change if it doesn’t work.


despo (approved):
👍🏼

PR #1672, opened 2021-11-08
Fix student feedback form

### LINKED ISSUE #1665: Not possible to add a coach to feedback, but coach is required
## Description of the issue 📄
When a student is attempting to give feedback about a coaches performance the coaches look up is not working.
<s>The coaches name is a required field
If a student can not remember the coaches name, they would be unable to leave feedback.</s>

## Screenshots 📷
The student in this video is attempting to find a coach called "liska"
https://user-images.githubusercontent.com/4013283/140178102-b0377f58-56d8-4d14-b7b3-fcabd6230151.mp4

## Steps to fix 🛠
I don't know

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Submit a pull request

gnclmorais:
This pull request fixes #1665, but I must write _what exactly_ does it fix.

@khamiltonuk reported that a student could not submit feedback about a coach, with the culprit pointing out to the coach field being required. Upon closer inspection, this field turned out to be optional, so there must be something else preventing us from submitting…

Going to the corresponding page and trying to submit something totally empty did not work, as expected… but there was no errors, no browser messages highlighting that field X or Y had to be filled in. Looking at it a bit closer took me to https://github.com/harvesthq/chosen/issues/515, the `<select>`s we use on that page are being replaced with Chosen, and while doing that the original _required_ fields are being hidden. When they are hidden, they cannot show the required tooltips browsers do, so the first fix this PR has it to make them “visible” (i.e. hidden behind Chosen but you can see the required tooltips).

After taking care of this, the submit was still not working. Peeking under the hood, I noticed the start rating script we have gets setup (displaying seemingly interactive stars) but the actual selected star was not being converted into a number and set on the hidden input, thus the data reaching the controller was empty — in other words, invalid. Tweaking a bit the JavaScript we had fixed this, we were now submitting feedback! Huzzah!

But… surely we had a test for this. That test was passing, but the feature was kind of broken. Why is that? Well, the test we had was, in summary, testing the page _without_ JavaScript, which would offer the user a very different experience: no stars to select, no search-included “select” powered by Chosen. I had to adapt the test to be ran in a JavaScript supported environment and introduce some lines of code that would allow us to pick items from a Chosen-powered “select”.

Having done all of this, I finally ran into a small wall. Capybara was erring because a sponsor image was not found. I must admit I gave up on properly fixing this ‘cause it looked like magic to me. The image _was_ there (i.e. the file system) but I could not access it on the browser — even with a direct link. After a lot of digging, I could only make this go away basically by telling Capybara not to break our tests if images are missing. If someone has a better fix for this, I’d love to know about it! 🙇‍♂️ 

Update: on the last commit I tried to bump Ruby’s version to see if GitHub Actions stopped complaining about not being able to setup a database. From other runs on other PRs, that seems to be a common issue.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -148,7 +148,7 @@ GEM
     faraday-net_http_persistent (1.2.0)
     faraday-patron (1.0.0)
     faraday-rack (1.0.0)
-    ffi (1.12.2)
+    ffi (1.15.4)
     font-awesome-rails (4.7.0.7)
       railties (>= 3.2, < 7)
     foreman (0.87.2)
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -45,6 +45,23 @@ $(function(){
 
   $('body').removeClass('no-js');
 
+  // Chosen hides inputs and selects, which becomes problematic when they are
+  // required: browser validation doesn't get shown to the user.
+  // This fix places "the original input behind the Chosen input, matching the
+  // height and width so that the warning appears in the correct position."
+  // https://github.com/harvesthq/chosen/issues/515#issuecomment-474588057
+  $('select').on('chosen:ready', function () {
+    var height = $(this).next('.chosen-container').height();
+    var width = $(this).next('.chosen-container').width();
+
+    $(this).css({
+      position: 'absolute',
+      height: height,
+      width: width,
+      opacity: 0
+    }).show();
+  });
+
   $('select').chosen(function(){
     allow_single_deselect: true
     no_results_text: 'No results matched'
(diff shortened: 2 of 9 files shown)


gnclmorais:
@despo or @matyikriszta, could one of you review this please? Maybe @KimberleyCook even? This is preventing our students from submitting feedback about coaches…

KimberleyCook:
I don't 100% know what is going on (because Ruby) but trust you fully that it fixes the issue.

KimberleyCook (approved):

PR #1675, opened 2021-11-24
Bootstrap edit sponsor breadcrumb

KimberleyCook:
| Before | After |
|-------|------
|Screenshot 2021-11-24 at 16 28 23|Screenshot 2021-11-24 at 16 28 09|

### DIFF
--- a/app/views/admin/sponsors/edit.html.haml
+++ b/app/views/admin/sponsors/edit.html.haml
@@ -1,12 +1,13 @@
-%section#admin#banner
+.containuer-fluid
   .row
-    .large-12.columns
-      %ul.breadcrumbs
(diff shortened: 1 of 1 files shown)


app/views/admin/sponsors/edit.html.haml
```diff
@@ -1,12 +1,13 @@
-%section#admin#banner
+.containuer-fluid#admin#banner
```
matyikriszta, line 1: @KimberleyCook I think we could remove the `#admin` and `#banner` ids from here and instead apply spacing via a Bootstrap class. I've used `.pt-3` on other pages.

matyikriszta (approved):

PR #1696, opened 2022-01-03
Monthlies emails copy tweaks

KimberleyCook:


matyikriszta (approved):
This look good, to fix the failing tests update the copy in the test files as well (the tests are still checking for the old copy).

PR #1704, opened 2022-01-19
Bootstrap top half of event admin page

KimberleyCook:
| Before | After |
|-------|------|
|screencapture-localhost-3001-admin-events-corrupti-1-2022-01-19-14_25_29|screencapture-localhost-3001-admin-events-corrupti-1-2022-01-19-15_08_34|

### DIFF
--- a/app/views/admin/events/_invitation_management.html.haml
+++ b/app/views/admin/events/_invitation_management.html.haml
@@ -1,32 +1,34 @@
 - if @event.invitations.any?
   .row
-    .large-12.columns
-      = simple_form_for :invitation, url: admin_invitation_path, method: :put do |f|
-        .medium-6.columns
-          = f.select :id,
-                     @event.invitations.not_accepted.all.map { |u| [ "#{u.member.full_name} (#{u.role})", u.token] },
-                     { include_blank: true },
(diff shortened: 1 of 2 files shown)


app/views/admin/events/_invitation_management.html.haml
```diff
@@ -1,32 +1,34 @@
...
+    .col
+      .card.border-info
+        .card-body
+          %p.mb-0
+            <strong>#{@event.invitations.count}</strong> members have been invited to this event.
+            %br
+            <strong>#{@attending_students.count}</strong> are attending as students and <strong>#{@attending_coaches.count}</strong> as coaches.
+
+
+  %div.mt-4
```
matyikriszta, line 12: @KimberleyCook not blocking but in HAML you do not need to specify the %div tag. If you do a class or id (`.some-class` or `#some-id`) it will automatically create a div (basically div is the default tag in HAML).

matyikriszta (approved):

app/views/admin/events/_invitation_management.html.haml
KimberleyCook: Eugh, I do this often and always forget. Damn you HAML!

PR #1705, opened 2022-01-19
Update alerts

KimberleyCook:
| Before | After |
|-------|------|
|screencapture-localhost-3001-admin-events-corrupti-1-2022-01-19-15_17_17|screencapture-localhost-3001-admin-events-corrupti-1-2022-01-19-15_25_27|


matyikriszta (approved):
@KimberleyCook have you confirmed the alerts are still dismissable with the close button? I think I had an issue with them when I tried to convert them before.

Otherwise looks good 👍🏻

PR #1717, opened 2022-03-10
Delete unused skills show view

matyikriszta:
### Description
The SkillsController and associated files were deleted in https://github.com/codebar/planner/pull/1498 but somehow we missed this file. Since this route no longer exists it's safe to delete it.

### Status
Ready for Review


gnclmorais (approved):
Good spot! 😍

PR #1718, opened 2022-03-14
Remove mentions of food given remote setting

gnclmorais:
Small change, but I actually had a coach reaching out to me about this email oddity. Given the current remote setting, might as well remove this… And whenever we move to IRL gatherings, I don’t think this mention will be missed.


matyikriszta (approved):
@gnclmorais makes sense, thanks for implementing the change 👍🏻

PR #1725, opened 2022-04-04
Bootstrap Workshop Page and Attendee List

KimberleyCook:
Some after pictures (this PR **DOES NOT** include the waitlist component)

Screenshot 2022-04-04 at 17 10 32
Screenshot 2022-04-04 at 17 09 38
Screenshot 2022-04-04 at 17 07 36

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -71,4 +71,7 @@ $(function(){
   $('#member_lookup_id').chosen().change(function(e) {
     $('#view_profile').attr('href', '/admin/members/' + $(this).val())
   });
+
+  $('[data-bs-toggle="tooltip"]').tooltip();
+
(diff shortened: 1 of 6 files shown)


matyikriszta (commented):
@KimberleyCook this looks good. The test are failing because they are looking for some markup that you have now changed in the view. Once the tests are fixed it's good to go.

matyikriszta:
@KimberleyCook I don't know how to mark an item as Won't fix in CodeFactor or who set it up and why (and most importantly where it's configured) so for now feel free to just delete the offending line out from the test file on line 141 -> spec/support/shared_examples/behaves_like_an_invitation_route.rb:141, it actually is an unused variable and it should go anyways.

matyikriszta (approved):

PR #1728, opened 2022-04-06
Bootstrap Event Attendance List

KimberleyCook:
| Before | After |
|-------|------|
|screencapture-localhost-3001-admin-events-corrupti-1-2022-04-06-12_27_00|Screenshot 2022-04-06 at 12 35 12|


app/views/admin/events/_attendances.html.haml
```diff
@@ -1,18 +1,18 @@
...
             %p
               %em New attendee!
           %p
             %small= invitation.note
-        .medium-4.small-12.columns
+        .col-4
           - if invitation.verified.blank? && @event.confirmation_required
             = link_to 'Verify', admin_event_invitation_verify_path(@event, invitation), class: 'button round expand tiny sucess', method: :post
-          = link_to 'Cancel attendance', admin_event_invitation_cancel_path(@event, invitation), class: 'button round expand tiny alert', method: :post
+          = link_to 'Cancel attendance', admin_event_invitation_cancel_path(@event, invitation), class: 'btn btn btn-danger', method: :post
```
matyikriszta, line 16:
```suggestion
          = link_to 'Cancel attendance', admin_event_invitation_cancel_path(@event, invitation), class: 'btn btn-small btn-danger', method: :post
```

Accidentally added the `btn` class twice. Also, just so the button is not so big can we try using a small button here? They really look huge in the interface 😂.

KimberleyCook: Done, updated the image in the PR too

matyikriszta: @KimberleyCook nice, I think it looks much better.

matyikriszta (approved):

PR #1729, opened 2022-04-06
Fix event sponsor alignment

KimberleyCook:
| Before | After |
|-------|------|
|Screenshot 2022-04-06 at 12 40 41|Screenshot 2022-04-06 at 12 55 07|

### DIFF
--- a/app/views/shared/_sponsors.html.haml
+++ b/app/views/shared/_sponsors.html.haml
@@ -1,4 +1,4 @@
-.row.align-items-center
+.row.d-flex.align-items-center.justify-content-center
   - sponsors.each do |sponsor|
-    .col-4.col-md-3.col-lg-2.mt-4.d-flex.justify-content-center
-      = link_to image_tag(sponsor.avatar, class: 'small-image mw-100', alt: sponsor.name), sponsor.website, title: sponsor.name, class: 'd-inline-block border-0'
+    .col-4.col-md-3.col-lg-2.mt-4
+      = link_to image_tag(sponsor.avatar, alt: sponsor.name), sponsor.website, title: sponsor.name, class: 'border-0 mw-100'
--- a/app/views/shared/_sponsors.html.haml
+++ b/app/views/shared/_sponsors.html.haml
@@ -1,4 +1,4 @@
 .row.d-flex.align-items-center.justify-content-center
   - sponsors.each do |sponsor|
     .col-4.col-md-3.col-lg-2.mt-4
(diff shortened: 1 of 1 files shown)


app/views/shared/_sponsors.html.haml
```diff
@@ -1,4 +1,4 @@
-.row.align-items-center
+.row.d-flex.align-items-center.justify-content-center
   - sponsors.each do |sponsor|
-    .col-4.col-md-3.col-lg-2.mt-4.d-flex.justify-content-center
-      = link_to image_tag(sponsor.avatar, class: 'small-image mw-100', alt: sponsor.name), sponsor.website, title: sponsor.name, class: 'd-inline-block border-0'
+    .col-4.col-md-3.col-lg-2.mt-4
+      = link_to image_tag(sponsor.avatar, class: '', alt: sponsor.name), sponsor.website, title: sponsor.name, class: 'border-0 mw-100'
```
matyikriszta, line 4:
```suggestion
      = link_to image_tag(sponsor.avatar, alt: sponsor.name), sponsor.website, title: sponsor.name, class: 'border-0 mw-100'
```

If we are not going to add a class onto the image itself we can just remove the class declaration from here.

matyikriszta (approved):

app/views/shared/_sponsors.html.haml
```diff
@@ -1,4 +1,4 @@
-.row.align-items-center
+.row.d-flex.align-items-center.justify-content-center
   - sponsors.each do |sponsor|
-    .col-4.col-md-3.col-lg-2.mt-4.d-flex.justify-content-center
-      = link_to image_tag(sponsor.avatar, class: 'small-image mw-100', alt: sponsor.name), sponsor.website, title: sponsor.name, class: 'd-inline-block border-0'
+    .col-4.col-md-3.col-lg-2.mt-4
+      = link_to image_tag(sponsor.avatar, alt: sponsor.name), sponsor.website, title: sponsor.name, class: 'border-0 mw-100'
```
matyikriszta, line 4: @KimberleyCook sorry, I just remembered there was a good reason why this image had the `small-image` class. It was to restrict the height of vertical logos, otherwise sometimes they'd get too tall. This class adds a `max-width: 100px` style to logos. I think we are going to be massaging this section for forever, it's really hard to find a style that works for the myriads of logo formats we have but this is needed.

KimberleyCook: Ok gotcha, I'll add it back.

matyikriszta (approved):

PR #1733, opened 2022-04-07
Fix undefined method avatar for nil:NilClass error on admin feedback page

matyikriszta:
### Description
As it says on the tin.

### Status
Ready for Review


matyikriszta:
@KimberleyCook this PR fixes the error with the feedback page.

PR #1738, opened 2022-04-12
Delete the shared _title partial

matyikriszta:
### Description
This PR deletes the old shared _title partial as it was only used in one place (on the Subscriptions page) and looked a bit outdated.

Might be worth looking into creating a new title partial to standardize on how we present page titles but this can wait until we finished migrating the site to Bootstrap.

Depends on #1737 which is the only other place we use title but that view is about to be deleted.

### Status
Ready for Review

### Screenshots
| Subscription page before | Subscription page after |
| --- | --- |
| Screenshot 2022-04-11 at 19-49-12 codebar | Screenshot 2022-04-11 at 19-48-55 codebar |

### DIFF
--- a/app/views/shared/_title.html.haml
+++ b/app/views/shared/_title.html.haml
@@ -1,6 +1,0 @@
-%section.hero.opaque
-  .row
-    .medium-offset-1.medium-7.columns
-      %h2= title
-      - if date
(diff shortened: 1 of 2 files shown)


KimberleyCook (approved):
🎉

PR #1744, opened 2022-04-19
[Bugfix] Updates waitlist form on workshop invitation page to not require the presence of a tutorial for coaches

matyikriszta:
### Problem
Two days ago a coach tried to join the waitlist for a workshop via the Join waiting list button on the workshop invitation page but couldn't because the model requires the presence of a tutorial which is not relevant for coaches.

The bug as reported by the coach:
image

### Solution
Updated the code to only check for the presence of a tutorial if a student tries to join the student waiting list. I also cheekily fixed a Rubocop offence in `app/controllers/waiting_lists_controller.rb` while I was there.

### Status
Ready for Review

### DIFF
--- a/app/controllers/waiting_lists_controller.rb
+++ b/app/controllers/waiting_lists_controller.rb
@@ -7,8 +7,13 @@ class WaitingListsController < ApplicationController
     return back_with_message(@invitation.errors.full_messages) unless @invitation.valid?(:waitinglist)
 
     @invitation.save && WaitingList.add(@invitation, auto_rsvp)
-    message = auto_rsvp ? 'You have been added to the waiting list' :
+
+    message = if auto_rsvp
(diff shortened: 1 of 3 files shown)


gnclmorais (approved):
Thanks for fixing this! Do you have any theories why we’ve only spotted this now? Why hasn’t it come up before for all the other workshops (remote or not) we’ve ran? 🤔

KimberleyCook (approved):

matyikriszta:
> Thanks for fixing this! Do you have any theories why we’ve only spotted this now? Why hasn’t it come up before for all the other workshops (remote or not) we’ve ran? 🤔

@gnclmorais my only theory is that the coach waiting list functionality is not used all that often. Maybe a couple of coaches tumbled upon this issue but did not bother reporting it up until now.

PR #1750, opened 2022-05-14
Stop emailing students & coaches without TOC accepted

gnclmorais:
After talking with Kim about this a few weeks ago, we realised it doesn’t make much sense to email members that have not accepted our terms about new events. We see that as not having come back to codebar in a while, thus making more sense to save our resources.

### DIFF
--- a/app/models/invitation_manager.rb
+++ b/app/models/invitation_manager.rb
@@ -29,8 +29,6 @@ class InvitationManager
 
   def invite_students_to_event(event, chapter)
     chapter_students(chapter).each do |student|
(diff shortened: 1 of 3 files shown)


gnclmorais:
I kind of want to raise again that our parameters for Code Climate feel a bit high… The issues raised feel a bit overzealous.

matyikriszta:
@gnclmorais was this PR ever finished? I think it would be still worth implementing this so I'm reopening the PR.

matyikriszta (approved):
Nice one 🎉, thanks @gnclmorais

Issue #1753, opened 2022-05-18
dashboard controller and event controller refactoration


carlosdanielpohlod:
## Description of the issue 📄
   Dasboard controller and event controller has methods that can be refactored. I will take this issue to solve.
## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

PR #1754, opened 2022-05-18
refactored dashboard_controller and events_controller; issue https://…

carlosdanielpohlod:
related issue: […github.com/codebar/planner/issues/1753](https://github.com/codebar/planner/issues/1753)

### DIFF
--- a/app/services/dashboard_service.rb
+++ b/app/services/dashboard_service.rb
@@ -1,0 +1,75 @@
+class DashboardService 
+
+    DEFAULT_UPCOMING_EVENTS = 5
+
+    def initialize(user = nil)
+      @user = user
+    end
+
+    def ordered_events  
+        upcoming_events_for_user.map.inject({}) do |hash, (key, value)|
+            hash[key] = EventPresenter.decorate_collection(value)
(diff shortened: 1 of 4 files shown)


app/services/dashboard_service.rb
```diff
@@ -0,0 +1,75 @@
+class DashboardService
```
wenderjean, line 1: Ideally your service fits better as a stateless object performing a single action. It seems to me that this service acts more like a query object than a service itself, for example:

```ruby
OrderedEventsQuery.new
UpcomingWorkshopsQuery.new
```

app/services/dashboard_service.rb
```diff
@@ -0,0 +1,75 @@
+class DashboardService 
+
+    DEFAULT_UPCOMING_EVENTS = 5
+
+    def initialize(user = nil)
+      @user = user
+    end
+
+    def ordered_events  
+        upcoming_events_for_user.map.inject({}) do |hash, (key, value)|
```
wenderjean, line 10: You have some indentation problems :nail_care:

PR #1763, opened 2022-06-03
Fix double underline bug on links after migrating to Bootstrap 5

matyikriszta:
### Description
This PR fixes the double underline issue on links that surfaced after migrating to Bootstrap 5 classes and removing Foundation.

### Status
Ready for Review


KimberleyCook (approved):
🎉

PR #1776, opened 2022-07-04
Replace Foundation with Bootstrap in README

matyikriszta:
### Description
Update the Front-end Framework section of the README to reference Bootstrap instead of Foundation


KimberleyCook (approved):
🎉

Issue #1765, opened 2022-06-07
Allow remote "Events"


KimberleyCook:
## Description of the issue 📄
Currently, you have to include a venue when creating an `Event`. I would like this to not be the case and have a checkbox with a Remote option. 

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

PR #1798, opened 2022-08-20
feat(1765): remote events

jhackett1:
fixes: #1765

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -74,7 +74,7 @@ class Admin::EventsController < Admin::ApplicationController
 
   def event_params
     params.require(:event).permit(
-      :name, :slug, :date_and_time, :begins_at, :ends_at, :description, :info, :schedule, :venue_id, :external_url,
+      :remote, :name, :slug, :date_and_time, :begins_at, :ends_at, :description, :info, :schedule, :venue_id, :external_url,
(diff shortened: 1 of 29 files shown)


app/controllers/admin/events_controller.rb
```diff
@@ -74,7 +74,7 @@ def set_event
 
   def event_params
     params.require(:event).permit(
-      :name, :slug, :date_and_time, :begins_at, :ends_at, :description, :info, :schedule, :venue_id, :external_url,
+      :remote, :name, :slug, :date_and_time, :begins_at, :ends_at, :description, :info, :schedule, :venue_id, :external_url,
```
matyikriszta, line 77: @jhackett1 can we rename this field to `virtual` please? So it aligns with the name of the field on `Workshop`.

jhackett1: done

matyikriszta:
> the remaining codeclimate issues _seem_ low priority to me. i can invest some more time in fixing them, but tbh i don't _think_ i caused them—they don't seem to be attached to any code i wrote or changed, even though they're showing up as "new" in the report.
> 
> lemme know what you think @KimberleyCook

@jhackett1 you are right, I have marked those issues for now as won't fix, they are not relevant to this change.

matyikriszta (commented):
@jhackett1 can you add an integration test for this change?

matyikriszta (approved):

Issue #1746, opened 2022-04-26
Be able to filter sponsors for just one chapter


jftsang:
## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

brunogirin:
I can see a few use cases for this:

## New workshops form

The list of sponsors should be filtered according to the chapter that is selected at the top. This is particularly important here as the list is long and we have some sponsors that have multiple entries because they sponsor us in multiple locations: if it's not filtered, we risk choosing the wrong one.

![Screenshot 2022-05-03 105033](https://user-images.githubusercontent.com/216028/166434378-dee7915d-4789-4311-aa09-1bc704344024.png)

## Sponsor list

The list of sponsors available in the admin could have a way to filter.

![Screenshot 2022-05-03 105027](https://user-images.githubusercontent.com/216028/166435829-8fe01e05-767b-47b8-abf2-b87f7bd2e996.png)

## Chapter menu

The admin menu for a chapter could have a "View {chapter name} sponsors" option that would open a filtered list of sponsors.

![Screenshot 2022-05-03 105142](https://user-images.githubusercontent.com/216028/166435928-3e71e66d-5f18-4d8a-9ed5-78a215f87326.png)

PR #1799, opened 2022-08-20
feat(1746): filter sponsors by chapter

jhackett1:
fixes: #1746


matyikriszta:
@jhackett1 can we write an integration test for this please? Plus looks like older tests might have to be updated too.

jhackett1:
added a passing feature test

matyikriszta (approved):

Issue #1493, opened 2021-03-18
Update timezone to an "Event"


KimberleyCook:
## Description of the issue 📄
When creating an "Event" on the codebar website defaults to UK time, with the expansion of virtual events it would be good to attach a timezone to the event. Have it an optional field, with UK being the default, but change if another is added.

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

PR #1800, opened 2022-08-20
feat(1493): events have timezones

jhackett1:
fixes: #1493

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -224,6 +224,7 @@ ActiveRecord::Schema[7.0].define(version: 2023_08_03_225601) do
     t.boolean "surveys_required", default: false
     t.string "audience"
     t.boolean "virtual", default: false, null: false
+    t.string "time_zone", default: "London", null: false
     t.index ["slug"], name: "index_events_on_slug", unique: true
     t.index ["venue_id"], name: "index_events_on_venue_id"
   end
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -235,6 +235,8 @@
     t.boolean  "confirmation_required", default: false
     t.boolean  "surveys_required",      default: false
     t.string   "audience"
+    t.boolean  "remote",                default: false, null: false
```
matyikriszta, line 238: @jhackett1 this change sneaked in here from another P.

KimberleyCook (approved):
🎉

biggianteye:
This PR has been approved but hasn't been merged. Is there anything that's missing that would stop it from being merged (other than being out of sync given how old it is)?

matyikriszta:
@biggianteye yes, we have tested this PR further and it does not actually work. I'm going to remove the Approval for now. We'll have to revisit this issue, it's on our roadmap.

db/schema.rb
```diff
@@ -224,6 +224,7 @@
     t.boolean "surveys_required", default: false
     t.string "audience"
     t.boolean "virtual", default: false, null: false
+    t.string "timezone", default: "London", null: false
```
matyikriszta, line 227: Note to self, time zone is two words so change this to time_zone

gnclmorais (approved):
Nice change! Seems to work locally:
image

matyikriszta (approved):

PR #1809, opened 2022-09-10
Delete unused/deprecated CSS

matyikriszta:
### Description
Delete some old CSS that was deprecated as part of the migration from Foundation to Bootstrap.

### Status
Ready for Review


app/assets/stylesheets/main.scss
```diff
@@ -98,6 +30,10 @@ form .hint {
   max-height: 100px;
 }
 
+.sponsor-sm {
+  max-width: 120px;
```
KimberleyCook, line 34: is this not a class in bootstrap?

KimberleyCook (approved):
One small question, but any PR that has a bunch of delations I'm a fan of so nice :)

app/assets/stylesheets/main.scss
matyikriszta: @KimberleyCook unfortunately no, would be with Tailwind but not Bootstrap.

PR #1811, opened 2022-09-15
Delete compass-rails gem from Gemfile

matyikriszta:
### Description
We do not use compass anymore, into the sea with it! Also, it's no longer supported and maintained.

### Status
Ready for Review


KimberleyCook (approved):
🎉

PR #1814, opened 2022-09-27
Updates Dockerfile to run locally

tbutterwith:
While trying to run the app via Docker I noticed the image doesn't build anymore. This PR

- Adds the `--force-yes` flag which is required when using the `-y` flag in `apt-get`
- Adds the nodesource GPG key which is required in the Chome dev package install


KimberleyCook (approved):
Thanks for updating this @tbutterwith 🎉

PR #1819, opened 2022-10-04
New banner image and added another button

KimberleyCook:
Updated the banner image, added another CTA to sign up, made that text a bit bigger and made our logo a bit bigger (feedback was it's a little underwhelming)

| Before | After |
|-------|------|
|screencapture-localhost-3000-2022-10-04-19_01_57|screencapture-localhost-3000-2022-10-04-19_20_17|


priyapatil1 (approved):

matyikriszta:
I was going to comment that the overlay text is really hard to read now on the new darker background. Can we maybe swap it to white bg and black text to see if that's easier to read?

matyikriszta:
But really, we should remove that overlay altogether and redesign the hero so the missing statement is bigger and it's on it's own (not overlayed over an image).

PR #1820, opened 2022-10-16
Don’t raise an exception on SIGTERM

gnclmorais:
image

Looking at our Rollbar quota, the large majority of the errors we get there come from, I believe, the recurring Heroku restart that happens. I see no reason to report these to Rollbar, since they are expected and, at least while using Heroku, necessary.

Luckily, looks like [Puma has an option for this][0], so it makes sense to me to turn it on (or rather off, since we’re setting it to `false`). [Found it][1] while doing some research on this subject, since I’ve had a feeling I’ve seen it before.

[0]: https://www.rubydoc.info/gems/puma/Puma%2FDSL:raise_exception_on_sigterm
[1]: https://github.com/puma/puma/issues/1438

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -6,6 +6,7 @@ preload_app!
 rackup      DefaultRackup
 port        ENV['PORT']     || 3000
 environment ENV['RACK_ENV'] || 'development'
+raise_exception_on_sigterm false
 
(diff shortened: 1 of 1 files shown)


KimberleyCook (approved):
Thaaaanks

PR #1821, opened 2022-10-19
Spanish Strings for Admin Info

103percent:


KimberleyCook (approved):
Thank you so much @103percent

PR #1824, opened 2022-11-02
Add more context to background jobs

gnclmorais:
For some reason, these background jobs seem to break frequently. Heroku logs are not telling us much, so hopfully with this commit:
- we’ll get more information thanks to the `--trace` option
- we’ll stop the worker when there is no works to do, instead of keeping it running


gnclmorais:
This seems to happen every 10 minutes, which is the frequency we run this task:
image

matyikriszta (approved):

gnclmorais:
I can’t merge it 😭 
image

PR #1826, opened 2022-11-10
Adds a photo to the initial sign up page

KimberleyCook:
| Before | After |
|-------|------|
|screencapture-localhost-3001-member-new-2022-11-10-16_36_07|screencapture-localhost-3001-member-new-2022-11-17-19_34_07|


app/views/members/new.html.haml
```diff
@@ -17,6 +17,7 @@
           = t('members.new.students.criteria_brief')
         %p.lead
           = t('members.new.students.github')
+        = image_tag("nyc-workshop.jpg", alt: "a photo from a codebar NYC workshop", class: "mw-100 pb-4")
```
gnclmorais, line 20: Maybe we can also add a small description of what is shows…?

gnclmorais (approved):

app/views/members/new.html.haml
```diff
@@ -1,26 +1,30 @@
 .container.py-4.py-lg-5
   .row.justify-content-md-center
-    .col-md-10.col-lg-8
+    .col
       %h1= @page_title
 
 .py-4.py-lg-5.bg-light
   .container
     .row.justify-content-md-center
```
matyikriszta, line 8: @KimberleyCook we could remove this .row and the .col before and use the .row on line 10.

matyikriszta:
@KimberleyCook I hope you don't mind, I checked out and reworked this PR a little. This is what I came up with in terms of layout, I wanted to get rid of that unsightly student button.

| Mobile | Desktop |
| --- | --- |
| Screenshot 2023-10-23 at 17-36-26 codebar | Screenshot 2023-10-23 at 17-44-16 codebar |

PR #1829, opened 2022-11-14
Update EventCalendar service to only include address if event has one

matyikriszta:
### Description
We recently introduced remote events in #1798 but forgot to update the `EventCalendar` service to only include the address in the invitation if the event has one.

### Status
Ready for Review

### DIFF
--- a/lib/services/event_calendar.rb
+++ b/lib/services/event_calendar.rb
@@ -13,13 +13,13 @@ class EventCalendar
   private
 
   def setup_event
-    address = AddressPresenter.new(event.venue.address)
+    address = AddressPresenter.new(event.venue.address) if event.venue.present?
     calendar.event do |e|
       e.organizer = event.email.to_s
       e.dtstart = event.date_and_time
       e.dtend = event.ends_at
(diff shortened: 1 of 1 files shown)


lib/services/event_calendar.rb
```diff
@@ -13,13 +13,13 @@ def calendar
...
   def setup_event
-    address = AddressPresenter.new(event.venue.address)
+    address = AddressPresenter.new(event.venue.address) if event.venue.present?
     calendar.event do |e|
       e.organizer = event.email.to_s
       e.dtstart = event.date_and_time
       e.dtend = event.ends_at
       e.summary = event.name
-      e.location = address.to_s
+      e.location = address.to_s if address.present?
```
gnclmorais, line 22: If you want something shorter:
```rb
e.location = address&.to_s
```

gnclmorais (approved):

lib/services/event_calendar.rb
matyikriszta: @gnclmorais good point, updated!

PR #1843, opened 2023-01-19
Prevent pagination injection errors

gnclmorais:
The following Rollbar errors inspired this PR:
- https://rollbar.com/codebar-production/codebar-production/items/384
- https://rollbar.com/codebar-production/codebar-production/items/359
- https://rollbar.com/codebar-production/codebar-production/items/357

These three instances are the biggest culprits in filling up our Rollbar instance. These are not the kinds of errors we want to keep an eye out for, these are just machines poking around trying to inject SQL into our website.

This pull request will 1) prevent any SQL injection on the pages we use, and also 2) by sanitising input to pagination, prevent errors that can be avoided from adding noise to the signal we want Rollbar to provide.

Here are two examples of what the `.to_i` does here:
image


gnclmorais:
Tests passing!

matyikriszta (approved):
Thanks @gnclmorais

PR #1861, opened 2023-04-19
Amend welcome email copy

KimberleyCook:
The London chapter does send emails now so I have removed that part, and tweaked a few other bits.

### DIFF
--- a/app/views/member_mailer/welcome_student.html.haml
+++ b/app/views/member_mailer/welcome_student.html.haml
@@ -4,13 +4,7 @@
 
 %p
   - if @member.subscriptions.any?
-    You'll automatically receive email invitations to our workshops as soon as they are open for RSVP.
-    - if @member.chapters.pluck(:name).include?("London")
-      However, our London chapter doesn't send workshop invitations via email, so you won't receive email invites for regular London events. We open RSVPs at around 1pm on the Friday before the workshop, and you will be able to register through the workshop page on our website. To be notified when we open RSVPs, join the #london-chapters channel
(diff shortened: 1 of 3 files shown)


matyikriszta:
@KimberleyCook we'll have to update the tests for these emails as well, I can do that if you want so we can merge the PR.

KimberleyCook:
@matyikriszta completely forgot about this PR, yes please if you have time :)

app/views/member_mailer/welcome_student.html.haml
```diff
@@ -4,26 +4,18 @@
...
   - if @member.subscriptions.any?
-    You'll automatically receive email invitations to our workshops as soon as they are open for RSVP.
-    - if @member.chapters.pluck(:name).include?("London")
-      However, our London chapter doesn't send workshop invitations via email, so you won't receive email invites for regular London events. We open RSVPs at around 1pm on the Friday before the workshop, and you will be able to register through the workshop page on our website. To be notified when we open RSVPs, join the #london-chapters channel
-      = link_to 'on our Slack', 'http://slack.codebar.io'
-      or
-      = link_to 'follow us', 'https://twitter.com/codebarLDN'
-      on Twitter.
-  - else
-    We send out invitations to our events via email.
```
matyikriszta, line 15: @KimberleyCook I've updated the tests to pass, but was wondering, do we want to remove this link from the email? I think it would still be useful to have it.

KimberleyCook: Yeah, let's not remove it.

matyikriszta (approved):

PR #1871, opened 2023-07-04
Raise 404 when event is not found

gnclmorais:
fixes rb#464

We were trying to call a method on the `venue` of an event that was never found.


matyikriszta (approved):
Thanks @gnclmorais

PR #1872, opened 2023-07-08
Ignore subscription if not found

gnclmorais:
If a user is trying to unsubscribe from a non-existing subscription (i.e. no longer a student/coach), just say yes and move on. No need to record that on Rollbar. Rely on the group to show the success toast.

Fixes: https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/474


KimberleyCook (approved):
Amazing, thanks for this @gnclmorais

PR #1875, opened 2023-07-16
Docker on GitHub actions

gnclmorais:


Dockerfile
```diff
@@ -1,22 +1,22 @@
...
 
 WORKDIR /planner
 
+# Install project dependencies using all available CPU cores
 COPY Gemfile Gemfile.lock ./
-RUN bundle install
+RUN bundle install --jobs $(getconf _NPROCESSORS_ONLN) --retry 3
+
+# Setup our database
+RUN bundle exec rake db:create db:migrate db:seed db:test:prepare
```
asideofcode-dev, line 20: Fly by comment: I wouldn't imagine this would work since within the build environment there is no database

gnclmorais: Yeah, I’m not the most well-versed in Docker, so I was just giving it a stab, see where I’d go… What you said makes sense now and it explains why it was not working! 😬

gnclmorais:
Giving up on this, maybe one day someone else will pick it up…

Issue #1881, opened 2023-07-23
Navbar toggler disappears after 2 clicks


asideofcode-dev:
## Description of the issue 📄

The navbar toggle disappears after 2 clicks. This is because there's a global css rule to hide elements with the `.collapsed` class:
https://github.com/codebar/planner/blob/2d4726b292270ed1f238b1debbb59ab95d9e2d31/app/assets/stylesheets/main.scss#L6-L8

## Screenshots 📷

<img width="674" alt="image" src="https://github.com/codebar/planner/assets/133222359/3a3a52f5-377f-4ad0-b58f-a58bf3ee5802">

## Steps to fix 🛠

Git blame suggested this global css rule used to be more specific as `.sign-up-wizard .collapsed` but was made general via https://github.com/codebar/planner/pull/938. It can be fixed by either 
1. ~Making the global css rule more specific~
2. ~Creating a specific rule for the navbar e.g.~

   ```css
   button.navbar-toggler.collapsed {
     display: block;
   }
   ```
3. See the comments for a better approach

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

asideofcode-dev:
After a little digging it looks like the specific form of the css rule was introduce in this PR https://github.com/codebar/planner/pull/938. It turns out that the css rule is referring to `.collapsed` in relation to some custom toggle functionality https://github.com/codebar/planner/blob/2d4726b292270ed1f238b1debbb59ab95d9e2d31/app/assets/javascripts/subscriptions-toggle.js#L8

This issue is that bootstrap has already this css class. A reasonable fix is to just namespace the custom toggle functionality and use `.codebar-collapsed` isntead.

Issue #1880, opened 2023-07-22
modernizr is nowhere to be found


asideofcode-dev:
## Description of the issue 📄

There's an include tag for `modernizr` in the application view template.

https://github.com/codebar/planner/blob/2d4726b292270ed1f238b1debbb59ab95d9e2d31/app/views/layouts/application.html.haml#L13

Perhaps this used to be available in some earlier version of some gem, but now it's nowhere to be found. If you run the site from scratch you'll notice a 404 for that particular resource in the Javascript console.

I'm a little bit confused about how this file is available in production, though I suspect caching might be to blame.

I also noticed this warning in the logs:
```
 (called from _app_views_layouts_application_html_haml__4168757267524813890_60640 at /planner/app/views/layouts/application.html.haml:13)
DEPRECATION WARNING: The asset "vendor/modernizr.js" is not present in the asset pipeline.
Falling back to an asset that may be in the public folder.
This behavior is deprecated and will be removed.
To bypass the asset pipeline and preserve this behavior,
use the `skip_pipeline: true` option.
```

## Steps to fix 🛠

Not sure if necessary. Might be a good time to be rid of it.

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

matyikriszta:
@asideofcode-dev if I remember correctly an older version of Foundation we were using required modernizer which is why that line is there. We have now migrated to Bootstrap which does not rely on modernizer (or if it does under the hood it's bundled with the gem and does not have to be imported separately) so this could be removed.

PR #1882, opened 2023-07-23
Update to Ruby 2.7.2 and Rails 5.2.8.1 

asideofcode-dev:
## What does the PR do ?

Resolves https://github.com/codebar/planner/issues/1881

Resolves https://github.com/codebar/planner/issues/1883

Resolves https://github.com/codebar/planner/issues/1880

Update to Ruby 2.7.2 and Rails 5.2.8.1.

The main idea is to update the gems, run [`rails app:update`](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-update-task) (while keeping an eye on the [upgrade documentation](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0), then put in place the fixes necessary to make sure things are still working. 

1.  Updates `render text: ` to `render plain:`
2. Update `.gitignore`
3. Update `jquery-ui-rails` and `boostrap`
4. For `bootstrap`, change `data-x` to `data-bs-x` e.g. `data-toggle`
5. Reduce the count of some of the instances created by the seed file
6. Docker related things to make developing easier
7. Replace `Planner::Application` with `Rails.application`
8. Moved from `secret_token` to `secret_key_base`
9. Updated request specs to use params. See below
10. For Fabricator, update `after_save` to `after_create`
11.  Use `allow(ENV).to receive(:[]).and_call_original` to avoid stub issues in updated rspec
12. Take screenshots at failure time after js feature tests
13. Added a temporary fix for webdrivers issue, see https://github.com/titusfortner/webdrivers/issues/247
14. Moved from `jquery_ujs` to `rails-ujs`
15. Fixed `ajax:success` and firing correct events for form submission on remote forms for invitations
16. Namespaced toggle related CSS classes from `toggle` and `collapsed` to `codebar-x`
17. Updated `redirect_to :back` to `redirect_back fallback_location: root_path`
18. Defined `ApplicationRecord` and `ApplicationMailer` as per upgrade instructions
19. Fixed haml doctype `!!!`
20. `Bullet.raise = false` in tests. Maybe Bullet got better at detecting n+1 queries, I tried to fix some of them but there's quite a few. Best to scope that as a different effort. The queries don't seem to have caused issues until now.

For (9):
```
-      put :update, id: invitation.token, workshop_id: workshop.id, attending: "true"
+      put :update, params: { id: invitation.token, workshop_id: workshop.id, attending: "true" }
```


Gemfile
```diff
@@ -53,6 +56,12 @@ gem 'public_activity'
 group :development do
   gem 'better_errors'
   gem 'letter_opener'
+  # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
+  gem 'web-console', '>= 3.3.0'
+  gem 'listen', '>= 3.0.5', '< 3.2'
+  # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
+  gem 'spring'
```
matyikriszta, line 63: @asideofcode-dev just a quick question, was this needed for the upgrade or did you add it for other reasons (e.g. the performance boost it gives in development mode)

matyikriszta:
@asideofcode-dev could you elaborate on point 12? Why was this added? To help debug tests?

app/assets/config/manifest.js
matyikriszta, line 1: @asideofcode-dev is this a file required by Rails 5?

Gemfile
asideofcode-dev: I use the `rails app:update` task to reconcile boilerplate files between versions (as illustrated here https://railsdiff.org/4.2.11.3/5.2.8). This gem came from there.

app/assets/config/manifest.js
asideofcode-dev: This is a file required by sprockets, which was bumped as part of going to Rails 5. See  https://github.com/rails/sprockets/blob/main/UPGRADING.md#manifestjs

asideofcode-dev:
@matyikriszta Yup, 12 was added to help debug feature tests that use Javascript. There's a class of issues that a screenshot helps debug

Gemfile
matyikriszta: Ok, makes sense, thanks for clarifying

app/assets/config/manifest.js
matyikriszta: Cool, thanks for clarifying

matyikriszta:
@asideofcode-dev there's a bunch of new files in the `bin` folder, were those generated by `rails app:update`?

app/views/layouts/application.html.haml
```diff
@@ -9,16 +9,17 @@
     %title
       = retrieve_title
     = favicon_link_tag 'favicon.ico'
-    = stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': true
-    = javascript_include_tag 'vendor/modernizr'
-    = javascript_include_tag 'application', 'data-turbolinks-track': true
+    = stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload'
+    // = javascript_include_tag 'vendor/modernizr'
```
matyikriszta, line 13: @asideofcode-dev might as well just delete this line, unless you think we might add it back later for some reason?

asideofcode-dev:
@matyikriszta Happy to add that in an upcoming commit. Don’t wanna get in
the way of your current review

config/environments/test.rb
```diff
@@ -35,12 +41,23 @@
...
+  
+  # Configure static asset server for tests with Cache-Control for performance.
+  config.serve_static_files = true
+  config.static_cache_control = 'public, max-age=3600'
+
   # Fake omniauth for testing
   OmniAuth.config.test_mode = true
 
   config.after_initialize do
+    # TODO: re-enable and go through failures
```
matyikriszta, line 58: @asideofcode-dev can we create a Github issue for this? We want to make sure we remember to do it.

asideofcode-dev: Done, here you go https://github.com/codebar/planner/issues/1885

matyikriszta: Thanks :)

asideofcode-dev:
@matyikriszta  Yup, can do. I'll batch up the work once the review is done

matyikriszta:
@asideofcode-dev while testing the site in the browser I noticed a couple of places where the Bootstrap `data-` attribute was not updated resulting in broken functionality:
- note modal on the admin member page cannot be closed with the close button: /planner/app/views/admin/members/_note.html.haml
- none of the alerts can be dismissed, the close button does not work: /planner/app/views/layouts/_messages.html.haml
- the Testimonials carousel on the homepage doesn't seem to work: /planner/app/views/members/_testimonials.html.haml

matyikriszta:
@asideofcode-dev this is a small cosmetic issue but on desktop the dropdown Menu gets "cut off" when open.
Screenshot 2023-07-26 at 11 47 35 am

asideofcode-dev:
@matyikriszta  Thanks for flagging those `data-` related issues. I should have really done this in the first instance...

```
➜  ~ curl https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/js/bootstrap.min.js | grep -o -E 'data-bs-[a-zA-Z0-9_-]+' | sort | uniq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 60089  100 60089    0     0    99k      0 --:--:-- --:--:-- --:--:--  100k
data-bs-content
data-bs-dismiss
data-bs-interval
data-bs-no-jquery
data-bs-original-title
data-bs-parent
data-bs-ride
data-bs-slide
data-bs-slide-to
data-bs-spy
data-bs-target
data-bs-toggle
```

Updates coming shortly

matyikriszta:
@asideofcode-dev just one more thing, according the latest [Bootstrap 5 documentation](https://getbootstrap.com/docs/5.2/components/close-button/) the way to declare the a close button has slightly changed. The class has changed from `.close` to `.btn-close` and it not longer requires/supports a %span inside it with an `aria-hidden="true"` attribute on it. Can we update the close buttons? Currently the styles are broken. This is everywhere where close buttons are used (alerts, modals, etc).

Screenshot 2023-07-27 at 3 27 33 pm

matyikriszta (approved):
Once the issue with the close button is resolved this is good to go. Once merged going to do another round of testing on staging. Thanks @asideofcode-dev 🎉

asideofcode-dev:
Thanks @matyikriszta. It really helps to have someone who knows how things ought to look.

New commits should address the remaining items
1. Fix navigation menu alignment
2. Fix close button

gnclmorais:
Note: in order to run `bin/rails s` locally, I need to set `OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES` ([like so](https://stackoverflow.com/a/53404317/590525), for example).

biggianteye:
Hi @asideofcode-dev. Do you know if this change was intentional?

```diff
commit 32f888ebc2567f286575ef61c10ee6429436d5a8
Author: asideofcode-dev <133222359+asideofcode-dev@users.noreply.github.com>
Date:   Sun Jul 23 13:03:23 2023 +0000

    Run rails app:update

diff --git a/docker-compose.yml b/docker-compose.yml
index e6e73e81..55c27f50 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,13 +10,7 @@ services:
     env_file:
       - docker-compose.env
     build: .
-    command:
-      - bundle
-      - exec
-      - rails
-      - server
-      - --binding=0.0.0.0
-      - --port=3000
+    command: tail -f /dev/null
     environment:
       DB_HOST: db
     volumes:
```

The consequence of this change is that anyone running `./bin/dstart` as per the README will get a Docker container that starts up but does nothing. The expectation based on the instructions in the README is that the web server would be running in the container.

asideofcode-dev:
@biggianteye Yes, the change was intentional. By updating the command in `docker-compose.yml` to `tail -f /dev/null`, we allow the Docker container to remain active indefinitely without actively running the server process. This setup enables developers to manually start the server or other services within the container at their discretion, facilitating a more flexible development environment. This approach also reduces the dependency on the server process being the primary activity of the container, which can be beneficial for certain testing and development workflows.

biggianteye:
Thanks for that information. If that's the way you want people to work, it would be good to update the README to reflect this way of working. Currently it says this:

> 5. Start the app
>
> Run bin/dstart to start the app.
>
>This should run a process which starts a server in a Docker container on your computer. This process won't finish - you'll know the server is ready when it stops doing anything and displays a message like this:
>
>     Rails 4.2.11 application starting in development on http://localhost:3000

PR #1887, opened 2023-07-29
Use ruby-terser to compress javascript

asideofcode-dev:
It seems some of our dependencies use ES6 and `uglifier` seems to be struggling to carry out the compression task.
The `uglifier` gem suggests using ruby-terser in such cases, see https://github.com/lautis/uglifier#uglifier.


matyikriszta (approved):
Thanks @asideofcode-dev

PR #1892, opened 2023-08-03
Update Ruby to 3.2.2 and Rails to 6.1.7.4

### LINKED ISSUE #1890: Duplication of assets under ./app and under ./public
## Description of the issue 📄

There are some images under both folders. I'm not sure why, they could all just be under one folder

## To do 📋

* [ ] Submit a pull request

asideofcode-dev:
1. Update Ruby to 3.2.2 and Rails to 6.1.7.4
2. Ran the `rails app:update` tasks as usual
3. ~(1) includes adding webpacker, the new mechanism for compiling JS. But I think this repo is pretty much married to Sprockets. Maybe webpacker can be forward looking. Or the project can opt to not use it, in which case I can rip it out.~ Took out webpacker since it will be removed in Rails 7
4. Update seeds so that when you go to the coaches page it's not empty
5.  Dedupe by moving assets in `./public` to `./app/assets`. Closes #1890. It turns out that the assets under `./public` were set to go since `config.assets.unknown_asset_fallback` is set to false by default in Rails 6.x


app/assets/images/codebar-girl.gif
matyikriszta, line 1: @asideofcode-dev can we check whether the 404 page still displays this image? I have a feeling these images were in public because they are used on the 404 page and if I remember correctly the 404 page only has access to assets in /public. Might be old behavior though, not sure if newer Rails versions work this way.

asideofcode-dev: It should work. Some context:

It was that 404 page that resulted in moving things into `./app/assets`. After updating to Rails 6 the tests started producing errors like:
```
 The asset "codebar-girl.gif" is not present in the asset pipeline.
```

Did some digging, and found out that the image on the 404 page is rendered like so:
```html
=image_tag 'codebar-girl.gif'
```
Internally, the helper `image_tag` will try to resolve this image. Before Rails 6, if the image could not be found within the asset pipeline then Rails would fallback to looking in the public folder. This behavior is now deprecated, but you would only get the warning, see below, and not error, by turning on the fallback, with `config.assets.unknown_asset_fallback = true`. Rails defaults to disabling the fallback and that's why I saw errors in the first instance 😢  and spent hours trying to figure out what happened.

```
DEPRECATION WARNING: The asset "codebar-girl.gif" is not present in the asset pipeline.
Falling back to an asset that may be in the public folder.
To bypass the asset pipeline and preserve this behavior,
use the `skip_pipeline: true` option.
```

> While Rails did have a fallback mechanism to look for assets in the public folder before version 6, it's considered a better practice to use the asset pipeline and avoid using the public folder for assets. This is because the asset pipeline provides various benefits and is designed to handle asset management more efficiently and consistently.

matyikriszta: Thanks for the explanation, that makes sense

matyikriszta:
@asideofcode-dev re: webpacker, I don't think we are going to use it anytime soon so let's just leave it out. It looks like it's retired as well and not in active development anymore: https://github.com/rails/webpacker. I think it's been ripped out of Rails 7 anyways (and that's where we want to end up with this upgrade).

config/initializers/omniauth.rb
```diff
@@ -13,3 +13,4 @@
 end
 
 OmniAuth.config.allowed_request_methods = [:post, :get]
+OmniAuth.config.silence_get_warning = true
```
matyikriszta, line 16: @asideofcode-dev is this needed because Omniauth v2.0.0 only allows post requests? But we are also allowing get on line 15?

matyikriszta (commented):
@asideofcode-dev can we also update the Ruby version in the .ruby-version file?

config/initializers/omniauth.rb
asideofcode-dev: Yup, that's it. `GET` is considered unsafe. Technically we can migrate to using `POST` but the signin/signup flow is currently using redirects in various places which don't play well with `GET` e.g. `AuthServicesController#new`

matyikriszta (approved):

PR #1894, opened 2023-08-04
Add missing .btn class to buttons on member/new page

matyikriszta:
### Description
The two buttons on the member/new page were missing the required .btn class. Looks like previously Bootstrap inferred buttonness from .btn-primary etc but latest Bootstrap requires the .btn class.

### Status
Ready for Review

### Screenshots
| Before | After |
| --- | --- |
| Screenshot 2023-08-04 at 12 16 53 am | Screenshot 2023-08-04 at 12 16 32 am |


KimberleyCook (approved):
Nice catch 🎉

PR #1897, opened 2023-08-09
Migrate Bootstrap spacing classes to latest v5 version

matyikriszta:
### Description
The latest Bootstrap version updated spacing classes from left (l) and right (r) to start (s) and end (e). We need to update our classes accordingly.

### Status
Ready for Review


KimberleyCook (approved):
Nice, good catch 🎉

PR #1916, opened 2023-09-02
gShow today’s workshop(s) as upcoming

gnclmorais:
As an admin, something that is constantly annoying to me is losing quick access to the workshop I’m organising today because it is past 6.30pm and it no longer shows up in the “Upcoming workshops” section.

Adding a new scope that includes today’s workshops as well, so that organisers have access to the workshop at hands while running it.

image


KimberleyCook (approved):
Nice amend, thanks @gnclmorais

PR #1918, opened 2023-09-02
Fix two rogue messages on tests

gnclmorais:
These two messages would appear on some tests, on the console. With this commit those messages are still caught but don’t show up on the console anymore (in the middle of RSpec’s dots).

The messages were:
- You have to provide an email address. Usage: rake member:delete'[email@address.com]'
- Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead.

Technically, the second one is not a message we display on the code, but rather an RSpec warning telling us one of our variables was nil. Adding content to the constructor was the solution.


matyikriszta (approved):
Thanks @gnclmorais

Issue #1926, opened 2023-09-14
Deprecated Node install script when building Docker image


CodeHex:
## Description of the issue 📄
Running the line
`curl -fsSL https://deb.nodesource.com/setup_20.x` in the Dockerfile comes up with a script deprecation warning and freezes the build for 60 secs. Looking at the contents of the scrpt, the warning says

```
This script, located at ${bold}https://deb.nodesource.com/setup_X${normal}, used to
  install Node.js is deprecated now and will eventually be made inactive.

  Please visit the NodeSource ${bold}distributions${normal} Github and follow the
  instructions to migrate your repo.
  ${underline}${green}${bold}https://github.com/nodesource/distributions${normal}

  The ${bold}NodeSource${normal} Node.js Linux distributions GitHub repository contains
  information about which versions of Node.js and which Linux distributions
  are supported and how to install it.
  ${underline}${green}${bold}https://github.com/nodesource/distributions${normal}


                          ${underline}${bold}${yellow}SCRIPT DEPRECATION WARNING${normal}
"

        echo
        echo "${cyan}${bold}TO AVOID THIS WAIT MIGRATE THE SCRIPT${normal}"
        echo "Continuing in 60 seconds (press Ctrl-C to abort) ..."
        echo
        sleep 60
```
## Steps to fix 🛠
Replace using the script with the [updated commands](https://github.com/nodesource/distributions/blob/master/README.md#debian-and-ubuntu-based-distributions) described in the nodesource repo readme.

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

CodeHex:
Hi 👋, I'm happy to have a crack at fixing this if its ok for me to work on it

asideofcode-dev:
@CodeHex Thanks for creating the issue. It's a recent deprecation.

Please go ahead. Happy to review the work

matyikriszta:
As of 26 Jul 2024 the install scripts are working again. Closing this issue.

PR #1900, opened 2023-08-12
Apple Silicon support for Docker

biggianteye:
Building the application using Docker on Apple Silicon (M1, M2, etc) fails with the following error in the step where it attempts to install Google Chrome:

```
7.332 + apt-get install -y --force-yes google-chrome-stable --no-install-recommends
7.341 Reading package lists...
7.723 Building dependency tree...
7.805 Reading state information...
7.864 W: --force-yes is deprecated, use one of the options starting with --allow instead.
7.864 E: Unable to locate package google-chrome-stable
------
failed to solve: process "/bin/sh -c set -ex;     wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | apt-key add -     && wget -q -O - https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -     && sh -c 'echo \"deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main\" >> /etc/apt/sources.list.d/google.list'     && apt-get update     && apt-get install -y --force-yes google-chrome-stable --no-install-recommends     && rm -rf /var/lib/apt/lists/*" did not complete successfully: exit code: 100
```

By default, Docker will use images that match the host architecture (ie. `linux/arm64/v8` on Apple Silicon). The above `apt` repo does not have an ARM version of Google Chrome.

This PR addresses the problem by making it explicit that Docker should always use the `linux/amd64` versions of images.

It also addresses another Apple Silicon-related error message and a bundler-related warning. See the individual commit messages for more details.

---

Note that the main goal of this PR is to get _building_ of the project to complete via Docker.

PR #1927, opened 2023-09-14
Replace deprecated script and install Node directly from nodesource

CodeHex:
Installing Node from nodesource scripts has been very recently deprecated (and sleeps the build for 60s). This replaces the script with installing the signed debian repo directly using the instructions provided [here](https://github.com/nodesource/distributions#debian-and-ubuntu-based-distributions).

Fixes https://github.com/codebar/planner/issues/1926

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,15 @@
 FROM ruby:3.2.2
 
 # Default node version on apt is old. This makes sure a recent version is installed
-# This step also runs apt-get update
-RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
+# Instructions for installing node from nodesource can be found here:
+# https://github.com/nodesource/distributions
+RUN set -ex; \
+    apt-get update \
+    && apt-get install -y ca-certificates curl gnupg \
+    && mkdir -p /etc/apt/keyrings \
+    && curl -fsSL https://deb.nodesource.com/gpgkey/nodesource-repo.gpg.key | gpg --dearmor -o /etc/apt/keyrings/nodesource.gpg \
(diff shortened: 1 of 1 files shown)


CodeHex:
When I tested this, I had to use the changes in PR https://github.com/codebar/planner/pull/1900, as I'm working on an M1 Mac. It got through the install steps fine, but someone might want to double check it on a different machine

biggianteye:
> When I tested this, I had to use the changes in PR #1900, as I'm working on an M1 Mac.

I'm glad you used them. Good to get some independent confirmation that they work.

CodeHex:
> > When I tested this, I had to use the changes in PR #1900, as I'm working on an M1 Mac.
> 
> I'm glad you used them. Good to get some independent confirmation that they work.

👍 Yeah thanks, I did get a couple of test failures but running the website seems fine

Dockerfile
```diff
@@ -1,8 +1,15 @@
 FROM ruby:3.2.2
 
 # Default node version on apt is old. This makes sure a recent version is installed
-# This step also runs apt-get update
-RUN curl -fsSL https://deb.nodesource.com/setup_20.x | bash -
+# Instructions for installing node from nodesource can be found here:
+# https://github.com/nodesource/distributions
+RUN set -ex; \
```
akarsh, line 6: Why is `set -ex;` needed?

matyikriszta:
Looks like the install scripts have been undeprecated? https://github.com/nodesource/distributions now suggest we use the script we have in the current Dockerfile.

The docs at https://github.com/nodesource/distributions/blob/master/README.md#faq and https://github.com/nodesource/distributions/wiki/Repository-Manual-Installation#-simplified-installation-using-scripts suggest that the script at https://deb.nodesource.com/setup_20.x is supported and recommended.

Reading the source of https://deb.nodesource.com/setup_20.x it no longer contains any sleep behaviour.

Issue #1928, opened 2023-09-14
Coaches webpage gives 404 when moving to next page


CodeHex:
## Description of the issue 📄
On the `/coaches` route, clicking on either a number or `Next ->` on the paging control at the bottom of the page causes a 404. 

The URL displays `https://codebar.io/{:controller=>"dashboard", :action=>"wall_of_fame", :only_path=>true, :page=>2}` which seems odd.

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

CodeHex:
I was thinking of looking at this by adding more coaches in the local database, but I haven't used Postgres before and my Rails knowledge is a lttle rusty, so might take a little time. Happy to give it a go if its free?

matyikriszta:
@CodeHex go for it!

CodeHex:
Interesting..., looks like the href for the link is being generated incorrectly e.g. 
```
<a class="next_page" aria-label="Next page" rel="next" href="{:controller=>&quot;admin/feedback&quot;, :action=>&quot;index&quot;, :only_path=>true, :page=>2}">Next →</a>
```

CodeHex:
Its seems to be related to how `will_paginate` is rendering the links. I tested this monkey patch ([original here](https://github.com/mislav/will_paginate/blob/master/lib/will_paginate/view_helpers/action_view.rb#L108-L119)) which fixed it
```ruby
module WillPaginate
  module ActionView
    class LinkRenderer < ViewHelpers::LinkRenderer
      protected

      def url(page)
        @base_url_params ||= begin
          url_params = merge_get_params(default_url_params)
          url_params[:only_path] = true
          merge_optional_params(url_params)
        end

        url_params = @base_url_params.dup
        add_current_page_param(url_params, page)

        #@template.url_for(url_params)  - this line is producing a String version of url_params hash
        Rails.application.routes.url_helpers.url_for(url_params)
      end
    end
  end
end
```
[...]

PR #1930, opened 2023-09-15
Rename clashing method name to fix pagination

CodeHex:
Fixes https://github.com/codebar/planner/issues/1928

Pagination wasn't working as the page urls were being generation via the `full_url_for` mathod in the email helper. This renames the method so it doesn't clash.

Added a test to check a pagination link is generated correctly.

DISCLAIMER: the renaming was done by an auto search and replace. I've looked through the code and the changes make sense to me, but I'm not a Rails expert and I'm not very familiar with the codebase.

If there's a different way you want me to fix this, or tips on any further testing I can do, let me know.

### DIFF
--- a/app/helpers/email_helper.rb
+++ b/app/helpers/email_helper.rb
@@ -1,7 +1,7 @@
 module EmailHelper
   private
 
-  def full_url_for(path)
+  def email_full_url_for(path)
     "#{@host}#{path}"
   end
 end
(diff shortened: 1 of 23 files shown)


matyikriszta:
@CodeHex thanks for your PR and reporting this issue. Ultimately we decided to replace will_paginate with a new gem that supports the Bootstrap CSS framework out of the box as we wanted to make use of those pagination styles. I've opened a PR for this work here: https://github.com/codebar/planner/pull/1997. I will be closing this PR once that one is merged, once again thanks for your work on this.

matyikriszta:
Addressed in #1997

PR #1937, opened 2023-09-26
Fix RSVP button on Event invitation page 

matyikriszta:
### Description
We were getting reports that the RSVP button on the Event invitation page was not working, the link 404ing even though the URL was correct. This PR fixes the link so it loads the page correctly.

### Status
Ready for Review


KimberleyCook (approved):
Nice catch 🎉

PR #1969, opened 2023-11-09
Add controls to homepage Testimonials carousel

matyikriszta:
### Description
This PR adds controls to the homepage Testimonials carousel to increase the accessibility of the section. I've added all the built in controls that Bootstrap provides. Since the carousel autoplays we should also add a play/pause button, this is not provided by Bootstrap and will have to be custom built (but we can make use of methods provided by BS).

### Status
Ready for Review

### Screenshots
| Before | After |
| --- | --- |
| Screenshot 2023-11-08 at 5 26 59 pm | Screenshot 2023-11-08 at 5 19 50 pm |

### DIFF
--- a/app/assets/stylesheets/_bootstrap-custom.scss
+++ b/app/assets/stylesheets/_bootstrap-custom.scss
@@ -6,6 +6,7 @@
  */
 
 $enable-negative-margins: true;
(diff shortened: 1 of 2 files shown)


KimberleyCook (approved):
🎉

PR #1987, opened 2023-12-07
Fix Rollbar warning #521 - ActionController::RoutingError: No route matches [GET] "/images/logo-square.png"

matyikriszta:
### Description
When we upgraded to Rails 7 we moved some assets around and the logo-square.png asset was deleted from public/images, but we forgot to update the reference to this image in that location in the app/views/layouts/_meta.html.haml file. The og:image meta tag is still looking for this image in that location. Update the og:image tag to serve the image from the asset pipeline instead. This fix should get rid of a warning in Rollbar with a large amount of occurrences.

### Status
Ready for Review

### Related Rollbar error/warning
https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/521


KimberleyCook (approved):
🎉

PR #2007, opened 2024-01-02
Improve layout of event & event invitation page

matyikriszta:
### Description
Just a quick PR to clean up the event and event invitation pages. Mostly ensure the event and the event invitation pages present the Venue section the same way for consistency.

### Status
Ready for Review

### Screenshots
**Current event page**
Screenshot 2024-01-02 at 14-25-58 codebar

**Event invitation page**
| Before | After |
| --- | --- |
| Screenshot 2024-01-02 at 14-23-19 codebar | Screenshot 2024-01-02 at 14-22-41 codebar |

### DIFF
--- a/app/views/events/show.html.haml
+++ b/app/views/events/show.html.haml
@@ -1,14 +1,14 @@
 .container.py-4.py-lg-5
   .row
-    .col-12
-      %h1
+    .col-12.col-lg-9
+      %h1.mb-4
         = @event.name
         %br
         %small #{humanize_date(@event.date_and_time, @event.ends_at, with_time: true)}
-    .col-12.col-md-9
-      %p.lead= sanitize(@event.description)
+    .col-12.col-lg-9
+      .lead= sanitize(@event.description)
 
-  .row
+  .row.mt-4
     .col
       = render partial: 'event_actions'
 
@@ -18,7 +18,7 @@
       - if @event.venue.present?
         = render partial: 'shared/venue', locals: { venue: @event.venue, address: @host_address}
       - else
-        %p= "#{t('events.virtual')}."
+        %p= t('events.virtual')
 
     .row.mt-4
       .col
@@ -44,7 +44,7 @@
     .container
       = render partial: 'event_sponsors', locals: { event: @event }
 
-- if @event.verified_coaches.any?
+- if @event.verified_coaches.any? && @event.display_coaches
   .py-4.py-lg-5.section
     .container
       .row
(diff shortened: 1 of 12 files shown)


app/views/events/show.html.haml
```diff
@@ -1,14 +1,14 @@
...
-      %h1
+    .col-12.col-lg-9
+      %h1.mb-4
         = @event.name
         %br
         %small #{humanize_date(@event.date_and_time, @event.ends_at, with_time: true)}
-    .col-12.col-md-9
-      %p.lead= sanitize(@event.description)
+    .col-12.col-lg-9
+      .lead= sanitize(@event.description)
```
KimberleyCook, line 9: Does this not still need the p tag to ensure it's a paragraph?

app/views/invitations/show.html.haml
```diff
@@ -63,23 +63,18 @@
 .py-4.py-lg-5.bg-light
   .container
     .row
-      .col-12.col-lg-6
-        %h3 Schedule
+      - if @event.venue.present?
```
KimberleyCook, line 66: Nice 🎉

app/views/events/show.html.haml
matyikriszta: @KimberleyCook so this is tricky because if you use HTML in the input and it's wrapped in a p tag itself, it outputs a bunch of empty p tags (+ the HTML you put in the input). Maybe make it so that we require the use of HTML in the input?
Screenshot 2024-01-03 at 2 24 33 pm

app/views/invitations/show.html.haml
```diff
@@ -63,23 +63,18 @@
...
+        = render partial: 'shared/venue', locals: { venue: @event.venue, address: @host_address}
+      - else
+        %p= "#{t('events.virtual')}."
```
gnclmorais, line 69: Is the string interpolation here because of the full stop (`.`)? Do we need that dot? Could it just be the following?
```suggestion
        %p= t('events.virtual')
```

matyikriszta: @gnclmorais yeah, we probably do not need that dot or we can move it into the translation string itself

matyikriszta: @gnclmorais I've moved the dot into the translation string

gnclmorais (approved):

app/views/events/show.html.haml
KimberleyCook: Gotcha, makes sense. I'd say make sure all the description field needs markdown so you've done. I get it now :)

KimberleyCook (approved):

PR #2064, opened 2024-05-02
Update codebar's registered address

KimberleyCook:
A PR updating the address mentioned in our Privacy Policy to our new registered address.

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -434,9 +434,9 @@ de:
         address_html: '
           codebar Ltd,<br>
           International House,<br>
-          24 Holborn Viaduct,<br>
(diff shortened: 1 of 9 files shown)


config/locales/de.yml
```diff
@@ -434,9 +434,9 @@ de:
         address_html: '
```
matyikriszta, line 434: @KimberleyCook because the new address has a single quote `'` in 101 King's Cross Road, you now have to update the quotes that wrap the `address_html` entry to double quotes `"`.

KimberleyCook: Oh gotcha, thanks. Should be done now

PR #2076, opened 2024-05-15
Update docker-compose.yml

Hiiiij:
updating the path; Adding Ruby command to run a local server

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
     env_file:
       - docker-compose.env
     build: .
-    command: tail -f /dev/null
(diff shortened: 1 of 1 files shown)


Hiiiij:
Adding Ruby command to.run a local server

biggianteye:
Some of the conversation following this comment is relevant: https://github.com/codebar/planner/pull/1882#issuecomment-2094351062

matyikriszta:
Thanks for this PR @Hiiiij . I'm closing it because we deliberately want to stick to the current approach of starting a container and running a server in the container being two distinct steps. Making developers run a command to start the server lets them see the server logs immediately, while running the server as part of the container startup makes the server logs harder to find.

PR #2083, opened 2024-05-27
Use render collection to for main page

Rukomoynikov:


app/views/dashboard/show.html.haml
```diff
@@ -71,7 +71,7 @@
         %h2.h3.mb-4= t('homepage.events.upcoming')
         - @upcoming_workshops.each do |date, workshops|
           %h3.h5= date
-          = render workshops
+          = render collection: workshops, partial: "events/event"
```
matyikriszta, line 74: @Rukomoynikov can you please explain how this improves the code. Is it that it's easier to read or are there performance implications? The `= render workshops` pattern is used in a bunch of other files, such as app/views/chapter/show.html.haml, app/views/dashboard/dashboard.html.haml and app/views/events/_events.html.haml. I'm not against this change (I think it makes it clearer what partial is being used to render the event cards) but if we are going to update here we should update everywhere else and get rid of the app/views/workshops/_workshop.html.haml partial.

PR #2103, opened 2024-07-29
Fix up Docker config and README instructions so setting up the project via Docker works once again

matyikriszta:
### Description
Installing the application via Docker (the recommended method in the README) hasn't been working for a while. This PR updates the Docker config and the README to ensure installing the project via Docker works once again.

Changes made: 
- Installs chromium into Docker image to run js tests in the container. We prefer this approach to emulating a different architecture because emulating a diff architecture is slow (as per #1900).
- Removes deprecated version key from docker-compose.yml
- Reworks the commands in bin/ so their name reflects the docker-compose commands they call
- Updates the Docker installation instructions in the README, adding the missing step (starting the docker container and running the Rails server are two separate steps now).

### Status
Ready for Review

### Screenshots
N/A

Shout out to @BPScott for a massive assist with this.


README.md
```diff
@@ -61,43 +61,39 @@ from a command prompt in your project folder:
...
+We recommend using Docker to install and run the application. However alternatively if you prefer, [you can install directly to your system environment instead](./native-installation-instructions.md).
 
-#### Install using Docker
+Before you start, you will need to have Docker installed and running. You can [download it from the Docker website](https://docker.com/). Once downloaded, install and start the Docker application.
 
-Before you start, you will need to have Docker installed and running. You can [download it from the Docker website](https://docker.com/).
+Run `bin/dup` to build and create the docker container. This will also set up the Rails application within the container and start the container. You will only have to run this command once. After initial setup, use `bin/dstart` to start an existing container - using `bin/dup` will recreate an existing container and reset the database.
 
-The current Dockerfile and docker-compose were closely copied from [the Docker guide](https://docs.docker.com/compose/rails/).
```
matyikriszta, line 72: Removed this line because the linked page 404s

bin/dadmin
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env bash
 set -e
 
-docker-compose up -d web
```
matyikriszta, line 4: Removed because the container should already exist when running this command

bin/dexec
```diff
@@ -2,5 +2,4 @@
 
 set -e
 
-docker-compose up -d web
```
matyikriszta, line 5: Removed because the container should already exist when running this command

bin/drake
```diff
@@ -2,18 +2,4 @@
 
 set -e
 
-function cleanup {
```
matyikriszta, line 5: We do not want to remove the container every time we run a rake command

bin/drails
```diff
@@ -2,4 +2,4 @@
 
 set -e
 
-docker-compose run --rm web rails "$@"
```
matyikriszta, line 5: We do not want to remove the container every time we run a rails command

matyikriszta:
@biggianteye would you be able to review this PR for us?

MichalSmyk:
bin/dup throws `bin/dup: line 5: docker-compose: command not found`

biggianteye:
> @biggianteye would you be able to review this PR for us?

Yep. Will do.

biggianteye:
On my Intel Mac, following the README, I was able to successfully do the following:
- Build the docker image and set up the application.
- Have a running web application.
- Run the tests.

On my M1 Mac I wasn't able to finish installing the Ruby dependencies, but I think that's still a separate issue that's unrelated to this PR.

In my eyes, this PR does what it intends by making sure the Docker instructions are up to date.

MichalSmyk:
> > bin/dup throws `bin/dup: line 5: docker-compose: command not found`
> 
> It looks like `docker-compose` is no longer provided as a separate command out of the box. It can be installed as a separate package. Though the functionality has been built into the main `docker` binary for a while now (via `docker compose`).

Not sure what the issue was, I don't really work with docker on daily basis. I did `docker compose up` first and then I was able to do bin/dserver and run the app.

matyikriszta:
@biggianteye updated `docker-command` to `docker command` everywhere we use it. @MichalSmyk this should fix your issue as well. If you do `bin/ddown` (to destroy the container) and then do `bin/dup` again it should run without issue. Let me know if it works.

MichalSmyk:
@matyikriszta looks good now, no issues. Thanks for sorting it out

README.md
```diff
@@ -61,43 +61,39 @@ from a command prompt in your project folder:
...
+### 4. Start the app
 
-1. Run `bin/dbuild` to build and setup the docker environment.
-2. Run `bin/drake` to run all the tests and make sure everything works.
+Run `bin/dserver` to start the Rails server.
 
-**Note:** If you are using Windows, you can run `bin/dbuild` using Git Bash.
+*If you have previously stopped the container, you will have to first start the container using `bin/dstart`.
```
KimberleyCook, line 76: `bin/dstart` didn't seem to work for me to get the container up and running again after closing it down. I was able to do `bin/dserver` to get it working again

matyikriszta: @KimberleyCook do you remember the exact steps you took? You ran `bin/dstop` and then `bin/start`?

BPScott: Might be worth noting that "running the container" is not the same thing as "running the rails http server in the container".

`dstart` will start the container - but the rails server won't be running
`dserver` will start the rails https server once a container is running

matyikriszta: @KimberleyCook I'm thinking this might have been your issue. I'm going to merge the PR for now, since you got it working eventually. But yes, running a container and running the rails server are two different things.

PR #2120, opened 2024-10-04
Lots of copy changes

KimberleyCook:
I also moved the placement of the donate button and made the font size 16px (and semi-bold) as it was 14 and the rest of the nav so didn't match up.

Screenshot 2024-10-04 at 14 08 21

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -89,7 +89,7 @@ en:
       students_explanation_p2: "Join our vibrant community and take part in our workshops!"
       students_button: "Join us as a student"
       coaches_title: "Coaches"
-      coaches_explanation_p1: "We encourage new students to start with our <a href='http://tutorials.codebar.io'>tutorials</a> on HTML/CSS, JavaScript, Ruby, Python, or Git, while also helping them grasp key programming concepts."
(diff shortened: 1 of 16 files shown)


config/locales/en.yml
```diff
@@ -340,12 +340,10 @@ en:
     sign_up_as_coach: "Join us as a coach"
     code_of_conduct: "code of conduct"
     new:
-      intro: "We offer a welcoming, inclusive and learning-friendly environment, and have a zero tolerance policy towards any form of harassment or inappropriate behavior. Before you sign up, read our"
+      intro_html: "We provide a welcoming, inclusive, and supportive learning environment, with a strict zero-tolerance policy for any form of harassment or inappropriate behavior. Please take a moment to review our <a href='https://codebar.io/code-of-conduct'>code of conduct</a> before signing up."
```
matyikriszta, line 343: Instead of hard coding the href of the links like this can we pass it in as a variable to the translation string? That way we can make use of the path helpers (where they are available) and if the path every changes we'll only have to update it in one place. https://guides.rubyonrails.org/i18n.html#passing-variables-to-translations

matyikriszta (approved):

PR #2127, opened 2024-10-25
Handle student & coach event emailing as separate

gnclmorais:
Send them both in the background, separately. This should help with the timeouts we have sometimes on Heroku, I hope.

### DIFF
--- a/app/models/invitation_manager.rb
+++ b/app/models/invitation_manager.rb
@@ -33,6 +33,7 @@ class InvitationManager
       EventInvitationMailer.invite_student(event, student, invitation).deliver_now if invitation.save
     end
   end
+  handle_asynchronously :invite_students_to_event
 
   def invite_coaches_to_event(event, chapter)
     chapter_coaches(chapter).each do |coach|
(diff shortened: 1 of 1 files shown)


matyikriszta:
@gnclmorais we currently can't set up review apps, we'd have to configure that feature on Heroku, we've never done so. we could test on staging, I do wonder if we have enough seed data to meaningfully test this feature, we'd have to have thousands of members subscribed to a chapter to really be able to test it. I think this should help already, one other thing we could do is set the coach emails to `deliver_later`. Not sure how this works together with the `handle_asynchronously` method. Coach spots do not fill up as fast as student spots so I think it's not a big issue if everyone doesn't get the email at about the same time.

till:
@gnclmorais @matyikriszta — hey 👋🏼 — we doing some housekeeping and are gonna close this PR as it's a bit stale. 

We also haven't seen timeouts recently. 🥳 

I think for the future of email sending, we should explore the idea of using a template and batch sending via Sendgrid/Twillio in order to make it more resilient. Then no async is required here as we can submit up to ~1000 emails in one request.

PR #2155, opened 2024-12-20
Dietary restrictions

### LINKED ISSUE #1842: Add Dietary Requirements on profiles
## Description of the issue 📄
Hosts sometimes need dietary requirements from us, it'll be much easier if we could get the requirements from the profiles of our attendees rather than asking on the day of events.

When someone is signing up for codebar ask them their dietary requirements, and then show it on the persons profile.

## To do 📋

* [ ] Submit a pull request

jonodrew:
image

This change should:
- generate a database migration that adds a 'dietary_restrictions' column
- add a field to the edit Member page

this should close #1842

### DIFF
--- a/app/views/member/details/edit.html.haml
+++ b/app/views/member/details/edit.html.haml
@@ -11,6 +11,7 @@
         = f.input :surname, required: true
         = f.input :pronouns
         = f.input :email, required: true
+        = f.input :dietary_restrictions, label: 'If you have dietary restrictions, please add them here'
         - if @member.coach?
           = f.input :about_you, as: :text, label: t('member.details.edit.coach.about_you'), input_html: { rows: 3 }, required: true
         - else
--- a/db/migrate/20241220215044_add_dietary_restrictions_to_member.rb
+++ b/db/migrate/20241220215044_add_dietary_restrictions_to_member.rb
@@ -1,0 +1,5 @@
+class AddDietaryRestrictionsToMember < ActiveRecord::Migration[7.0]
+  def change
+    add_column :members, :dietary_restrictions, :string
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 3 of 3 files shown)


app/views/member/details/edit.html.haml
```diff
@@ -11,6 +11,7 @@
         = f.input :surname, required: true
         = f.input :pronouns
         = f.input :email, required: true
+        = f.input :dietary_restrictions, label: 'If you have dietary restrictions, please add them here'
```
biggianteye, line 14: The label should probably use the translation mechanism rather than a raw English string, to allow it to be translated for non-English chapters. eg.
```suggestion
        = f.input :dietary_restrictions, label: t('member.details.edit.dietary_restrictions')
```

With a corresponding entry in `config/locales/en.yml`.

Someone more familiar with Ruby on Rails might be able to give you more specific guidance.

jonodrew: Good shout, done that!

jonodrew:
I can see there's a failing test here, but I don't think there's anything I can do about it - unless there are some credentials I'm missing? I can't add `GITHUB_KEY` as a secret, because Github doesn't allow secrets that start with 'GITHUB'

mikej:
Hi @KimberleyCook I'm taking a look into picking this one up from @jonodrew 

Were you envisioning a set of checkboxes so e.g. someone who is vegetarian and gluten free can pick both, something like this:

New Wireframe 1 (Dietary restrictons - checkboxes)

Or a dropdown that allows a single selection, and then use "Other" if multiple apply, so more like:

New Wireframe 1 (Dietary restrictions dropdown)

Whichever approach, I'm thinking of getting this onto the "Update your details" form as an initial release and then looking at making it available on the Attend page as a way to encourage more existing members to provide the information.

KimberleyCook:
@mikej I think the first checkbox options works best :)

mikej:
> I think the first checkbox options works best :)

Thanks for getting back to me @KimberleyCook, will go with that approach.

Hope you're doing ok 😊

olleolleolle:
@jonodrew Hi! Is this superseded?

jonodrew:
Yup!

PR #2156, opened 2024-12-22
Fix link to sponsor #2512 

Zelda-Cat:
Fixes link to sponsor website from admin area, resolves #2512 by using name as param in link_to, alternatively I could set only_path: false. Also opens external link in new tab


matyikriszta (approved):
Thanks for the fix @Zelda-Cat

PR #2218, opened 2025-07-11
Improve repository configuration

mroderick:
Reviewers are encouraged to read the individual commits to maintain their sanity.

- **chore: group minor and patch level**
- **chore: group github action updates into one pr**
- **chore: remove extra ruby version from workflow**


mroderick:
@matyikriszta I think all the checks we care about are passing. Do you have powers to merge this?

PR #2220, opened 2025-07-16
CI: ruby/setup-ruby reads .ruby-version

olleolleolle:
This PR relies on ruby/setup-ruby to pick good defaults for installing packages.

And, we don't want to write the version number more than once (in the .ruby-version file).


olleolleolle:
@matyikriszta Here! It passed!

matyikriszta (approved):

PR #2226, opened 2025-07-19
Aws s3

jonodrew:
This is a longer-term plan, but evidences that it is low-complexity to move to S3 for image storage.

Before this is merged, we have to:

- [x] create an S3 bucket
- [x] create an IAM role with S3 access
- [x] download the credentials for that role
- [ ] set them in the production environment

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -231,7 +231,7 @@ GEM
     method_source (1.1.0)
     mini_magick (4.12.0)
     mini_mime (1.1.5)
-    mini_portile2 (2.8.8)
+    mini_portile2 (2.8.9)
     minitest (5.25.5)
     msgpack (1.7.5)
     multi_xml (0.6.0)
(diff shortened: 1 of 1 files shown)


jonodrew:
If we implement this:
- all admins will need to re-upload sponsor logos
- we will need an AWS account with an IAM role to access the bucket

jonodrew:
Screenshot, just to show it's working:
image

jonodrew:
Remove the FTP gem

app/uploaders/avatar_uploader.rb
```diff
@@ -1,14 +1,18 @@
...
 
+  def content_type_allowlist
+    [/image\//]
+  end
+
   # Override the directory where uploaded files will be stored.
   # This is a sensible default for uploaders that are meant to be mounted:
   def store_dir
-    "uploads/#{model.class.to_s.underscore}/#{mounted_as}/#{model.id}"
+    "uploads/#{model.class.to_s.underscore}/#{model.id}"
```
matyikriszta, line 15: @jonodrew I don't really mind either way but any reason we removed the mounted_as from the directory?

PR #2238, opened 2025-07-31
Patch tests

jonodrew:
A couple of tweaks to the test suite with the aim of making the tests a little bit less flakey. They pass locally, because of course they do. This PR will demonstrate whether or not it works (I hope!)


olleolleolle (approved):
Looks reasonable!

PR #2244, opened 2025-08-05
chore: stop pinning image_processing

mroderick:
This is completely unnecessary and is not helping anyone


olleolleolle (approved):
Awesome!

PR #2245, opened 2025-08-05
chore: remove unused New Relic config

mroderick:
This was added in 0650a721c2d358ed1dec908fc579e5847603f8c6, but the instrumentation was never addedd and in both staging and production heroku, we can access completely empty New Relic accounts.

Once this has been merged, we can remove New Relic in Heroku `staging` and `production`.


olleolleolle (approved):
This cleans up!

PR #2251, opened 2025-08-06
chore: add .editorconfig

mroderick:
In some PRs there are missing newlines at end of files, because we have inconsistent editor settings. This fixes that


olleolleolle (approved):
Thanks!

PR #2254, opened 2025-08-06
fix: failing integration tests

davidmillen50:
The failing test in `member_feedback_spec.rb` was due to the test not waiting for the url to change.  This was fixed by using `have_current_path` to wait for the url to change.  

The test in `manage_workshop_attendances_spec.rb` was failing because the RSVP member dropdown was appearing in front of the h4 title 'Students' and as a result the dropdown was not being clicked.  This was fixed by checking the dropdown is visible before clicking.  If the element click error is raised a custom message is logged.


mroderick (approved):
Nice fixes!

PR #2260, opened 2025-08-09
Member search

jonodrew:
This adds a controller and little view for doing member searches. Full implementation of this pattern will require calling services to add a `callback` param to the request to `MemberSearch`, which it will then `GET` once the Member(s) have been found.

If only one member is found from the search, I've chosen to assume that it's correct and we skip straight to redirecting to the calling service. That might be a faulty assumption.

I'm going to clear up the commit history, but it'll give you an early sense of how this is working. Screenshots!

Screenshot 2025-08-09 at 22 14 09

Screenshot 2025-08-09 at 22 14 29

### DIFF
--- a/app/controllers/admin/member_search_controller.rb
+++ b/app/controllers/admin/member_search_controller.rb
@@ -2,7 +2,7 @@ class Admin::MemberSearchController < Admin::ApplicationController
   def index
     member_params = params[:member_search] || {}
     name = member_params[:name]
-    members = name.blank? ? Member.none : Member.find_members(name).select(:id, :name, :surname, :pronouns)
+    members = name.blank? ? Member.none : Member.find_members_by_name(name).select(:id, :name, :surname, :pronouns)
     callback_url = member_params[:callback_url] || params[:callback_url] || results_admin_member_search_index_path
     if members.size == 1
       query = { member_pick: { members: [members.first.id] } }
--- a/app/controllers/admin/member_search_controller.rb
+++ b/app/controllers/admin/member_search_controller.rb
@@ -2,20 +2,20 @@ class Admin::MemberSearchController < Admin::ApplicationController
   def index
     member_params = params[:member_search] || {}
     name = member_params[:name]
-    @members = name.blank? ? Member.none : Member.find_members(name).select(:id, :name, :surname, :pronouns)
-    @callback_url = member_params[:callback_url] || params[:callback_url] || results_admin_member_search_index_path
-    if @members.size == 1
-      query = { member_pick: { members: [@members.first.id] } }
+    members = name.blank? ? Member.none : Member.find_members(name).select(:id, :name, :surname, :pronouns)
+    callback_url = member_params[:callback_url] || params[:callback_url] || results_admin_member_search_index_path
+    if members.size == 1
+      query = { member_pick: { members: [members.first.id] } }
       query_string = query.to_query
-      callback_url = "#{@callback_url}?#{query_string}"
+      callback_url = "#{callback_url}?#{query_string}"
       redirect_to callback_url and return
     end
 
-    render 'index', locals: { members: @members, callback_url: @callback_url }
+    render 'index', locals: { members: members, callback_url: callback_url }
   end
 
   def results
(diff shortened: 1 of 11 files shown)


app/views/admin/member_search/_search_form.html.haml
```diff
@@ -0,0 +1,6 @@
+= simple_form_for :member_search, url: admin_member_search_index_path, method: :get, html: {:multipart => true, novalidate: true } do |f|
```
olleolleolle, line 1: Here we can uniformly use colon instead of hashrocket.

config/routes.rb
```diff
@@ -140,6 +140,13 @@
     end
 
     resources :testimonials, only: %i[index]
+
+    resources :member_search, :path => 'member-search', only: [:index, :results] do
```
olleolleolle, line 144: Prefer colon over hashrocket.

jonodrew:
@matyikriszta do you mind giving me a sense-check on this? Does it look like it's going in the right direction?

app/views/admin/member_search/index.html.haml
```diff
@@ -0,0 +1,23 @@
...
+    .col
+      %h1 Member search
+
+  .row
+    .col.col-md-10.col-lg-8
+      = render partial: 'search_form'
+
+  .row 
+    .col.col-md-10.col-lg-8
+      - if @members.present?
```
olleolleolle, line 12: We pass in named locals in the controller to this view. And, all views and partials.

Here, we should use that `members` local variable, not the `@members` instance variable.

The benefit is that each partial or view is told explicitly which variables it should have access to. And the errors the developer sees are NameErrors, not NoMethodError on nil (an @instance variable is nil if unset).

app/controllers/admin/member_search_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @members = @name.blank? ? Member.none : Member.find_members(@name).select(:id, :name, :surname, :pronouns)
+    @callback_url = @params[:callback] || results_admin_member_search_index_path
+    if @members.size == 1
+      query = { member_pick: { members: [@members.first.id] } }
+      query_string = query.to_query
+      @callback_url = "#{@callback_url}?#{query_string}"
+      redirect_to @callback_url and return
+    end
+
+    render 'index', locals: { members: @members, callback: @callback_url }
```
olleolleolle, line 14:
```suggestion
    render 'index', locals: { members: @members, callback_url: @callback_url }
```

olleolleolle: So, the @callback_url can also become a local variable, not an instance variable.

olleolleolle (commented):
Just wanted to share some Rails thoughts.

app/views/admin/member_search/index.html.haml
olleolleolle: So, we need to pass in the local variable name `members` to this view, from the controller action #index.

Long story short: either one uses instance variables all over, OR locals all over.

jonodrew:
@olleolleolle @matyikriszta - I'd like to integrate this code into the codebase before I start working on calling it from elsewhere. I suspect the approach is going to be a little complex, and I don't want to make a massive change all at once.

Tests are passing locally but not here - there's more flakiness around I think.

Going forward, I'm thinking about how this can be incorporated into other calling services. I'm considering moving the logic in `Admin::InvitationsController#update_attendance` into a method on the `Invitation`  model, so I can access it from within any Controller. Let me know what you think.

spec/models/member_spec.rb
```diff
@@ -137,4 +137,33 @@
       expect(member.flag_to_organisers?).to be true
     end
   end
+
+  describe '#find_members' do
+    describe 'search by first name' do
+      it "should find the member" do
```
olleolleolle, line 143: The way the other tests (just above this one) express themselves is more confident: `returns the member`. I feel that is clearer, less "perhaps".

```suggestion
      it "returns the member" do
```

jonodrew: This rebase was trickier than expected, but I think it's fixed now!

PR #2261, opened 2025-08-10
Fix: tests failing due to incorrect capybara driver assignment

davidmillen50:
The problem
This PR fixes an issue where tests fail intermittently due to the capybara driver staying as chrome after a js: true test even when the following tests are not javascript tests and should therefore be using rack_test.  When the non js tests run with the chrome driver test failures occur with `Selenium::WebDriver::Error::InvalidSessionIdError: invalid session id`.  If the driver used for non js tests is `rack_test` these failures don't occur and the tests pass. 

The solution
This PR creates an around block and uses driven_by to set the capybara for system tests and current_driver for feature tests.  This is done in an around block to be thread safe as before and after test blocks as changes can leak across threads if there isn't proper isolation.  An around block ensure thread safety. System specs use driven_by which uses Rails built in capybara driver management and would use :selenium instead of chrome.  This fix should stop rails from deciding on the driver as the driver is now explicitly set for each test depending on whether the test is JS or not and whether the test type is system or feature.  I've included a test file which should enforce this driver consistency and stop regressions

### DIFF
--- a/spec/capybara_driver_isolation_spec.rb
+++ b/spec/capybara_driver_isolation_spec.rb
@@ -1,0 +1,46 @@
+require 'spec_helper'
+
+RSpec.describe "Capybara driver isolation in spec types feature, system and neither", order: :defined do
+  context "feature spec", type: :feature do
+    it "uses default driver (rack_test) for non JavaScript test" do
+      expect(Capybara.current_driver).to eq(Capybara.default_driver)
+    end
+
+    it "uses JavaScript driver (chrome) when test is js: true", js: true do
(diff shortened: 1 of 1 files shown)


davidmillen50:
these errors happen intermittently and as a result it's not possible to test this problem consistently.  I tried creating a capybara driver test file but the tests in that file pass even though the session id errors still occur.  The different approaches I've tried to fix this haven't worked and I can put this down to it being a local issue or occasionally the capybara session just crashes and you have to rerun the tests.

PR #2267, opened 2025-08-12
chore: use associations for finding workshop sponsors with hosts

davidmillen50:
I have implemented workshop associations' hosts to be returned by the host method on the Workshop model and I've removed the scopes to do the same on the workshop_sponsor model


olleolleolle (approved):
I assume this gets the amount of queries down.

PR #2268, opened 2025-08-12
chore: infer RSpec metadata about "type" of test from file location

olleolleolle:
This quality-of-life PR configures RSpec's inference of types of test by where they are placed in the filesystem. We follow all the regular conventions, so it ought to just work.

How it works:

- RSpec Rails (the gem) finds files in conventional directories
- It automatically adds `type: :model` for example, for files in spec/models
- That way, all the metadata that we use to add type-specific behavior is kept, but out of the way

### DIFF
--- a/spec/controllers/auth_services_controller_spec.rb
+++ b/spec/controllers/auth_services_controller_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe AuthServicesController, type: :controller do
+RSpec.describe AuthServicesController do
   describe "GET #new" do
     it "redirects when referer is missing" do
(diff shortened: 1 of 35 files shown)


spec/controllers/admin/invitations_controller_spec.rb
```diff
@@ -1,5 +1,3 @@
-require "spec_helper"
-
 RSpec.describe Admin::InvitationsController, type: :controller do
```
matyikriszta, line 1: @olleolleolle should we remove the type here as well?

PR #2269, opened 2025-08-12
spec helper: Remove removed keyword argument full: true from #save_screenshot

jonodrew:
As far as I can tell, `#save_screenshot` doesn't accept the `:full` option. This means that when a js test fails, it fails twice.

This should ensure it only fails once.


jonodrew:
@olleolleolle quick one!

PR #2270, opened 2025-08-13
chore: Remove unused gosquard

mroderick:
This was introduced in 34de45fa4c2846abab0c8907ef343688791b2043, but has not been used for several years.

Removing this will make the JS bundle smaller, improving the user experience.

This was [discussed with @matyikriszta in the codebar Slack](https://codebar.slack.com/archives/C0970A68344/p1754467315999379)


olleolleolle (approved):
Lovely!

PR #2271, opened 2025-08-13
chore: Remove unused (and deprecated) Google Analytics

mroderick:
This removes Google Analytics, which was introduced in 76c1537364fe9c7d6f557c6cce4ef0d6901df38e

> For a while we also had GA and we used that more I think. Not sure
> what happened to it (e.g. we never migrated to GA4 AFAIK), we don’t
> look at website analytics that much.

As [discussed with @matyikriszta in codebar Slack](https://codebar.slack.com/archives/C0970A68344/p1755019509066129?thread_ts=1754467315.999379&cid=C0970A68344)

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -22,7 +22,6 @@
 //= require pickadate/picker
 //= require pickadate/picker.date
 //= require pickadate/picker.time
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Excellent!

PR #2277, opened 2025-08-13
chore: remove 2024 impact report banner

mroderick:
I'm working on a PR to improve the accessibility of the website, and there was a warning about the banner.

We're more than half way through 2024, 2025 is coming up soon :)

So, instead of putting effort into fixing this banner, let's not display it until we have the next thing to draw attention to.

#### Before

image

#### After

image

### DIFF
--- a/app/views/dashboard/show.html.haml
+++ b/app/views/dashboard/show.html.haml
@@ -1,7 +1,7 @@
 - title t('homepage.title')
 
-.alert.alert-dark.mb-0.text-center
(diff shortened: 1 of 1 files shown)


olleolleolle:
cc @matyikriszta - we moved quickly, and I hope this change aligns with your assessment of the situation.

jonodrew:
cc. @KimberleyCook as Director also

KimberleyCook:
Sorry only just seeing this, can we revert the PR please the banner should be kept in all year as we're always wanting to promote it :)

olleolleolle:
Revert coming! (But first, a strong Arabic lunch - Malmö is super at that.)

jonodrew:
Reversion: https://github.com/codebar/planner/pull/2284

PR #2285, opened 2025-08-14
CI: Do not fail the build if Coveralls fails

olleolleolle:
We had many build failures with "oh, the code coverage was not sent to that service", which is a little "them problem" rather than an "us problem", so I configured this to not be considered a failure.

To think about for next week's meeting: Do we wish to keep sending these coverage reports? How do we want to use them?

Screenshot of today's Coveralls:

image

With this change, the Checks look like this, **non-failing** (the icon for the "Report to Coveralls" section is a circle with a checkmark in it), even though that non-essential service is down:

image

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -43,6 +43,7 @@ jobs:
         RAILS_ENV: test
       run: bundle exec rake spec
     - name: Report to Coveralls
+      continue-on-error: true # Don't fail the build if Coveralls fails to upload.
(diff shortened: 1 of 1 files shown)


matyikriszta (approved):
I think it's useful to keep some sort of coverage reporting for the repo. Coveralls has had a lot of issues recently, maybe we should think about migrating to another service? [Qlty by Code Climate](https://qlty.sh/) has coverage reporting too. That said, this is not our main priority at the moment.

PR #2287, opened 2025-08-17
fix: flaky test

davidmillen50:
Fix attendees emails test which was trying to match two strings together when order can change and cause flakiness.  Resolved by expecting string to include emails so order shouldn't matter


KimberleyCook (approved):
🎉

PR #2288, opened 2025-08-17
Fix: bug when you try to add empty organiser to chapter

davidmillen50:
I added validation in the organisers controller to show a message if user tried to click add organiser without selecting one from the dropdown


KimberleyCook (approved):
Nice 🎉

PR #2293, opened 2025-08-19
chore: initialise docs folder

mroderick:
This PR adds a `docs/` folder, where we will capture what we learn about this application.


olleolleolle (approved):
Good idea!

We can add architecture explainers, we can add [ADRs ](https://adr.github.io), all kinds of "repo reader-facing documentation".

Issue #2296, opened 2025-08-20
New announcement endpoint allows for announcing to entire membership


jonodrew:
The endpoint `/admin/announcements/new` has no limits on which chapters you can contact. We should reduce that to only those that an admin has responsibility for, ie their chapter

davidmillen50:
If an admin has to make an announcement to all chapters they have to undo this change?

biggianteye:
Looking at the [list of announcements](https://codebar.io/admin/announcements/) made under `/admin/announcements`, it looks like they are infrequent and have only been done by senior codebar people (Kim, Despo, Kriszta). Given that the page allows communicating with all chapters, the combination of these two things makes me think that this page is only intended for admin users rather than for organisers to communicate with their chapter. So I suspect that the page as a whole should be restricted to those with "admin" permission rather than keeping the current permission level and filtering for chapters an organiser is in.

@KimberleyCook The newest announcement is by you in December 2024 (CFP for Codebar Festival). Are you able to give us guidance on the intended users of the announcements page?

matyikriszta:
@jonodrew @biggianteye is correct, this functionality is only intended to be used by admins. It was added so we can communicate codebar wide announcements to the whole community (aka everyone across all geographical regions) or a subset of chapters (e.g. all UK chapters) or even just a couple of chapters. Please do leave it as it is for now (aka only admins can access and set announcements). If this is something organisers would find useful I'm happy to explore how we could open this functionality up to them, allowing them to post messages to their own chapter. It is however not a priority for us at the moment.

PR #2297, opened 2025-08-20
Limit access to announcements

jonodrew:
Quick fix for #2296 

This updates the base controller for the AnnouncementController so that only admins, not admins + organisers, can access the announcements endpoint


olleolleolle:
The query seems a little off:

```
     ActiveRecord::StatementInvalid:
       PG::UndefinedColumn: ERROR:  column chapters.chapter_id does not exist
       LINE 1: ..."."chapter_id" WHERE "chapters"."active" = $1 AND "chapters"...
```

mroderick:
Looking at the discussion in #2296, I think this PR can be replaced with one that adds a requirement for `admin` role, and adds documentation into `docs/` :)

PR #2299, opened 2025-08-22
fix: Improve introduction language

mroderick:
> Our workshops and events are exclusively for women, non-binary, LGBTQ+
> members, and people from underrepresented **minority** groups (emphasis mine)

This might seem like nitpicking, but women are not a minority group. Whenever we divide up the general population along gender axis, women will always be the majority group.

However, women are underrepresented in programming roles, along with people from other underrepresented groups.

This changeset improves the copy on the website a bit.


mroderick:
Ping @KimberleyCook: are you into these kinds of improvements to the website?

PR #2305, opened 2025-08-24
Identify slow requests

jonodrew:
I'm not completely happy with this yet.

I've written some helper methods to grab all of the 'things' a member is attending and find them once, when the controller method is called. This replaces the map loop that generated two queries per event.

However, I think there's some clash here and the tests will therefore be flakey. Advice welcome


jonodrew:
In total, this cuts loading the front page down to below 4s

app/presenters/member_presenter.rb
```diff
@@ -23,6 +39,8 @@ def pairing_details_array(role, tutorial, note)
     role.eql?('Coach') ? coach_pairing_details(note) : student_pairing_details(tutorial, note)
   end
 
+
```
davidmillen50, line 42: are these two spaces needed?

app/views/events/_event.html.haml
```diff
@@ -5,10 +5,11 @@
         - if event.chapter
           %span.badge.bg-primary.mb-3.mb-md-0
             = link_to event.chapter.name, event.chapter.slug, class: 'text-light text-decoration-none' 
-        - if @user
-          - if @user.attending?(event.__getobj__)
+        - if !attending_ids.blank?
```
davidmillen50, line 8: `attending_ids.present?` might be better here

app/views/events/index.html.haml
```diff
@@ -5,11 +5,11 @@
     .col
       - if @events.any?
         %h3.mb-4 Upcoming Events
-  = render partial: 'events', locals: { grouped_events: @events }
+  = render partial: 'events', locals: { grouped_events: @events, attending_ids: @attending_ids}
```
davidmillen50, line 8: I think you need a space before the closing `}`

app/controllers/concerns/attendance_concerns.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+module AttendanceConcerns
+  extend ActiveSupport::Concern
+
+  def attending_workshops
+    current_user.nil? ? Set.new : current_user.workshop_invitations.accepted.pluck(:id).to_set
```
olleolleolle, line 7: Micro detail: the `ids` method - https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-ids

PR #2308, opened 2025-08-25
chore: remove skylight

### LINKED ISSUE #2307: Remove skylight
In [https://github.com/codebar/planner/discussions/2292] we learned that Skylight is currently unused and that we don't have an account for it.

So, we can remove it for now and reduce the amount of warnings in the logs and improve the startup time and memory load of the application.

mroderick:
Fixes #2307

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -117,6 +117,5 @@ group :production do
 end
 
 gem 'rollbar'
-gem 'skylight'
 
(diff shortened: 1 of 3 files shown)


KimberleyCook (approved):
Nice tidy up, thank you so much for all your work on the website recently :D

PR #2312, opened 2025-08-26
Add support for dietary restrictions on member profiles

mikej:
Add the ability for members to specify any dietary restrictions on their profile and then display these on the workshop admin page in the list of members who are attending.

The main highlights are:

 * Allow a member to specify any dietary restrictions a) during signup and b) via "Update your details"
 * Dietary restrictions are from a fixed, built-in list with the option to also tick "Other" and then enter free text in the "other dietary restrictions field"
  * Dietary restrictions are displayed a) on a member's own profile b) when an admin views a profile c) next to each of the RSVPs on a workshop admin page
  * Invite page text updated to encourage members to use "Update your details" to specify any dietary restrictions instead of emailing them to the chapter email address as before


app/views/invitation/_coach.html.haml
```diff
@@ -12,4 +12,4 @@
   - if @workshop.virtual?
     = t('workshop.virtual.invitation.shared_intro_3_html', email_link: mail_to(@workshop.chapter.email, @workshop.chapter.email))
   - else
-    %p= t('workshop.invitation.shared_intro_4_html', email_link: mail_to(@workshop.chapter.email, @workshop.chapter.email).html_safe)
+    %p= t('workshop.invitation.shared_intro_4_html', update_your_details_link: link_to(t('members.update_your_details_lower'), edit_member_path).html_safe)
```
davidmillen50, line 15: Are these changes needed for this PR? If they are an additional improvement I'd say extract them into a separate PR

config/locales/en.yml
```diff
@@ -671,6 +671,7 @@ en:
         name: 'Grace'
         surname: 'Hopper'
         pronouns: 'e.g she/her/they'
+        skill_list: 'e.g. HTML, CSS, JavaScript, React, Ruby, Python, SQL'
```
davidmillen50, line 674: same as above

davidmillen50:
Nice! very good, comprehensive solution for this feature.  Maybe the improvements that don't relate to dietary restrictions could be moved into a separate PR to reduce the changes here and make reviewing a little bit easier.

As you probably know my PR for the 'how you found us' section will create conflicts with this PR depending on which is merged first so I suppose we have to be mindful of that.  

You added a comprehensive solution for ticking 'other' and having the text input field appear conditionally whereas my implementation for 'how you found us' is simpler with a static input field so the user experience won't be consistent across these two parts of onboarding and updating their profile.  I don't think this is a major issue but it raises the question of whether we want them to be consistent or not.

app/views/invitation/_coach.html.haml
mikej: Thanks for looking, David 🙏🏻 This change is related to this PR because as part of introducing the dietary restriction settings we can now encourage members to update their profile themselves rather than sending an email to the chapter email address with any dietary restrictions they have 🙂

jonodrew:
Couple of questions here from @mroderick and a conflict on locales before I can give it a tick @mikej!

PR #2314, opened 2025-08-26
chore: ignore unneccessary install messages

mroderick:
In collaboration with @olleolleolle, we are removing a bit of junk mail during install in CI


olleolleolle (approved):
We also configured these ENV vars on Heroku (staging, production).

PR #2316, opened 2025-08-27
Make event rendering faster

mroderick:
This PR makes event rendering slightly faster, by memoising `MemberPresenter#organiser?` and using it to prevent unnecessary db queries (`@user.event_organiser?(event)`)

This PR overlaps slightly with @jonodrew's work in https://github.com/codebar/planner/pull/2305, which will make this much faster.

#### Before

image

#### After

image


mroderick:
@olleolleolle @jonodrew could either of you merge this?

PR #2319, opened 2025-08-27
chore: remove unused i18n feature

mroderick:
The files were added in
- https://github.com/codebar/planner/pull/647
- https://github.com/codebar/planner/pull/813

But has not been put to use.

Once we're at Rails 8, we can re-visit if we want to implement
localisation. I'm sure things have moved forwards since this was
introduced in 2018.


matyikriszta:
@mroderick is this getting in the way of the upgrade to Rails v8? Why are we trying to remove it? These were definitely used for a while, I forgot now in what context (e.g. which chapter used it but there was efforts to translate the site to Spanish and German for sure). @KimberleyCook what do you think? I say instead of deleting it we need to focus on improving it.

PR #2326, opened 2025-08-29
chore: Use spaces, not tabs, in HAML file

olleolleolle:
This was the one file with wrong formatting like this that I found when using haml-lint.


mroderick:
I reviewed the `.editorconfig`, all ok. That should help prevent this in the future (provided people use it).

mroderick (approved):

PR #2328, opened 2025-08-29
Issue 1827 add how you found us to user profile

davidmillen50:
I've added a section for 'How you found us' to the profile page.  This section is not yet editable from the profile page.  If that's needed I'll create a separate PR. 

Screenshot 2025-08-29 at 14 04 55


KimberleyCook:
@davidmillen50 when do you think this will be ready for review? I’m really keen to start knowing this information as soon as possible :)

davidmillen50:
@KimberleyCook I have updated the other PR and it is now up for code review

PR #2331, opened 2025-09-01
feat: add option to seed more workshops

davidmillen50:
To have more workshops seeded in local planner database I've added the option of seeding high volumes of workshops or fewer based on the `SEED_HIGH_VOLUME_OF_WORKSHOPS` env variable. When set to `false` the number of workshops seeded will be the historical value of 6 compared to when set to true when 10,000 workshops will be seeded.  

I’ve also added the option to seed without logging each record creation so that seeding can take place faster.  If you set the environment variable `SILENT_SEED` to true then the logging will be hidden.  When both env variables are not set in the env file the seeding will occur as it has in the past with fewer workshops and with the creation of records being logged.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,97 +1,106 @@
-if Rails.env.development?
-  begin
-    Rails.logger.info 'Running migrations...'
-    Rails.application.config.log_level = :info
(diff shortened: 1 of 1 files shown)


till:
This was done in #2499

Issue #2339, opened 2025-09-09
ensure logging quota is not exceeded


till:
## Description of the issue 📄

We currently exceed the free log quota on Solarwinds/Papertrail.

## Steps to fix 🛠

Install the semantic logging gem to condense log lines:

Add this to the `Gemfile`:

```Gemfile
gem "amazing_print"
gem "rails_semantic_logger"
```

Run: `bundle install`

For staging:

```rb
# config/environments/staging.rb
if ENV["RAILS_LOG_TO_STDOUT"].present?
  $stdout.sync = true
  config.rails_semantic_logger.add_file_appender = false
  config.semantic_logger.add_appender(io: $stdout, formatter: config.rails_semantic_logger.format)
end
```

For production:

```rb
# config/environments/production.rb
if ENV["RAILS_LOG_TO_STDOUT"].present?
  $stdout.sync = true
  config.rails_semantic_logger.add_file_appender = false
  config.semantic_logger.add_appender(io: $stdout, formatter: config.rails_semantic_logger.format)
end
```

For Puma (in both environments):

```rb
# config/puma.rb
on_worker_boot do
  # Re-open appenders after forking the process
  SemanticLogger.reopen
end
```

Profit!

## To do 📋

* [x] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [x] Fork and clone the repository
* [x] Update the relevant files. Follow the steps to fix section in this issue.
* [x] Commit your changes as one commit. Use the title of this issue as your commit message
* [x] Submit a pull request
* [x] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

PR #2342, opened 2025-09-09
fix(logging): implement semantic logger

till:
we currently run out of log quota on a daily basis, so this
change aims to condense the log lines so we don't. :)

therefor we now use the rails_semantic_logger gem as its purpose
is to help with that. this commit also replaces rails_12factor
(which is why it's removed) since it would otherwise conflict with
the new gem.

logging config for all "environments" gets appeneded in application.rb
now.

resolves: #2339


jonodrew:
Nice one @till !

PR #2343, opened 2025-09-09
Update to Rails 7.2

olleolleolle:
This PR upgrades the configuration to Rails 7.2 configuration.

I attempted to keep close to what Rails outputs, so that the update to the next version of Rails is not so difficult.


olleolleolle:
@mroderick Now it is 1 commit.

config/environments/production.rb
```diff
@@ -21,7 +21,7 @@
   # config.require_master_key = true
 
   # Disable serving static files from `public/`, relying on NGINX/Apache to do so instead.
-  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
+  # config.public_file_server.enabled = false
```
till: How is this changing the app?

config/initializers/filter_parameter_logging.rb
```diff
@@ -4,5 +4,5 @@
 # Use this to limit dissemination of sensitive information.
 # See the ActiveSupport::ParameterFilter documentation for supported notations and behaviors.
 Rails.application.config.filter_parameters += [
-  :passw, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
+  :passw, :email, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
```
till: 👍

config/puma.rb
```diff
@@ -1,51 +1,42 @@
...
-
-# Specifies the `environment` that Puma will run in.
+# The ideal number of threads per worker depends both on how much time the
+# application spends waiting for IO operations and on how much you wish to
+# to prioritize throughput over latency.
 #
-environment ENV.fetch("RAILS_ENV") { "development" }
-
-# Specifies the `pidfile` that Puma will use.
-pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
```
till: I see, this moved to below.

config/environments/production.rb
olleolleolle: The default is true.

The ENV is set on Heroku. 

The result; the same.

config/cable.yml
```diff
@@ -0,0 +1,10 @@
+development:
+  adapter: async
+
+test:
+  adapter: test
+
+production:
+  adapter: redis
+  url: <%= ENV.fetch("REDIS_URL") { "redis://localhost:6379/1" } %>
```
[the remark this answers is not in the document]
olleolleolle: Right perhaps we also don't use Action Cable (websocket).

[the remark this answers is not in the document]
olleolleolle: Mostly because it is automatically generated each time the upgrade dance is performed. Benefit: By having it, unused, there's less diff to manage when upgrading.

mroderick (approved):

PR #2402, opened 2025-11-27
Stub Flodesk requests

### LINKED ISSUE #2322: Consider mocking out Flodesk::Client in specs
Follow up to #2321 

Currently the Flodesk client is intentionally enabled when running in the test environment as a result of setting the Flodesk API to "test" in the initializer:

https://github.com/codebar/planner/blob/45fd9d14f91bbbc80c13412983b51d75ff285c57/config/initializers/flodesk.rb#L3

This means all requests to Flodesk that are triggered during tests will be failing due to the invalid API key. 

Requests will be triggered whenever we fabricate a student or a coach (as a result of a `after_create` callback on `Subscription`).

We can avoid the unnecessary requests, and potentially improve test performance by mocking out `Flodesk::Client`

gnclmorais:
Addresses the issue by disabling web requests while running tests (something I believe to be a good practice) and stubs requests specifically to Flodesk’s API. Fixes #2322

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,7 @@
 require 'simplecov'
 require 'simplecov-lcov'
 require 'shoulda/matchers'
+require 'webmock/rspec'
 
 # Fix incompatibility of simplecov-lcov with older versions of simplecov that are not expresses in its gemspec.
 # https://github.com/fortissimo1997/simplecov-lcov/pull/25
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -70,6 +74,11 @@ def self.branch_coverage?
   end
 
   config.before(:each) do
+    # Stub all Flodesk API endpoints globally so tests don't make external requests
+    # when fabricating members (which trigger Subscription.after_create callback)
```
olleolleolle: Good context!

olleolleolle (approved):
Thank you! Good to have that in place.

PR #2410, opened 2025-12-02
fix: increase max_run_time for delayed job

mroderick:
This hopefully fixes the timeouts that we're seeing in https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/640/occurrence/449286860545#detail

The Heroku scheduler runs the job every 10 minutes, and thus gives us 10 minutes to complete the work. Hopefully we can complete it in 9 minutes.

image

Ping @mikej @olleolleolle: does it make sense to use 9 minutes timeout, or shall we just let the Heroku scheduler kill the job, hoping it will resume the work in the next iteration?

Also, would it make sense to switch to [`deliver_later`](https://api.rubyonrails.org/v8.1.1/classes/ActionMailer/MessageDelivery.html#method-i-deliver_later) in [`invite_coaches_to_virtual_workshop`](https://github.com/codebar/planner/blob/master/app/models/concerns/workshop_invitation_manager_concerns.rb#L68)?

This replaces https://github.com/codebar/planner/pull/2127

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -1,7 +1,10 @@
 Delayed::Worker.destroy_failed_jobs = false
 Delayed::Worker.sleep_delay = 60
 Delayed::Worker.max_attempts = 3
-Delayed::Worker.max_run_time = 5.minutes
+
+# The rake jobs:workoff task in Heroku scheduler is configured to run every 10 minutes
+Delayed::Worker.max_run_time = 9.minutes 
+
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
Worth a try!

Edit: to your question, I think that absolutely, async is easier.

gnclmorais:
If this works, it would be quite useful! Related, have we thought about using Sidekiq instead, like [Heroku recommends in its documentation](https://devcenter.heroku.com/articles/delayed-job)?

PR #2442, opened 2026-01-21
Avoid potential XSS vulnerability in flash messages

mikej:
Flash messages were being assumed to be HTML-safe which feels too fragile as it relies on every developer remembering to escape any user input that's included in flash messages.

Fix by removing the `.html_safe` from the messages partial.

The only place that seems to be relying on this behaviour is in `MeetingsController` where validation errors are joined into a single flash message using `<br>` tags. Updated this to just join the messages with commas instead.

### DIFF
--- a/spec/features/admin/meeting_spec.rb
+++ b/spec/features/admin/meeting_spec.rb
@@ -31,7 +31,7 @@ RSpec.feature 'Managing meetings', type: :feature do
 
       click_on 'Save'
 
-      expect(page).to have_content('Venue can\'t be blank')
+      expect(page).to have_content('Venue must be set')
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/views/layouts/_messages.html.haml
```diff
@@ -3,11 +3,11 @@
...
     - if msg.is_a?(String)
       .alert.alert-dismissible.fade.show.mb-0{ 'data-alert': '', class: "alert-#{name}", role: 'alert' }
-        = content_tag :div, msg.html_safe
+        = content_tag :div, msg
         %button.btn-close{ type: 'button', 'data-bs-dismiss': 'alert', 'aria-label': 'Close' }
     - elsif msg.is_a?(Array)
       - msg.each do |message|
         .alert.alert-dismissible.fade.show.mb-0{ 'data-alert': '', class: "alert-#{name}", role: 'alert' }
-          = content_tag :span, message.html_safe
+          = content_tag :span, message
```
olleolleolle:
```suggestion
          %span= message
```

PR #2444, opened 2026-01-22
Remove unnecessary use of content_tag in flash messages partial

mikej:
Small improvement to the partial used for flash messages that was suggested by @olleolleolle in the review comments for #2442


KimberleyCook (approved):
🎉

PR #2461, opened 2026-02-02
chore: set correct year for impact report link

mroderick:
This PR sets `2025` as the year for the banner about the impact report.

image


mroderick:
This was superseded by #2460

PR #2562, opened 2026-04-11
Replace Chosen.js with TomSelect for member lookup

mroderick:
## Why

The member lookup on `/admin/members` currently loads **all members** into a dropdown using Chosen.js. This approach has two problems:

1. **Chosen.js is unmaintained** - The repository has been archived and hasn't received updates since 2017. It has known issues with modern browsers and accessibility.

2. **Scaling problem** - The database has ~28,000 members. Loading all of them into a dropdown on every page load is slow and will only get worse as membership grows.

## What

Replace Chosen.js with TomSelect for the member lookup field, using remote data loading:

- TomSelect fetches members on-demand via a new `/admin/members/search` JSON endpoint
- Minimum 3 characters before search triggers
- 300ms debounce on requests
- Searches both name and email fields
- Results limited to 50 matches

## Changes

**Backend:**
- Add `GET /admin/members/search` JSON endpoint
- Search by name or email with ILIKE query
- Return `id`, `name`, `surname`, `email`, `full_name`

**Frontend:**
- Load TomSelect from CDN (CSS + JS) only on `/admin/members`
- Initialize TomSelect on `#member_lookup_id`
- Exclude from Chosen.js initialization

### 1. Make yourself an admin

Open a Rails console:

```bash
bundle exec rails console
```

Find your member record and add the admin role:

```ruby
member = Member.find_by(email: "your-email@example.com")
member.add_role(:admin)
```

### 2. Start the server

```bash
bundle exec rails server
```

### 3. Test the member search

1. Navigate to `http://localhost:3000/admin/members`
2. The dropdown should show "Type to search members..." placeholder
3. Type 2 characters - no search should trigger
4. Type 3+ characters - search triggers with 300ms debounce
5. Results show member name and email
6. Select a member - "View Profile" link updates
7. Click "View Profile" - navigates to member page

### 4. Run the tests

```bash
bundle exec rspec spec/controllers/admin/members_controller_spec.rb spec/features/admin/tom_select_member_lookup_spec.rb
```

All 8 tests should pass.

## Suggested Searches (from local dev database)

Try these searches to verify the functionality:

| Search Term | Matches |
|-------------|---------|
| `Sal` | **Salvador Tremblay** (cory_okon@blick.example) |
| `Salvador` | **Salvador Tremblay** (cory_okon@blick.example) |
| `blick` | **Salvador Tremblay** (cory_okon@blick.example) |
| `Emm` | **Emmie Robel** (wiley_simonis@smith-smith.example) |
[...]

PR #2564, opened 2026-04-11
Replace Chosen.js with TomSelect for member lookup

mroderick:
## Why

Chosen.js is unmaintained (last release 2014) and doesn't support remote data sources. Loading 28,000 members into a dropdown is impractical and causes performance issues.

TomSelect is actively maintained and supports remote search out of the box.

## What Changed

- Added `/admin/members/search` endpoint for remote member lookup (searches by name/email, min 3 chars, max 50 results)
- Integrated TomSelect loaded from CDN (avoided SassC incompatibility with modern CSS)
- Updated member lookup dropdown to use remote search with 300ms debounce

## How to Verify

1. Start the server locally: `bundle exec rails server`
2. Make yourself an admin: `bundle exec rails c` then `Member.first.add_role(:admin)`
3. Visit http://localhost:3000/admin/members
4. In the member lookup dropdown, try these searches:
   - `jane` - returns members with Jane in their name
   - `@example.com` - returns members with example.com emails
   - Type at least 3 characters to trigger search

## Suggested Searches (based on dev database)

Try searching for common patterns:
- First names: `jane`, `john`, `mary`
- Email domains: `@example.com`
- Partial names: `robe` for Robertson

All tests passing (6 controller + 2 feature).

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -40,12 +40,43 @@ $(function() {
     format: "HH:i",
   });
 
+  // TomSelect for admin member lookup
+  if ($('#member_lookup_id').length) {
+    new TomSelect('#member_lookup_id', {
+      placeholder: 'Type to search members...',
+      valueField: 'id',
+      labelField: 'full_name',
+      searchField: ['full_name', 'email'],
+      create: false,
+      loadThrottle: 300,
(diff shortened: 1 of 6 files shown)


mroderick:
Closing in favor of #2562

PR #2568, opened 2026-04-11
fix(test): freeze time in tests to prevent flaky failures

mroderick:
## Summary

Remove `timecop` gem dependency and fix flaky tests caused by time-based comparisons without freezing time.

## Problem

CI was failing due to flaky tests. The root cause: time-based scopes compare against `Time.zone.now` evaluated at query time, while test fixtures use relative times (`2.days.ago`, `1.week.from_now`) evaluated at creation time. Any time gap between fixture creation and query execution could cause incorrect results.

## Solution

Wrap time-dependent tests in `travel_to(Time.current)` blocks, ensuring all time calculations reference the same frozen moment.

## Files Changed

- `Gemfile`, `Gemfile.lock` - Remove timecop dependency
- `config/environments/test.rb` - Remove timecop require and config  
- `spec/support/shared_examples/behaves_like_an_invitation.rb` - Freeze time in invitation scope tests
- `spec/models/concerns/listable_spec.rb` - Replace Timecop with travel_to, add time freezing
- `spec/models/attendance_warning_spec.rb` - Freeze time in last_six_months test
- `spec/features/*` - Freeze time in feature tests with relative dates
- `spec/presenters/*` - Freeze time in presenter tests
- `spec/lib/tasks/*` - Freeze time in rake task tests

## Test Plan

- [x] 508 non-feature tests pass (1 unrelated flaky test)
- [x] All time-dependent tests verified with multiple seeds
- [x] Standardized on ActiveRecord `travel_to` over `timecop`

Fixes flaky test failures in CI.

### DIFF
--- a/spec/features/admin/announcements_spec.rb
+++ b/spec/features/admin/announcements_spec.rb
@@ -15,7 +15,7 @@ RSpec.feature 'Announcements', type: :feature do
         click_on 'announcement[create]'
 
         expect(page).to have_content('Announcement successfully created')
-        expect(page.current_path).to eq(admin_announcements_path)
+        expect(page).to have_current_path(admin_announcements_path, ignore_query: true)
       end
     end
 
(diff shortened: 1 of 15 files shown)


mroderick:
@olleolleolle what do you think about removing TimeCop? Is that the way forwards for Rails projects?

olleolleolle:
@mroderick It’s nice to not have to have a special dependency if we get all we need from the built-in. If we can get away with not having Timecop, I say thank it and let it go.

spec/features/admin/announcements_spec.rb
```diff
@@ -15,7 +15,7 @@
         click_on 'announcement[create]'
 
         expect(page).to have_content('Announcement successfully created')
-        expect(page.current_path).to eq(admin_announcements_path)
+        expect(page).to have_current_path(admin_announcements_path, ignore_query: true)
```
olleolleolle, line 18: This change is excellent and adds an implicit wait, which the previous form didn’t. That was also a source of flakes.

olleolleolle (approved):
I can see that there were multiple kinds of fixes (I assume lint on save made a few of them).

Among fixes in here:
- RSpec metadata `foo: true` can be just the Symbol `:foo`.
- Use single quotes when possible (perhaps a low-value lint rule, but it’s there, so this harmonizes the codebase).
- Use the `describe` RSpec directive to wrap tests about a method.

👍

PR #2571, opened 2026-04-13
fix: validate email format and add defensive mailer checks

mroderick:
## Summary

Fixes `Net::SMTPSyntaxError: 501 Recipient syntax error` that occurred in production when sending workshop attendance reminders.

## Root Cause Analysis

The error occurs in `WorkshopInvitationManagerConcerns#send_workshop_attendance_reminders` when iterating over `workshop.attendances.not_reminded` and sending reminder emails via `WorkshopInvitationMailer#attending_reminder`.

**Root cause**: Member email addresses are not validated before SMTP delivery. The `EmailHeaderHelper#mail_args` method passes `member.email` directly to the mailer without format validation.

### Affected Data

Found 18 members with malformed email addresses in production database. Examples include:

| Member ID | Email | Issue |
|----------|------|-------|
| XXXX | user@example.com | Valid (for comparison) |
| XXXX | missing-atsign.example.com | Missing @ |
| XXXX | user@missingtld | Missing TLD |
| XXXX | user@nodottld | No . |
| XXXX | Matt at made by lamp dot com | Human typo |
| XXXX | singlechar | Single character |

These records likely originated from:
- Very early data (before validation existed)
- Manual admin entry errors
- GitHub OAuth edge cases

### 1. Database Cleanup

Set invalid emails to NULL so they skip future mailings:

```sql
UPDATE members SET email = NULL 
WHERE email IS NOT NULL 
  AND email !~ '^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$';
```

### 2. Member Model Validation

`app/models/member.rb` — Added email validation using `email_validator` gem:

```ruby
validates :email, email: { mode: :strict }, if: :can_log_in?
```

### 3. Defensive Mailer Check

`app/helpers/email_header_helper.rb` — Returns `nil` for invalid emails with warning log:

```ruby
def invalid_email?(email, member_id)
  return false if EmailValidator.valid?(email, mode: :strict)
  
  Rails.logger.warn("[EmailHeaderHelper] Invalid email for member_id=#{member_id}: #{email}")
  true
end
```

### 4. Added Gem

Added `email_validator` gem (32M+ downloads, MIT licensed) for:
- Single source of truth for email validation
- Multiple validation modes (`:loose`, `:strict`, `:rfc`)
- Standalone API for use outside models

### 1. Run the cleanup SQL

Connect to production database (Heroku postgres or production Postgres):

```bash
# Preview affected records
SELECT id, email FROM members 
WHERE email IS NOT NULL 
  AND email !~ '^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$';

# Set to NULL
UPDATE members SET email = NULL 
WHERE email IS NOT NULL 
  AND email !~ '^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$';
```

### 2. Verify cleanup

```sql
-- Should return 0
SELECT COUNT(*) FROM members 
WHERE email IS NOT NULL 
  AND email !~ '^[A-Za-z0-9._%+-]+@[A-Za-z0-9.-]+\.[A-Za-z]{2,}$';
```

### 3. Check logs

After deploying, watch for warning logs when rake tasks run:

```
[EmailHeaderHelper] Invalid email for member_id=XXXX: invalid-email
```

## Tests

- 6 new tests for email format validation in `spec/models/member_spec.rb`
- 9 new tests for `EmailHeaderHelper#mail_args` in `spec/helpers/email_header_helper_spec.rb`

Run tests locally:
```bash
bundle exec rspec spec/models/member_spec.rb spec/helpers/email_header_helper_spec.rb
```

## Impact

- **New validations**: Members with `can_log_in=true` must provide valid email format
- **Backward compatibility**: Existing members without `can_log_in` can still have NULL emails
- **Mailer defense**: Even if bad data slips through, SMTP errors are prevented with logging

### DIFF
--- a/spec/helpers/email_header_helper_spec.rb
+++ b/spec/helpers/email_header_helper_spec.rb
@@ -1,63 +1,58 @@
-RSpec.describe EmailHeaderHelper do
-  subject do
-    Class.new do
-      include EmailHeaderHelper
-      public :mail_args
-    end.new
-  end
+RSpec.describe EmailHeaderHelper, type: :helper do
+  before { EmailHeaderHelper.module_eval { public :mail_args } }
 
   describe '#mail_args' do
     let(:member) { Struct.new(:id, :email).new(1, 'test@example.com') }
 
     it 'returns mail arguments for valid email' do
-      result = subject.mail_args(member, 'Test Subject')
+      result = helper.mail_args(member, 'Test Subject')
       expect(result[:to]).to eq('test@example.com')
       expect(result[:subject]).to eq('Test Subject')
     end
 
     it 'returns nil for nil email' do
       member = Struct.new(:id, :email).new(1, nil)
-      result = subject.mail_args(member, 'Test Subject')
+      result = helper.mail_args(member, 'Test Subject')
       expect(result).to be_nil
     end
 
     it 'returns nil for blank email' do
       member = Struct.new(:id, :email).new(1, '')
-      result = subject.mail_args(member, 'Test Subject')
+      result = helper.mail_args(member, 'Test Subject')
       expect(result).to be_nil
     end
 
     it 'returns nil for invalid email format' do
       member = Struct.new(:id, :email).new(1, 'invalid-email')
-      result = subject.mail_args(member, 'Test Subject')
+      result = helper.mail_args(member, 'Test Subject')
       expect(result).to be_nil
     end
 
     it 'returns nil for email missing @ symbol' do
       member = Struct.new(:id, :email).new(1, 'invalidexample.com')
-      result = subject.mail_args(member, 'Test Subject')
+      result = helper.mail_args(member, 'Test Subject')
       expect(result).to be_nil
     end
 
     it 'returns nil for email missing TLD' do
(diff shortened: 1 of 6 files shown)


spec/helpers/email_header_helper_spec.rb
```diff
@@ -0,0 +1,65 @@
+RSpec.describe EmailHeaderHelper do
+  subject do
+    Class.new do
+      include EmailHeaderHelper
+      public :mail_args
+    end.new
+  end
```
olleolleolle: If I recall correctly, there is a ready made helper variable which has this helper instantiated. So, this subject can be fully replaced by "helper" sans setup.

mroderick:
Thank you for the suggestion! You're right - I've updated the spec to use the idiomatic  pattern which automatically includes the helper module. Much cleaner now.

Commit: fcdf5154

olleolleolle (approved):
Nice!

PR #2579, opened 2026-04-17
fix: #students and #coaches return Chapter-specific students and coaches

jonodrew:
This fixes a bug where the methods returned _all_ of the students and coaches that the user could access

`#coaches` query: `"SELECT DISTINCT \"members\".* FROM \"members\" INNER JOIN \"subscriptions\" ON \"members\".\"id\" = \"subscriptions\".\"member_id\" INNER JOIN \"groups\" ON \"subscriptions\".\"group_id\" = \"groups\".\"id\" WHERE \"groups\".\"chapter_id\" = 1 AND \"groups\".\"name\" = 'Coaches'"`

`#students` query: `"SELECT DISTINCT \"members\".* FROM \"members\" INNER JOIN \"subscriptions\" ON \"members\".\"id\" = \"subscriptions\".\"member_id\" INNER JOIN \"groups\" ON \"subscriptions\".\"group_id\" = \"groups\".\"id\" WHERE \"groups\".\"chapter_id\" = 1 AND \"groups\".\"name\" = 'Students'"`

### DIFF
--- a/spec/models/chapter_spec.rb
+++ b/spec/models/chapter_spec.rb
@@ -65,4 +65,34 @@ RSpec.describe Chapter do
       expect(Rails.cache.read(cache_key)).to be_nil
     end
   end
+
+  describe "helper methods return only that chapter's coaches/students" do
+    RSpec.shared_examples 'group-scoped members' do |group_name, method_name|
+      let(:this_chapter) { Fabricate(:chapter) }
+      let(:that_chapter) { Fabricate(:chapter) }
+
+      let!(:this_group)  { Fabricate(:group, chapter: this_chapter, name: group_name) }
+      let!(:that_group)  { Fabricate(:group, chapter: that_chapter, name: group_name) }
+
+      let!(:this_member) { Fabricate(:member) }
+      let!(:that_member) { Fabricate(:member) }
+
+      before do
+        Fabricate(:subscription, group: this_group, member: this_member)
+        Fabricate(:subscription, group: that_group, member: that_member)
+      end
+
+      it "returns only #{group_name.downcase} for the chapter" do
+        expect(this_chapter.public_send(method_name))
+          .to contain_exactly(this_member)
+      end
+
+      it "does not include #{group_name.downcase} from another chapter" do
+        expect(this_chapter.public_send(method_name))
+          .not_to include(that_member)
+      end
+    end
+    it_behaves_like 'group-scoped members', 'Students', :students
+    it_behaves_like 'group-scoped members', 'Coaches', :coaches
+  end
 end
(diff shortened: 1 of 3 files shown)


mikej:
I'll give this one a quick test locally, but feels like a nice approach 👍

mroderick:
Could you add some tests to make sure this stays fixed?

jonodrew:
@mroderick done!

spec/models/chapter_spec.rb
```diff
@@ -65,4 +65,32 @@
...
+      it "returns only #{group_name.downcase} for the chapter" do
+        expect(this_chapter.public_send(method_name))
+          .to contain_exactly(this_member)
+      end
+
+      it "does not include #{group_name.downcase} from another chapter" do
+        expect(this_chapter.public_send(method_name))
+          .not_to include(that_member)
+      end
+    end
```
mikej, line 94: Good idea to use some shared examples here, but currently these test cases won't actually get run when running `chapter_spec.rb`. After defining the `shared_examples`, you also need to add `include_examples` lines with the appropriate `group_name` and `method_name` for students and coaches.

e.g. inside the `describe "helper methods...` but after the `RSpec.shared_examples` you can put:

```ruby
include_examples "group-scoped members", "Coaches", "coaches"
include_examples "group-scoped members", "Students", "students"
```

On doing this, the specs then fail with `Validation failed: Group has already been taken` so looks like the `before` results in trying to create the groups twice?

Are you OK to take a look at this as I agree having these tests in place is definitely worthwhile.

Lastly, if you want to check which test cases RSpec is running you can use the "documentation" format which lists the passing tests as well as any failures: `bin/drspec -fd spec/models/chapter_spec.rb`

mikej (changes requested):
Hi! The actual change to `chapter.rb` looks good (had a check of the queries that get generated).

Added a comment about how to include the shared examples in the spec though.

spec/models/chapter_spec.rb
jonodrew: Thank you so much for this! I hate Ruby 😅

jonodrew: I've taken AI advice and put in a fix that seems to work. can you look over it?

mroderick:
There are a couple of conflicts to resolve in this PR, then we can merge it! 🚀

KimberleyCook:
@jonodrew just a gentle nudge about solving these conflicts so we can get this merged in :) thanks!

jonodrew:
> @jonodrew just a gentle nudge about solving these conflicts so we can get this merged in :) thanks!

Done - was a bit trickier than anticipated

KimberleyCook:
@jonodrew thank you 🫶

KimberleyCook (approved):
🎉

PR #2582, opened 2026-04-22
fix(tc): handle unauthenticated and already-accepted users

mroderick:
## Summary

Two bug fixes for the Terms and Conditions page:

### Bug: Unauthenticated users crash when submitting form
- **Problem**: Unauthenticated users could visit `/terms_and_conditions` and click "Accept", which would crash because `current_user` is nil
- **Fix**: Controller checks login status and redirects to GitHub OAuth if not logged in
- **Commits**: `b28f8eb`

### Bug: Users who already accepted T&C see the form again
- **Problem**: Logged-in users who have already accepted T&C would see the acceptance form again
- **Fix**: Shows a message "You have already accepted our Code of Conduct on [date]." instead of the form
- **Commits**: `b56caa4`

### Changes
- `app/controllers/terms_and_conditions_controller.rb` - add login check
- `app/views/terms_and_conditions/show.html.haml` - conditional display for different states
- `config/locales/en.yml` - new locale strings
- `spec/features/accepting_terms_and_conditions_spec.rb` - new tests

### Testing
- All tests pass (7/8, 1 pre-existing Playwright setup issue unrelated to this change)

### DIFF
--- a/spec/features/accepting_terms_and_conditions_spec.rb
+++ b/spec/features/accepting_terms_and_conditions_spec.rb
@@ -4,7 +4,7 @@ RSpec.feature 'Accepting Terms and Conditions', type: :feature do
       mock_github_auth
     end
 
-    scenario 'they can''t proceed unless they accept the ToCs' do
+    scenario "they can't proceed unless they accept the ToCs" do
       visit root_path
       click_on 'Join us as a student'
       click_on 'Join us as a student'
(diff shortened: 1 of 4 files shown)


spec/features/accepting_terms_and_conditions_spec.rb
```diff
@@ -4,7 +4,7 @@
       mock_github_auth
     end
 
-    scenario 'they can''t proceed unless they accept the ToCs' do
+    scenario 'they can' + 't proceed unless they accept the ToCs' do
```
olleolleolle: Minor: it's possible to pick other quoting, there are many:

```suggestion
    scenario %q{they can't proceed unless they accept the ToCs} do
```

This goes for all of the other ones that had the wonky solution to apostrophes.

olleolleolle (approved):
Nice to have!

I made a note on code formatting.

PR #2584, opened 2026-04-22
fix: handle blank email from GitHub OAuth

mroderick:
## Summary

- Validates email is present before attempting to find or create a Member
- Shows friendly error message when GitHub returns a blank/missing email
- Prevents ActiveRecord::RecordInvalid error with "email can't be blank, already taken, is invalid"

## Root Cause

When GitHub returns a blank/nil email:
1. Controller tried to find/create Member with blank email
2. Found existing member with blank email in database  
3. Uniqueness validation failed with multiple errors

Fix adds early validation of email presence before any database operations.

This fixes https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/663#detail

image

### DIFF
--- a/app/controllers/auth_services_controller.rb
+++ b/app/controllers/auth_services_controller.rb
@@ -26,8 +26,15 @@ class AuthServicesController < ApplicationController
 
       finish_registration || redirect_to(referer_or_dashboard_path)
     else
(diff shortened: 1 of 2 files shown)


KimberleyCook (approved):
🎉

PR #2592, opened 2026-04-25
feat: add configurable async email delivery for workshop invitations

mroderick:
## Summary

- Configure ActiveJob to use DelayedJob adapter for background email processing
- Add `ApplicationJob` with global error handling for failed jobs
- Create `AsyncEmailConcern` for chapter-based async email feature flag
- Update `WorkshopInvitationManager` to send emails asynchronously for chapters in `ASYNC_EMAIL_CHAPTER_IDS`

## Hypothesis

DelayedJob jobs timeout after 540s on Heroku when processing bulk workshop invitations. This happens because emails are sent synchronously (`deliver_now`) inside loops, blocking on SMTP I/O for each email.

**At scale:**
- London chapter: 5,454 students + 1,988 coaches = 7,442 potential invitations
- Recent workshops send 2,000-4,000 invitations
- 4,000 × ~200ms per email = ~800 seconds **> 540s timeout**

## How the Fix Works

1. For chapters in `ASYNC_EMAIL_CHAPTER_IDS`, emails are enqueued with `.deliver_later` instead of blocking with `.deliver_now`
2. Each email becomes an independent DelayedJob job queued in the database
3. The original "send invitations" job completes in milliseconds, not minutes - no timeout
4. The worker processes queued email jobs separately

**What happens with Heroku Scheduler running every 10 minutes:**

| Time | Event |
|------|-------|
| T+0 | Invitation job runs, queues 2,000 emails as separate jobs |
| T+10min | `jobs:workoff` runs, picks up first batch of email jobs |
| T+20min | Next `jobs:workoff`, picks up more |
| ... | Continues until all emails sent |

**Expected behavior:**
- Emails arrive in batches over 10-20 minutes instead of all at once
- No single job exceeds 540s because each email is its own job (~1-2s each)
- Total send time increases but completes reliably

**Trade-off:** Emails are sent over ~10-20 minutes instead of ~13 minutes synchronously. This is preferable to timed-out failures.

## Feature Flag

`ASYNC_EMAIL_CHAPTER_IDS` env var:
- **Empty** (default): All chapters use sync delivery (safe fallback)
- **Specific IDs**: Only those chapters use async delivery

**Recommended rollout:**
1. Deploy with empty `ASYNC_EMAIL_CHAPTER_IDS` (safe - no behavior change)
2. Set `ASYNC_EMAIL_CHAPTER_IDS=1` to enable async for London only
3. Monitor for 1-2 workshops
4. Gradually add more chapters or remove the limit entirely

## Changes

1. **config/application.rb** - Set `config.active_job.queue_adapter = :delayed_job` and add feature flag config
2. **app/jobs/application_job.rb** - Base job class with global error handling  
3. **app/models/concerns/async_email_concern.rb** - Concern with `async_email_for_chapter?` helper
4. **app/models/concerns/workshop_invitation_manager_concerns.rb** - 7 conditional async email calls

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,2 +1,6 @@
 class ApplicationJob < ActiveJob::Base
+  rescue_from(Exception) do |exception|
+    Rails.error.report(exception)
+    raise
+  end
 end
--- a/app/models/concerns/async_email_concern.rb
+++ b/app/models/concerns/async_email_concern.rb
@@ -1,0 +1,11 @@
+module AsyncEmailConcern
+  extend ActiveSupport::Concern
+
+  private
+
+  def async_email_enabled?(chapter)
+    return false if chapter.nil?
+    return false if Rails.application.config.async_email_chapter_ids.empty?
+    Rails.application.config.async_email_chapter_ids.include?(chapter.id)
+  end
+end
--- a/app/models/concerns/workshop_invitation_manager_concerns.rb
+++ b/app/models/concerns/workshop_invitation_manager_concerns.rb
@@ -2,6 +2,7 @@ module WorkshopInvitationManagerConcerns
   extend ActiveSupport::Concern
 
   included do
+    include AsyncEmailConcern
     include InstanceMethods
   end
 
@@ -9,7 +10,8 @@ module WorkshopInvitationManagerConcerns
     def send_workshop_attendance_reminders(workshop)
       workshop_mailer = workshop.virtual? ? VirtualWorkshopInvitationMailer : WorkshopInvitationMailer
       workshop.attendances.not_reminded.each do |invitation|
-        workshop_mailer.send(:attending_reminder, workshop, invitation.member, invitation).deliver_now
+        deliver_method = async_email_enabled?(workshop.chapter) ? :deliver_later : :deliver_now
+        workshop_mailer.send(:attending_reminder, workshop, invitation.member, invitation).public_send(deliver_method)
         invitation.update(reminded_at: Time.zone.now)
       end
     end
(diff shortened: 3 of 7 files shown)


mroderick:
I am considering a way to use SendGrid's API to send email in batches of up to 1k each, but that's a bigger undertaking. This PR fixes the immediate problem _today_, and I can carefully design the other solution without rushing or jeopardising the workshop invitations functionality.

mroderick:
Once this is merged, I'll speak to some chapter organisers from the larger chapters to get them to help test drive this.

The last thing we need is for all chapter organisers to be disrupted at the same time.

app/jobs/application_job.rb
```diff
@@ -1,2 +1,6 @@
 class ApplicationJob < ActiveJob::Base
```
till, line 1: What's this for?

till (commented):
I am not overly familiar with all the Rubyism, but is codebar already using delay jobs? It seems so as there is not additional setup in this PR. Otherwise, LGTM.

Then my only question is: do we have enough database storage for all these new rows?

mroderick:
> Then my only question is: do we have enough database storage for all these new rows?

```
$ heroku pg:info -a codebar-production
=== DATABASE_URL, HEROKU_POSTGRESQL_MAUVE_URL

Plan:                  Standard 0
Status:                Available
Data Size:             593 MB / 64 GB (0.9%)
Tables:                48
```

I think we'll be ok for a while.

We could add a job that cleans out old (1-2y+) rows.

app/jobs/application_job.rb
mroderick: The `rescue_from(Exception)` block ensures that **any error in any job gets reported to our error tracking service** (Rollbar) before the job fails.

Without this, if a job raises an error:
- DelayedJob marks it as failed and retries
- But we have **no visibility** into what went wrong unless we manually check the logs

With this handler:
- `Rails.error.report(exception)` sends the error to Rollbar immediately
- We get alerts/notifications about job failures  
- We can debug issues without digging through Heroku logs

The `raise` at the end re-raises the exception so DelayedJob still handles retries/failures normally — we don't swallow errors, we just *report* them before they bubble up.

This is standard practice for background jobs in Rails apps. Let me know if you'd prefer a more specific exception class or different error handling approach.

till (approved):

PR #2600, opened 2026-04-28
fix: handle ActionController::UnknownFormat with 406 response

mroderick:
## Summary

Adds idiomatic Rails exception handling for `ActionController::UnknownFormat`, returning 406 Not Acceptable for unsupported request formats.

## Problem

Requests with non-HTML Accept headers (e.g., `Accept: application/json` from crawlers/API clients) cause `ActionController::UnknownFormat` exceptions when controllers don't have templates for those formats.

Example error from production:
```
ActionController::UnknownFormat: ChapterController#show is missing a template for this request format and variant.
request.formats: [\"application/json\"]
```

## Solution

Uses Rails' built-in `rescue_from` mechanism to handle `ActionController::UnknownFormat` exceptions centrally in `ApplicationController`.

This approach was inspired by @olleolleolle's suggestion to use `respond_to` blocks in controllers:
```ruby
respond_to do |format|
  format.html
  format.any { redirect_to support_path }
end
```

We adapted this to use `rescue_from` for a more centralized solution that follows the existing pattern in the codebase.

This is the idiomatic Rails approach, as documented in:
- [Rails API: ActionController::MimeResponds](https://api.rubyonrails.org/classes/ActionController/MimeResponds.html)
- [Rails Guides: Action Controller Advanced Topics - Exception Handling](https://guides.rubyonrails.org/action_controller_advanced_topics.html#rescue-from)

> \"Rails 4.0 raises `ActionController::UnknownFormat` when an action does not handle the requested format. By default, this exception is handled by responding with 406 Not Acceptable, but developers can now override this behavior.\"
> — [Rails Upgrade Guide](https://github.com/rails/rails/blob/main/guides/source/upgrading_ruby_on_rails.md)

## Changes

- Add `rescue_from ActionController::UnknownFormat, with: :render_not_acceptable` to `ApplicationController`
- Add `render_not_acceptable` method returning 406 status
- Add `not_acceptable.html.haml` error template
- Add comprehensive controller tests for HTML, JSON, and XML formats

## Response Behavior

| Format | Response |
|--------|----------|
| HTML | Renders `errors/not_acceptable` template with 406 status |
| JSON/XML/etc | Returns empty response with 406 status |

## Testing

```bash
bundle exec rspec spec/controllers/application_controller_spec.rb
```

All 66 controller tests pass.

## Related

- Replaces: #2585 (which used global config approach)
- Thanks to @olleolleolle for the `respond_to` suggestion that led to this solution!

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -18,6 +18,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from ActionController::RoutingError, with: :render_not_found
   rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
+  rescue_from ActionController::UnknownFormat, with: :render_not_acceptable
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
   rescue_from Pundit::AuthorizationNotPerformedError, with: :user_not_authorized
@@ -33,6 +34,13 @@ class ApplicationController < ActionController::Base
     respond_to do |format|
       format.html { render template: 'errors/not_found', layout: false, status: :not_found }
       format.all { head :not_found }
+    end
+  end
+
+  def render_not_acceptable
+    respond_to do |format|
+      format.html { render template: 'errors/not_acceptable', layout: false, status: :not_acceptable }
+      format.all { head :not_acceptable }
     end
   end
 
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Thanks!

PR #2605, opened 2026-05-04
Fix workshop capacity checks

mroderick:
## Summary

This PR fixes a bug in workshop capacity checks where the workshop fabricator had incorrect transient variable syntax, and the controller/view were using venue attributes instead of workshop attributes for capacity checks.

### 1. Fix workshop fabricator bug
- Fixed `transients[:coach_count || 10]` → `transients[:coach_count] || 10`
- The bug was causing the coach_count transient to always be nil (truthy || 10 returns 10)
- Added explicit `student_spaces` and `coach_spaces` attributes to workshop fabricator

### 2. Fix controller capacity checks
- Updated `available_spaces?` to use `event_student_spaces?` and `event_coach_spaces?`
- These methods check workshop attributes directly instead of delegating to venue

### 3. Fix view capacity checks
- Updated workshop invitation view to use `event_coach_spaces?` and `event_student_spaces?`
- Ensures consistent capacity checking between controller and view

### 4. Update test files
- Updated `accepting_invitation_spec.rb` and `coach_accepting_invitation_spec.rb`
- Tests now set `workshop.student_spaces` and `workshop.coach_spaces` directly
- This matches how the controller now checks capacity

## Why this fix is needed

The previous implementation checked capacity via the venue (sponsor) attributes:
```ruby
workshop.host.seats  # venue attribute
workshop.host.number_of_coaches  # venue attribute
```

But the controller should check workshop attributes directly:
```ruby
workshop.student_spaces  # workshop attribute
workshop.coach_spaces  # workshop attribute
```

This ensures that capacity limits are controlled per-workshop, not per-venue.

## Testing

All 995 tests pass with this fix.

### DIFF
--- a/app/controllers/workshop_invitation_controller.rb
+++ b/app/controllers/workshop_invitation_controller.rb
@@ -91,8 +91,8 @@ class WorkshopInvitationController < ApplicationController
   end
 
   def available_spaces?(workshop, invitation)
-    (invitation.role.eql?('Student') && workshop.student_spaces?) ||
-      (invitation.role.eql?('Coach') && workshop.coach_spaces?)
+    (invitation.role.eql?('Student') && workshop.event_student_spaces?) ||
+      (invitation.role.eql?('Coach') && workshop.event_coach_spaces?)
   end
 
   # Inline from InvitationControllerConcerns
--- a/app/views/workshop_invitation/show.html.haml
+++ b/app/views/workshop_invitation/show.html.haml
@@ -80,7 +80,7 @@
                   = @invitation.member.bans.active.first.reason
             - else
               - if @invitation.for_coach?
-                - if @workshop.coach_spaces?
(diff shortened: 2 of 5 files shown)


mroderick:
**Note on the nature of this bug:**

This is a pre-existing bug that we discovered during test optimization work. The capacity checks were inadvertently using venue (sponsor) attributes instead of workshop attributes:

**Before (buggy):**
-  → checked  (sponsor attribute)
-  → checked  (sponsor attribute)

**After (fixed):**
-  → checks  directly
-  → checks  directly

The tests were passing before because the fabricator was setting up the host with matching attributes, and tests were updating  to simulate capacity limits. We discovered the bug when trying to simplify tests to use workshop attributes directly.

KimberleyCook (approved):
🎉

PR #2623, opened 2026-05-27
Fix broken donate link from enthuse

KimberleyCook:


olleolleolle (approved):
Thanks for the fix!

KimberleyCook:
@olleolleolle Thaaanks, this is the type of dev work on the website I can do haha

PR #2493, opened 2026-02-18
feat: split how-to-support-us page into separate volunteer, fundraise, and donate pages

mroderick:
## Summary

- Split the monolithic `/how-to-support-us` page into three dedicated pages:
  - `/volunteer` - Volunteer opportunities (coach, organiser, other skills)
  - `/fundraise` - Fundraising events (running challenges)
  - `/donate` - Donation platforms
- Old `/how-to-support-us` now redirects to `/volunteer`
- Added new menu items: Volunteer, Fundraise in header
- Updated footer with links to all three new pages
- Removed "Buy us a coffee" button from header, added coffee icon in footer
- Changed "Jobs" to "Job Board" in nav and footer
- Removed Sponsors menu item
- Updated donation links to point directly to Enthuse

image
image

PR #2625, opened 2026-05-27
Delete old "How to support us" page

KimberleyCook:


KimberleyCook:
@mroderick Am I right in thinking this page can now be deleted?

mroderick:
Yes, you're right — this page can be deleted.

Here's what I checked to confirm:

- **Route**: `config/routes.rb` already redirects `/how-to-support-us` → `/volunteer`
- **Split pages**: The content was split into `/volunteer`, `/fundraise`, and `/donate` in #2493
- **References**: No views, controllers, or tests reference the old `how-to-support-us.html.haml` file
- **i18n keys**: The `pages.how_to_support_us.*` keys are still used by the new split pages (`volunteer.html.haml`, `fundraise.html.haml`), so those should stay
- **Images**: The `how-to-support-us/` image assets are also still used by the new pages, so those should stay too

The PR correctly removes only the now-unused view file. The route redirect means existing bookmarks/links still work. 👍

mroderick (approved):

PR #2627, opened 2026-05-27
add codebar courses links

AroraXD:
add links to the codebar courses microsite to the navigation and footer


KimberleyCook (approved):
Thank you 🎉

PR #2631, opened 2026-05-29
fix: wrap fabricator time attributes in blocks to prevent flaky tests

mroderick:
## Problem

The Fabrication gem evaluates attributes without blocks once and caches the value. When `Fabricate(:event)` was first called inside a `travel_to` block (e.g. in `event_spec.rb`), the `date_and_time` default was cached as a past date. All subsequent events fabricated in the same process reused that cached value, making them appear as past events.

This caused `view_event_spec` feature tests to fail because the event page showed **'This event has already occurred'** instead of the **'Log in'** link.

## Changes

- Wrap `date_and_time`, `rsvp_opens_at`, and `past_workshop.date_and_time` in Fabrication blocks so they evaluate dynamically at fabrication time
- Reset Capybara sessions after each feature spec as a defensive measure against session leaks between tests

## Verification

Reproduced the failure locally with the exact CI command and seed (41835, group 2). After the fix, the same command passes with 0 failures. Ran multiple additional seeds and the full test suite — all green.

### DIFF
--- a/spec/fabricators/event_fabricator.rb
+++ b/spec/fabricators/event_fabricator.rb
@@ -1,7 +1,7 @@
 Fabricate.sequence(:slug) { |i| "#{Faker::Lorem.word}-#{i}" }
 
 Fabricator(:event) do
-  date_and_time Time.zone.now + 2.days
+  date_and_time { Time.zone.now + 2.days }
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Looks good, did you experience flakes when not having the `reset_sessions!`? Anyway, good fix!

PR #2635, opened 2026-06-05
fix: safely skip emails to members with invalid addresses

mroderick:
## Problem

`NoMethodError: undefined method '[]' for nil` crashed `WorkshopInvitationMailer#invite_coach` (and other mailers) when a member's email was invalid. The root cause: `EmailHeaderHelper#mail_args` returned `nil` for invalid emails, and every mailer called `mail(mail_args(...))` directly, passing `nil` into ActionMailer.

Rollbar: https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/690#detail

## Changes

- Add `EmailHeaderHelper::SkippedEmail` null object with no-op delivery methods
- Add `mail_to_member` helper that validates email before calling `mail()`, logs the skip, and returns `SkippedEmail` when invalid
- Migrate all 8 mailers (27 methods) from `mail(mail_args(...))` to `mail_to_member(...)`
- Guard `EmailDelivery#log_sent_email` when `mail()` was never called (skips would otherwise trigger `MissingTemplate`)
- Remove the now-unused `mail_args` method
- Add skip-path tests for every mailer method (39 new tests)

## Verification

`bundle exec rspec spec/mailers/ spec/helpers/email_header_helper_spec.rb` — 80 examples, 0 failures

### DIFF
--- a/app/helpers/email_header_helper.rb
+++ b/app/helpers/email_header_helper.rb
@@ -1,14 +1,28 @@
 module EmailHeaderHelper
+  class SkippedEmail
+    def deliver_now; self; end
+    def deliver_later; self; end
+    def deliver; self; end
+    def deliver!; self; end
+  end
+
   private
 
-  def mail_args(member, subject, from_email = 'meetings@codebar.io', cc = '', bcc = '')
-    return nil if invalid_email?(member.email, member.id)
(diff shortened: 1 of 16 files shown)


olleolleolle (approved):
This fixes an issue in a fun way. Thanks!

PR #2639, opened 2026-06-10
feat: replace dotenv-rails with mise for env management

mroderick:
The local development setup is a bit clunky for me still, because it assumes that Postgres is running on the default port.

I do other work on the same machine, and would like to at least be able to specify which Postgres port to use.

I've been having good results with using [`mise`](https://mise.jdx.dev) in my day job, so I am proposing that we use that (or set environment variables manually for those that prefer that), which allows us to also drop `dotenv` and `dotenv-rails`. I prefer to have fewer dependencies to manage.

---

Replace `dotenv-rails` + `.env` with `mise` for environment variable management. Add `DB_PORT` support to `database.yml` for running multiple Postgres instances.

### Changes
- **Create** `mise.toml` — shared env defaults (`DB_HOST`, `DB_PORT`)
- **Create** `mise.local.toml.example` — template for local secrets (native install)
- **Create** `docker-compose.override.yml.example` — template for Docker users
- **Add** `DB_PORT` support to `config/database.yml`
- **Add** `mise.local.toml` and `docker-compose.override.yml` to `.gitignore`
- **Remove** `dotenv-rails` gem
- **Update** `AGENTS.md` and `README.md` with new setup workflows

### Migration for Contributors
**Native:**
1. `brew install mise` and add `mise activate` to shell profile
2. `cp mise.local.toml.example mise.local.toml` and fill in GitHub OAuth creds
3. `bundle install` (removes dotenv-rails)
4. (Optional) `rm .env`

**Docker:**
1. `cp docker-compose.override.yml.example docker-compose.override.yml`
2. Fill in GitHub OAuth creds
3. No `.env` file needed

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,7 +87,7 @@ group :development do
 end
 
 group :development, :test do
-  gem 'dotenv-rails'
+
   gem 'fabrication'
   gem 'faker'
   gem 'irb' # LOCKED: Added because of byebug
--- a/Gemfile
+++ b/Gemfile
@@ -87,7 +87,6 @@ group :development do
 end
 
 group :development, :test do
-
   gem 'fabrication'
   gem 'faker'
   gem 'irb' # LOCKED: Added because of byebug
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@
 
 TAGS
 .env
+mise.local.toml
+docker-compose.override.yml
 .DS_Store
 
 .byebug_history
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ TAGS
 .env
 mise.local.toml
 docker-compose.override.yml
+docker-compose.env
 .DS_Store
 
 .byebug_history
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -27,9 +27,21 @@ codebar planner is a Rails 8.1 application for managing [codebar.io](https://cod
 
 **IMPORTANT**: Always use native installation with `bundle exec` commands. Never use Docker or `bin/d*` commands.
 
+### Prerequisites
+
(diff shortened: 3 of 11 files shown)


Gemfile
```diff
@@ -87,7 +87,7 @@ group :development do
 end
 
 group :development, :test do
-  gem 'dotenv-rails'
+
```
olleolleolle:
```suggestion
```

olleolleolle (approved):
I believe it is easier to teach mise than the previous system.

till (approved):
Not a fan of toml 🫣, I prefer direnv. But I agree this looks easier than what was there before.

olleolleolle:
Another, that we use at work is [Shadowenv](https://shopify.github.io/shadowenv/).

mroderick:
**Suggestion:** Add `DB_PORT` to `docker-compose.env` for parity with the `mise.toml` setup.

The PR adds `DB_PORT` to `config/database.yml` and `mise.toml` for native installs, but the Docker `web` service currently relies on the implicit Postgres default. For consistency across both setups, could we add `DB_PORT=5432` to `docker-compose.env`?

```diff
# docker-compose.env
 DB_HOST=db
+DB_PORT=5432
 DB_USER=postgres
 POSTGRES_PASSWORD=password
```

This keeps the `web` service explicitly aligned with the same env var contract used in `database.yml`, and makes it clearer to anyone overriding the file that the port is configurable. The internal container port stays `5432` (the `db` service maps to host `5433` to avoid clashing with native Postgres), so no changes to the `db` ports section are needed.

What do you think?

till:
Maybe get rid off docker-compose.env? No need to maintain 2 files.

olleolleolle:
Nice that we got improvements in, good discussion.

till:
I would merge this as is for now?

What I meant with the docker stuff is, that you can re-use mise instead of duplicating it again. But I wouldn't go down this rabbit hole now.

Issue #535, opened 2017-03-20
Security warning in Firefox


danielquinn:
The home page warns about insecure elements on the homepage:

![codebar](https://cloud.githubusercontent.com/assets/218205/24103489/578f4188-0d77-11e7-9840-ec3300a30966.png)

Environment:
* Firefox 52 64-bit
* Ubuntu Linux 16.10, GNOME 3.x.

BPScott:
Looks like things loaded from assets.codebar.io are always loaded over HTTP. We should update that to be either protocol-less or always load over HTTPS.

danielquinn:
👍 for always over HTTPS.  Moar encryption == moar better :-)  http://codebar.io/ appears to redirect to https:// anyway, so there likely isn't a case where you'll be pulling from an unencrypted place anyway.

matyikriszta:
@danielquinn this is now fixed, thanks for bringing this to our attention. Could you verify it works as expected in your browser? This is what I get in FF now 

<img width="679" alt="screen shot 2017-08-02 at 20 23 44" src="https://user-images.githubusercontent.com/5873816/28890970-d4f8fffc-77c0-11e7-946d-f2b924b9f379.png">

danielquinn:
Yup.  Looks good to me :-)

matyikriszta:
excellent, closing this issue 👍

PR #2641, opened 2026-06-11
fix: skip CSRF protection for feedback form submission

mroderick:
## Problem

We are seeing `ActionController::InvalidAuthenticityToken` errors when students try to submit feedback. This has occurred 82 times in production (Rollbar #535).

The root cause is that the feedback form requires a session cookie for CSRF protection, but browsers like Safari withhold the session cookie when they classify the request as cross-site. This happens when:

- The user navigates to the feedback link from a third-party app (Mail, Slack, etc.)
- Safari Intelligent Tracking Prevention (ITP) is active
- The browser sets `Sec-Fetch-Site: cross-site`

## Why the fix is safe

The feedback form already has its own security mechanism: a unique, unguessable secret token in the URL (e.g. `/feedback/QecUjt60K2J94mBc169Myw/submit`). The `FeedbackRequest` is looked up by this token. An attacker would need to know the token to submit the form, which is exactly the same bar as CSRF protection.

This is the standard "one-time secret link" pattern — the token itself is sufficient to prevent CSRF.

## Changes

- `skip_forgery_protection only: :submit` in `FeedbackController`
- New controller specs covering:
  - Showing the feedback form with valid/invalid tokens
  - Submitting feedback with valid data
  - Submitting feedback with invalid data (renders form with errors)
  - Submitting feedback without a session cookie / CSRF token (the regression test)

## Rollbar issue

https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/535#detail

### DIFF
--- a/app/controllers/feedback_controller.rb
+++ b/app/controllers/feedback_controller.rb
@@ -1,4 +1,6 @@
 class FeedbackController < ApplicationController
+  skip_forgery_protection only: :submit
+
   def show
     feedback_request = FeedbackRequest.find_by(token: params[:id], submited: false)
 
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Thanks for giving good context to the fix!

PR #2648, opened 2026-06-15
fix: stabilise flaky TomSelect test helper

mroderick:
## Description

Fixes a flaky test in `Managing meeting invitations` that intermittently failed in CI with:

```
expected to find css ".ts-dropdown .option" but there were no matches
```

## Root cause

The `SelectFromTomSelect` test helper had two issues:

1. **`ensure_tom_select_initialized` hardcoded `meeting_organisers`** as the element ID to manually initialize. This helper is called from pages using `meeting_invitations_member` (meeting invitations) and `meeting_organisers` (meeting edit form). When called with the wrong ID, the manual init targeted a non-existent element, creating a race condition with the real init code.

2. **`wait: 5` for `.ts-dropdown .option`** was too short for CI environments where the TomSelect AJAX search (`/admin/members/search`) can take longer than 5 seconds to respond.

## Changes

- **Removed `ensure_tom_select_initialized`** — This was a fragile workaround that tried to manually re-initialize TomSelect if the real init hadn't run yet. The real init (jQuery handler in `application.js` running after CDN script loads) is reliable when given enough time. Manual init with a hardcoded element ID only made things worse.

- **Increased timeout for `.ts-dropdown .option`** from 5s to 10s — Gives CI more time for the AJAX search round-trip.

- **Increased `.ts-wrapper` wait** from 10s to 15s — Extra margin for the CDN script to load and TomSelect to initialize in slow CI environments.

- **Same cleanup applied to `remove_from_tom_select`** for consistency.

## Test Plan

- [x] `bundle exec rspec` — 1023 examples, 0 failures
- [x] `managing_meeting_invitations_spec.rb` passes 5/5 consecutive runs locally
- [x] `meeting_spec.rb` passes (also uses `select_from_tom_select`)

### DIFF
--- a/spec/support/select_from_tom_select.rb
+++ b/spec/support/select_from_tom_select.rb
@@ -7,11 +7,10 @@ module SelectFromTomSelect
   # @param item_text [String] The text to select
   # @param from [String, Symbol] The field ID (for documentation purposes)
   def select_from_tom_select(item_text, from: nil)
-    # Ensure TomSelect is initialized (workaround for pages where it doesn't auto-init)
-    ensure_tom_select_initialized('meeting_organisers')
-
-    # Wait for TomSelect to initialize - give extra time for page to fully load JS
-    expect(page).to have_css('.ts-wrapper', wait: 10)
+    # Wait for TomSelect to initialize - the real initialization runs via the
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
🚀

PR #2651, opened 2026-06-18
fix: prevent flaky member search test from matching leaked data

mroderick:
Fixes two pre-existing CI flakes that cause intermittent test failures.

## Fix 1: Member search controller spec

The member search test used generic partial queries (`Jan`, `john@tes`) that could match data leaked from other tests in the same CI group. Changed to use full, specific queries (`Jane Doe`, `john@test.com`) so only the test fixture matches.

## Fix 2: TomSelect helper debounce timing

The `select_from_tom_select` helper types characters in two batches (first 3, then the rest) with a sleep after the first batch. But there was **no sleep after the final keystrokes**, so TomSelect's `loadThrottle: 300` had not yet fired the AJAX request when the test checked for `.ts-dropdown .option`.

- Added `sleep 0.5` after typing remaining characters
- Increased `.ts-dropdown .option` wait timeout from 10s to 15s

### DIFF
--- a/spec/controllers/admin/members_controller_spec.rb
+++ b/spec/controllers/admin/members_controller_spec.rb
@@ -22,7 +22,7 @@ RSpec.describe Admin::MembersController, type: :controller do
 
     context 'with query 3 or more characters' do
       it 'returns matching members by name' do
-        get :search, params: { q: 'Jan' }, format: :json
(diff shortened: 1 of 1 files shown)


KimberleyCook (approved):
🎉

PR #2655, opened 2026-06-19
Update volunteer section on codebar website

KimberleyCook:
We're slightly amending how we announce **Volunteer Opportunities** at codebar. I have created a read only Google Doc that has a list of specific tasks people can help us with.

My thought with an external doc is that we can easily add/remove opportunites as they arise without needing to update and deploy the website


KimberleyCook:
@mroderick This is now ready for review, thank you

mroderick:
I'm not a fan of having this in a separate Google doc, but we can always iterate on this

mroderick (approved):

PR #2657, opened 2026-06-21
Remove unused MeetingTalk model and table

mroderick:
## Summary

Remove `MeetingTalk` — a model that's been disconnected from the application since **February 2016** (over 10 years).

| Change | Files |
|--------|-------|
| Deleted | `app/models/meeting_talk.rb`, `spec/models/meeting_talk_spec.rb` |
| Dropped table | New migration `DropMeetingTalks` removes `meeting_talks` |

## History

| Date | Commit | Event |
|------|--------|-------|
| Dec 2013 | `ba99d318` | **Introduced** — added alongside `Meeting` model with `has_many :meeting_talks` |
| Dec 2013 | `4f95f73d` | Added `default_scope -> { order('title asc') }` — last meaningful code change |
| Feb 2016 | `2b773701` | **Disconnected** — `has_many :meeting_talks` removed from `Meeting` model, replaced with Rolify permissions. No controller, view, or route for talks was ever added. |
| Dec 2017 | `6d302dbb` | Rubocop auto-format |
| Jul 2023 | `7afff142` | Upgraded to `ApplicationRecord` |
| Aug 2025 | `5e879db4` | Added `optional: true` to associations to fix test failures |

## Proof it's unused

- **No controller actions** — `MeetingTalk` has never had a controller. No CRUD, no admin UI.
- **No views** — no partials, no templates render talk data.
- **No routes** — no routes reference meeting_talks.
- **No associations** — `Meeting` no longer has `has_many :meeting_talks` (removed in 2016).
- **No service/job references** — zero references in `app/services/`, `app/jobs/`, `lib/tasks/`.
- **No application code creates it** — `grep -rn "MeetingTalk" app/` returns only the model file itself.

The only code referencing `MeetingTalk` is its own spec, which validates `belongs_to` presence constraints — constraints that can never be satisfied at runtime since no code creates records.

## Migration

```ruby
def up
  drop_table :meeting_talks
end

def down
  create_table :meeting_talks do |t|
    t.references :meeting, index: true
    t.string :title
    t.string :description
    t.text :abstract
    t.integer :speaker_id, index: true
    t.timestamps
  end
end
```

## Verification

- `bundle exec rspec spec/models/ spec/services/ spec/jobs/` — 394 examples, 0 failures
- Full test suite passes (364 examples, 0 failures; 2 pre-existing `PendingMigrationError` from race condition in parallel test runner)


KimberleyCook (approved):

db/migrate/20260621050948_drop_meeting_talks.rb
```diff
@@ -1,12 +1,15 @@
-class CreateMeetingTalks < ActiveRecord::Migration[4.2]
-  def change
+class DropMeetingTalks < ActiveRecord::Migration[8.1]
```
olleolleolle, line 1: This PR: We can probably keep the old migration, even if unused. EDIT: Done

Near future: And, perhaps we can prune ancient migration files, and rely on a current schema.rb.

olleolleolle (commented):
Good find!

olleolleolle (approved):

db/migrate/20260621050948_drop_meeting_talks.rb
mroderick: Thank you, @olleolleolle

PR #2662, opened 2026-06-22
refactor: extract InvitationManager from models to services

mroderick:
## Motivation

`InvitationManager` lived under `app/models/` but was not a domain model — it orchestrates sending invitation emails across multiple mailers (`WorkshopInvitationMailer`, `VirtualWorkshopInvitationMailer`, `MeetingInvitationMailer`, `EventInvitationMailer`) and manages invitation logging. Its logic was split across two files via a code-slicing concern (`WorkshopInvitationManagerConcerns`, 177 lines).

- Domain layer (`app/models/`) should not call mailers directly — that's an upward dependency to infrastructure
- Application layer (`app/services/`) is the correct home for orchestration logic
- Concerns should extract shared *behaviour*, not group all methods of one class

## Changes

- **Create** `app/services/invitation_manager.rb` — single file combining both former model and concern
- **Delete** `app/models/invitation_manager.rb`
- **Delete** `app/models/concerns/workshop_invitation_manager_concerns.rb`
- **Move** `spec/models/invitation_manager_*` → `spec/services/`

The class name (`InvitationManager`) is unchanged, so all callers (3 controllers, 3 rake tasks) need no updates. Rails autoloads from `app/services/` automatically.

## Testing

1042 examples, 0 failures (744 unit + 298 feature).


olleolleolle (approved):
Neatly extracted.

PR #2663, opened 2026-06-22
refactor: extract contact mailing list callbacks to service

mroderick:
## Motivation

`Contact` model had two `after_commit` callbacks that called `Services::MailingList` (Flodesk API) and `ContactMailer` directly — infrastructure concerns embedded in the domain layer via operation callbacks (score 1/5).

- Domain layer (`app/models/`) should not call mailers or external API services via callbacks
- Application layer (`app/services/`) is the correct home for orchestration
- Operation callbacks should be explicit, not implicit side effects of save

## Changes

- **Create** `app/services/contact_mailing_list_service.rb` — orchestrates Flodesk subscribe/unsubscribe + email notification
- **Remove** both `after_commit` callbacks and `#subscribe_to_mailing_list`, `#unsubscribe_from_mailing_list`, `#mailing_list` methods from `Contact` model
- **Keep** `before_create :set_token` (transformer callback, score 5/5)
- **Update** `Admin::SponsorsController#create` and `#update` — call `ContactMailingListService.sync(contact)` explicitly after save
- **Update** `ContactPreferencesController#update` — same explicit call

### Edge case fix (separate commit)

`nil` `mailing_list_consent` previously mapped to unsubscribe (same as `false`). Changed to a no-op using explicit `case` — only `true` subscribes, only `false` unsubscribes.

## Testing

1045 examples, 0 failures (3 new service specs covering true/false/nil paths).

### DIFF
--- a/app/controllers/admin/sponsors_controller.rb
+++ b/app/controllers/admin/sponsors_controller.rb
@@ -30,6 +30,7 @@ class Admin::SponsorsController < Admin::ApplicationController
     authorize @sponsor
 
     if @sponsor.valid? && @sponsor.save
+      @sponsor.contacts.each { |c| ContactMailingListService.sync(c) }
       return redirect_to [:admin, @sponsor], notice: "Sponsor #{@sponsor.name} created"
     end
 
@@ -49,6 +50,7 @@ class Admin::SponsorsController < Admin::ApplicationController
 
     if @sponsor.valid? && @sponsor.save
       @sponsor.contacts.each do |contact|
+        ContactMailingListService.sync(contact)
         audit_contact_subscription(contact)
       end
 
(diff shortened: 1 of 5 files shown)


app/controllers/admin/sponsors_controller.rb
```diff
@@ -30,6 +30,7 @@ def create
     authorize @sponsor
 
     if @sponsor.valid? && @sponsor.save
+      @sponsor.contacts.each { |c| ContactMailingListService.sync(c) }
```
olleolleolle: Opportunity to use: `it` as the block parameter. (Since Ruby 3.4.)


```suggestion
      @sponsor.contacts.each { ContactMailingListService.sync(it) }
```

olleolleolle (approved):

PR #2664, opened 2026-06-23
refactor: extract subscription mailing list callbacks to service

mroderick:
## Motivation

`Subscription` model had `after_create` and `after_destroy` callbacks calling `Services::MailingList` (Flodesk API) directly — infrastructure concerns embedded in the domain layer via operation callbacks (score 1/5). This is the same pattern fixed in #2663 for `Contact`.

- Domain layer (`app/models/`) should not call external API services via callbacks
- Application layer (`app/services/`) is the correct home for orchestration
- Operation callbacks should be explicit, not implicit side effects of save/destroy

## Changes

- **Create** `app/services/subscription_mailing_list_service.rb` — `subscribe` and `unsubscribe` class methods delegating to `Services::MailingList`
- **Remove** `require 'services/mailing_list'`, `after_create`, `after_destroy`, and both private callback methods from `Subscription` model
- **Update** `SubscriptionsController#create` — call `SubscriptionMailingListService.subscribe` after save
- **Update** `SubscriptionsController#destroy` — call `SubscriptionMailingListService.unsubscribe` before destroy
- **Update** `Admin::MembersController#update_subscriptions` — call `SubscriptionMailingListService.unsubscribe` before destroy

## Testing

1044 examples, 0 failures. New service unit test covers both subscribe and unsubscribe paths. Existing feature specs cover all controller paths.

### DIFF
--- a/app/controllers/admin/members_controller.rb
+++ b/app/controllers/admin/members_controller.rb
@@ -39,6 +39,7 @@ class Admin::MembersController < Admin::ApplicationController
 
   def update_subscriptions
     subscription = @member.subscriptions.find_by(group_id: params[:group])
+    SubscriptionMailingListService.unsubscribe(subscription)
     flash[:notice] = t('.unsubscribe', member: @member.full_name,
                                        chapter: subscription.group.chapter.city,
                                        group: subscription.group.name)
--- a/app/controllers/admin/members_controller.rb
+++ b/app/controllers/admin/members_controller.rb
@@ -38,7 +38,7 @@ class Admin::MembersController < Admin::ApplicationController
   end
 
   def update_subscriptions
-    subscription = @member.subscriptions.find_by(group_id: params[:group])
+    subscription = @member.subscriptions.find_by!(group_id: params[:group])
     SubscriptionMailingListService.unsubscribe(subscription)
     flash[:notice] = t('.unsubscribe', member: @member.full_name,
                                        chapter: subscription.group.chapter.city,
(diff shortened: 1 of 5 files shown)


app/controllers/admin/members_controller.rb
```diff
@@ -39,6 +39,7 @@ def events
 
   def update_subscriptions
     subscription = @member.subscriptions.find_by(group_id: params[:group])
```
olleolleolle: This find_by can return nil. Perhaps a find_by! to raise an error rather than continuing? https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21

app/controllers/subscriptions_controller.rb
```diff
@@ -11,6 +11,7 @@ def create
     @subscription = Subscription.new(group_id: group_id, member: current_user)
 
     if @subscription.save
+      SubscriptionMailingListService.subscribe(@subscription)
```
olleolleolle: Minor: not related to this change. Does subscription here NEED to be an instance variable? Does not _seem_ like it? With fewer instance variables, the code becomes easier to reason about.

app/controllers/admin/members_controller.rb
mroderick: Fixed in 111e19d5 — changed to find_by! to get a 404 instead of a 500 on bad params.

app/controllers/subscriptions_controller.rb
mroderick: Fixed in 34c6028c — changed to a local variable in a separate commit.

olleolleolle (approved):
Yay! Looks good to me!

PR #2665, opened 2026-06-23
Fix N+1 queries detected by Bullet

mroderick:
## What

Fixes N+1 queries caught by the Bullet gem during development.

### Changes

- Eager load `organisers` on Workshop and Event queries in the events controller (both upcoming and past views)
- Eager load `organisers` on Meeting queries (was causing N+1, same as Workshop and Event)
- Drop `:permissions` on Meeting queries -- genuinely unused in this view; `includes(:permissions)` adds a preload query without reducing the `organisers` queries (they fire their own SQL join regardless)
- Keep `:venue` on Meeting queries -- all 51 meetings have `venue_id` set in production, and `event.venue` is accessed in `_event.html.haml` line 18-20

### Verification

- [ ] /events/upcoming loads without Bullet::Notification::UnoptimizedQueryError
- [ ] /events/past loads without Bullet::Notification::UnoptimizedQueryError

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -73,13 +73,13 @@ class EventsController < ApplicationController
 
   def fetch_upcoming_events
     events = [
-      Workshop.includes(:chapter, :sponsors, :host, :permissions)
+      Workshop.includes(:chapter, :sponsors, :host, :permissions, :organisers)
               .upcoming
               .joins(:chapter)
               .merge(Chapter.active)
     ]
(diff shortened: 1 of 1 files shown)


Odaeus (approved):
This is totally sensible. 👍🏻
If performance on these pages is an issue you could also look at other options like pulling from a DB view or preemptive caching.

mroderick:
> If performance on these pages is an issue you could also look at other options like pulling from a DB view or preemptive caching.

I also created https://github.com/codebar/planner/pull/2666, which improves performance considerably.

PR #2674, opened 2026-06-29
fix: eliminate flaky address presenter test due to unescaped HTML entities

mroderick:
## Problem

The `AddressPresenter#to_html` spec builds an expected string from Faker-generated data. Faker cities can contain apostrophes (e.g. `O'Connerstad`), and `ERB::Util.html_escape` escapes `'` to `&#39;`. The expected string used raw characters, so the test passed or failed depending on which random city Faker returned — a classic flake.

## Fix

Wrap each component in `ERB::Util.html_escape` when building the expected value, matching what the presenter does internally.

## Verification

Ran the test with 12 different seeds (including the CI seed `32568`). All pass.


mroderick:
This is blocked by #2675

PR #2675, opened 2026-06-29
fix: bump crass 1.0.6 → 1.0.7 for 4 DoS advisories

mroderick:
## Problem

`bundler-audit` fails on `crass 1.0.6` with 4 new advisories (all DoS via deep nesting / large exponents / non-ASCII chars / many adjacent comments). These were added to the advisory DB since the last bundle update.

`crass` is a transitive dependency — `loofah` (pulled in by `rails-html-sanitizer`/ActionView) depends on it with `~> 1.0.2`. We cannot drop it, but the bump is safe and minimal.

## Fix

`bundle update crass --conservative` — crass 1.0.6 → 1.0.7.

## Verification

- `bundler-audit check` — No vulnerabilities found
- `rspec spec/presenters/address_presenter_spec.rb --seed 32568` — passes

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -164,7 +164,7 @@ GEM
     crack (1.0.1)
       bigdecimal
       rexml
-    crass (1.0.6)
+    crass (1.0.7)
     css_parser (2.2.0)
       addressable
     csv (3.3.5)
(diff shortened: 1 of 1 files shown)


olleolleolle (approved):
👍

PR #2685, opened 2026-07-02
perf: targeted row replacement for verify-attendance clicks

mroderick:
## Problem

Taking attendance during workshops was slow. Clicking the verify-attendance checkbox next to an attendee's name triggered a full re-render of the entire attendance panel (~60 rows of HAML for large workshops) plus ~150–200 N+1 queries per click.

## Changes

**Commit 1 — Eager-load attendance_warnings and overrider**
- Updates `with_notes_and_their_authors` scope in `WorkshopInvitation` and `WaitingList` to eager-load `attendance_warnings` and `overrider`
- Eliminates N+1 queries on the initial page load

**Commit 2 — Targeted row replacement for verify-attendance**
- Branches `Admin::InvitationsController#update` to render only the single attendee row when `params[:attended]` is `'true'`
- Skips `set_admin_workshop_data` for single clicks, avoiding loading all students/coaches/waiting lists
- UJS handler replaces only the clicked row via `.closest('.row.attendee')`

**Commit 3 — Unverify attendance**
- Adds `update_to_unattended` controller action (sets `attended: nil`)
- Renders the checked-square as a clickable link with `attended: false`
- Uses the same targeted row replacement as verify

**Commit 4 — Remove unused flash messages**
- `update_to_attended` and `update_to_unattended` no longer return unused strings
- For XHR requests (all verify/unverify clicks), the controller renders the row partial directly and the message was never displayed

## How to verify locally

1. **Ensure the dev server is running**
   ```
   bundle exec rails server
   ```

2. **Seed a testable workshop with attendees**
   In `bundle exec rails console`:
   ```ruby
   chapter = Chapter.first
   workshop = Workshop.create!(
     chapter: chapter,
     date_and_time: 1.hour.ago,
     ends_at: 1.hour.ago + 2.hours,
     coach_spaces: 10,
     student_spaces: 10
   )
   sponsor = Sponsor.first || Sponsor.create!(
     name: 'Test Sponsor',
     website: 'https://example.com',
     seats: 10,
     number_of_coaches: 10
   )
   WorkshopSponsor.create!(workshop: workshop, sponsor: sponsor, host: true)

   members = Member.limit(5)
   members.each_with_index do |m, i|
     WorkshopInvitation.create!(
       workshop: workshop,
       member: m,
       attending: true,
       role: i < 2 ? 'Coach' : 'Student',
       token: SecureRandom.urlsafe_base64,
       tutorial: (i >= 2 ? 'Rails tutorial' : nil),
       automated_rsvp: true
     )
   end
   puts "Workshop #{workshop.id} ready"
   ```

3. **Log in as an admin/organiser**
   Use the seeded member with admin role, or create one via the console.

4. **Visit the workshop admin page**
   ```
   http://localhost:3000/admin/workshops/{workshop_id}
   ```

5. **Click the empty square** next to an attendee's name — it flips to a checked square almost instantly. Click again to unverify. Neither action re-renders the full panel.

## Testing

- 21 examples, 0 failures in related specs
- Verified locally with a 5-attendee workshop


mroderick:
I've verified this locally

app/controllers/admin/invitations_controller.rb
```diff
@@ -38,6 +43,12 @@ def update_to_attended
     "You have verified #{@invitation.member.full_name}’s attendace."
   end
 
+  def update_to_unattended
+    @invitation.update(attended: nil)
+
+    "You have unverified #{@invitation.member.full_name}’s attendace."
```
biggianteye: Where does this text go? I can see similar text in the existing `update_to_attended`, but I don't recall ever seeing that text before.

mroderick: Good catch. The text is returned by `update_attendance` but for XHR requests (all verify/unverify clicks are `remote: true`), the controller renders the row partial directly — the message is never displayed.

```ruby
if request.xhr?
  if params[:attended].present?
    render partial: 'admin/workshop/attendance_row', ...
  else
    ...
  end
else
  redirect_back fallback_location: root_path, notice: message  # only non-XHR
end
```

The message string exists for the non-XHR fallback path. `update_to_attended` had the same pattern in the original code — it returned a message that was unused for XHR but used for the `redirect_back` branch. In my work with an LLM, it preserved that pattern for `update_to_unattended`.

In practice, every verify/unverify click is an XHR request, so neither the 'verified' nor the 'unverified' message is ever shown to the user. The visual feedback is the row HTML itself (square → check-square → square).

With that in mind, I've pushed a commit that removes the messages, since they're never actually used.

app/views/admin/workshop/_attendance_row.html.haml
```diff
@@ -0,0 +1,62 @@
+.row.attendee.mt-3{ id: "attendee-row-#{invitation.id}" }
```
olleolleolle, line 2: Minor: Opportunity to use the Rails Strict Locals declaration here, which is sort of a function arg declaration list. Example blog post mentioning the feature - https://masilotti.com/safer-rails-partials-with-strict-locals/#strict-locals

olleolleolle (approved):
Sounds grand!

(Modernizing the JavaScript falls outside the scope of this change.)

PR #2703, opened 2026-07-14
refactor: inline Permissions concern into Member

mroderick:
## Problem

`Permissions` was a single-model concern (`app/models/concerns/permissions.rb`) only included in `Member` — a code-slicing concern rather than a shared behavioral extraction per layered architecture principles.

## Solution

Inline the `rolify` call and 4 instance methods (`organiser?`, `admin_or_organiser?`, `monthlies_organiser?`, `organised_chapters`) directly into `Member`. Delete the concern file.

The spec was updated from `RSpec.describe Permissions` to `RSpec.describe Member` and renamed from `spec/models/concerns/permissions_spec.rb` to `spec/models/member_permissions_spec.rb`.

## Verification

- 1106 spec examples, 0 failures
- 19 insertions, 29 deletions


KimberleyCook (approved):
Nice 🎉

PR #2721, opened 2026-07-24
Add "Event" to all chapter pages

Agboolafeyikemi:
## Description

Adds an "Add to all" option at the top of the Chapters dropdown on the admin event form, so organisers can invite every chapter with one click instead of ticking each chapter individually.

This implements option 2 from the discussion on the issue (the in-dropdown "Add to all" entry), which @KimberleyCook preferred.

Closes #2490

## Changes

- New `add-all-chapters.js`: injects a bold "Add to all" row at the top of the Chosen dropdown for the chapters field. Clicking it selects every chapter, refreshes the widget, and closes the dropdown.
- Small style block in `main.scss` for the new row (bold, pointer cursor, hover highlight).
- Playwright feature spec covering the flow end-to-end: open the dropdown, click "Add to all", verify the dropdown closes, save, and assert the event is linked to every chapter.

## How to test

1. Sign in as an admin and open `/admin/events/new` (or edit an existing event).
2. Scroll to **Invitation details → Chapters** and click into the field.
3. Click the bold **Add to all** entry at the top of the dropdown.
4. All chapters are selected and the dropdown closes.

Or run the spec:

```
bundle exec rspec spec/features/admin/manage_event_spec.rb
```

PR #2723, opened 2026-07-25
fix(js): use Bootstrap 5 vanilla API for tooltips

mroderick:
Unblocks local verification of #2721.

## Problem

`$('[data-bs-toggle="tooltip"]').tooltip()` is not a function in Bootstrap 5, because the `bootstrap` gem ships vanilla JS rather than jQuery plugins. The resulting `TypeError` aborts the `$(function() {...})` callback in `application.js` before Chosen is initialised, which breaks any feature that depends on Chosen (e.g. the admin event chapters dropdown).

## Change

- Replace the jQuery tooltip call with Bootstrap 5's vanilla API in `app/assets/javascripts/application.js`.
- Update `spec/features/admin/meeting_spec.rb` to assert the organiser avatar's `alt` text instead of the tooltip `title` attribute. Bootstrap 5's Tooltip constructor moves `title` to `data-bs-original-title` and removes the original attribute, so the previous assertion only passed while tooltips were broken.

## Verification

- JS syntax check passes.
- `bundle exec rspec spec/features/admin/meeting_spec.rb --seed 6016` — 9 examples, 0 failures.
- `bundle exec parallel_rspec spec/ -n 6 --only-group 4 --test-options '--seed 6016'` — 175 examples, 0 failures.
- All CI groups pass.

## Related

- Unblocks local verification of #2721.
- Fixes the CI failure where `span[title="..."]` could no longer be found after tooltips were correctly initialised.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -152,5 +152,7 @@ $(function() {
     no_results_text: 'No results matched'
   });
 
-  $('[data-bs-toggle="tooltip"]').tooltip();
+  document.querySelectorAll('[data-bs-toggle="tooltip"]').forEach((el) => {
+    new bootstrap.Tooltip(el);
+  });
 });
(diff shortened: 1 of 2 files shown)


KimberleyCook (approved):
🎉

PR #2722, opened 2026-07-24
Background colour on social links in emails not working

Agboolafeyikemi:
## Description

Inlines the `background-color` on each social link in `shared_mailers/_social.html.haml`, so the buttons in email footers render with their colours regardless of which version of `email.css` gets inlined by premailer.

Closes #2634

## Root cause

Production emails have been inlining a **stale copy of `/assets/email.css`** — the undigested URL that was hard-coded in the mailer header until #2715. codebar.io still serves a pre-November-2024 version of that file (verified byte-for-byte against git history), cached with `Cache-Control: public, max-age=31556952`. That version only has background rules for `a.fb` and `a.tw`, so only the Facebook button gets a colour with today's markup  exactly what the screenshots on #2634 and #2437 show. This is also why the fix in #2437 didn't take effect: it changed the source file, but emails kept inlining the stale cached copy.

Full write-up in [this comment on the issue](https://github.com/codebar/planner/issues/2634#issuecomment-5070822261).

## Why inline styles

Email clients only reliably render inline styles, and critical visual styles in emails shouldn't depend on a separately fetched stylesheet. With the colours inline in the template, the buttons are correct even if `email.css` resolution fails or serves stale content. The CSS rules remain as a fallback.

## How this was tested

- Reproduced the production bug locally: rendered the mailer preview against the actual stale CSS downloaded from codebar.io, only Facebook got a background, matching the issue screenshots.
- Re-rendered with this fix against the **same stale CSS**,  all five buttons get their colours.
- Added a mailer spec asserting the rendered template carries an inline `background-color` for each social link (it asserts pre-premailer output, so it fails if the colours ever move back to being stylesheet-only).
- `bundle exec rspec spec/mailers/workshop_invitation_mailer_spec.rb` 14 examples, 0 failures. RuboCop clean on the changed lines.

PR #2715, opened 2026-07-16
Self-host email assets instead of relying on raw.githubusercontent.com

mroderick:
## Problem

The codebar logo in email templates was loaded from `raw.githubusercontent.com/codebar/assets/master/logo/png/website-logo.png`. This is brittle:
- Reorganising the `codebar/assets` repo silently breaks emails with no alert
- No visibility into failures (no server logs on GitHub's raw CDN)
- Bypasses CloudFlare caching available to the app
- Hardcoded path with no digest-based cache busting

The email CSS link in `_header.html.haml` was also hardcoded as `/assets/email.css` — this would 404 in production where the asset pipeline serves digest-stamped filenames.

## Changes

- **`app/assets/images/email/`** — new folder for email-specific assets, with a README warning contributors not to modify existing files (digest fingerprint changes would break previously-sent emails)
- **`app/assets/images/email/website-logo.png`** — copied from `codebar/assets` repo (5931 bytes, 370x100 PNG)
- **`_body_header.html.haml`** — replaced raw.githubusercontent.com URL with `image_tag("email/website-logo.png")`, removed the TODO
- **`_header.html.haml`** — replaced hardcoded `/assets/email.css` with `asset_path("email.css")` for correct digest resolution
- **`config/environments/production.rb`** — added `config.action_mailer.asset_host` (reads `ASSET_HOST` env var, defaults to `https://codebar.io`)
- **`config/environments/development.rb`** — added `config.action_mailer.asset_host` for local email preview

## Staging setup

Set `ASSET_HOST` env var on the `codebar-staging` Heroku app to its URL (e.g. `https://codebar-staging.herokuapp.com`).

## Other approaches considered (and rejected)

- **CID / data-URI embedding** — poor Gmail and Outlook Windows compatibility; both strip inline attachments and data URIs from images
- **Setting up `assets.codebar.io`** — DNS exists but no origin; unnecessary infra for a single logo
- **Custom Sprockets load path** — tried `app/assets/email/` but would conflict with default `app/assets/stylesheets/email.css`

Issue #2634, opened 2026-06-03
Background colour on social links in emails not working


KimberleyCook:
## Description of the issue 📄

The background colour for the codebar social links in emails is not working.

## Screenshots 📷

<img width="413" height="324" alt="Image" src="https://github.com/user-attachments/assets/b887b049-22da-445f-9055-d2f6a2e3148c" />

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

Agboolafeyikemi:
Hi @KimberleyCook, I'd like to claim this one, I dug into why January's fix (#2437) didn't take effect and found the root cause.

Production emails get their CSS inlined from `/assets/email.css` the URL that was hard-coded in the mailer header until #2715. That URL currently serves a copy of the stylesheet from **before November 2024** (I verified it byte-for-byte against git history), with `Cache-Control: public, max-age=31556952`, i.e. cached for a year. That old file only has background rules for `a.fb` and `a.tw` , so with today's markup only the Facebook button gets a colour, and Slack/LinkedIn/Bluesky/YouTube render as white text on no background. Editing `email.css` can never fix this while the stale copy keeps being served, which is why the previous fix didn't help.

#2715's digest-stamped URL should prevent this recurring, but belt-and-braces I've opened a PR that inlines the background colour on each social link directly in the partial email clients only reliably render inline styles anyway, so the buttons no longer depend on which version of `email.css` premailer resolves.

Ops note: it's probably worth purging the CloudFlare cache for `/assets/email.css` and restarting the dynos, since premailer-rails also caches CSS per-URL in-process in production.

gnclmorais:
@Agboolafeyikemi, thank you _so much_ for figuring this one out! 🙇🏻

biggianteye:
<img width="583" height="269" alt="Image" src="https://github.com/user-attachments/assets/afb71286-fce4-4739-8ff2-fee3c16b3d02" />

Winning. 🎉 

(From an email that went out today)

PR #2724, opened 2026-07-25
Embed email stylesheet directly in mailer header

mroderick:
## Problem

Emails rely on premailer fetching `/assets/email.css` from the configured asset host. Even after #2715 switched the link from a hardcoded path to `asset_path("email.css")`, the resolved asset can still be stale or cached incorrectly in production. PR #2722 worked around this by inlining the social-button background colours directly in the template.

## Solution

Move the entire email stylesheet into a `<style>` block inside `app/views/shared_mailers/_header.html.haml`. Premailer inlines these rules at delivery time, so the email carries its own CSS and no longer depends on resolving an external stylesheet.

## Changes

- Embed `app/assets/stylesheets/email.css` in the `<head>` of `shared_mailers/_header`.
- Delete `app/assets/stylesheets/email.css`.
- Remove the per-element inline `background-color` styles from `shared_mailers/_social` (now handled by the embedded `.soc-btn.*` rules).
- Add a shared RSpec example that asserts the social-button colours are inlined, the external stylesheet link is absent, and the embedded `<style>` block is present.
- Include the shared example in the five mailer specs that render `shared_mailers/_social`: `WorkshopInvitationMailer`, `VirtualWorkshopInvitationMailer`, `EventInvitationMailer`, `MeetingInvitationMailer`, and `FeedbackRequestMailer`.
- Update the email assets README.

## Verification

- `bundle exec rspec spec/mailers/` — 75 examples, 0 failures.
- `make test` — 1158 examples, 0 failures.
- `bundle exec haml-lint app/views/shared_mailers/_header.html.haml` — 0 lints.


gnclmorais:
Shouldn’t the `premailer-rails` gem inline our CSS, so we don’t link anything from anywhere (and allowing us to keep CSS in a `.css` file)?

mroderick:
Hi @gnclmorais, that’s a very fair question — I’m not married to embedding the CSS in HAML. I went that route because it was the simplest way to guarantee we never hit the stale-asset problem again.

You’re right that `premailer-rails` is built to inline linked CSS. The issue in #2634 was that production was inlining a year-cached copy of `/assets/email.css` from Cloudflare, which only had rules for Facebook/Twitter. That’s why the previous CSS-only fixes didn’t take effect.

The embedded `<style>` block removes any dependency on resolving that URL, but I agree it’s worse for maintainability.

A middle ground I’d be happy to switch to: keep `app/assets/stylesheets/email.css`, restore the `<link>` in `_header`, and configure `premailer-rails` to only use `:filesystem` and `:asset_pipeline`, removing `:network` from the fallback chain. That would let premailer-rails load the stylesheet from the precompiled manifest without ever making an HTTP request to the asset host, so we keep the CSS in a real `.css` file while avoiding the stale-cache risk.

```ruby
# config/initializers/premailer.rb
Premailer::Rails.config.merge!(
  preserve_styles: true,
  strategies: [:filesystem, :asset_pipeline]
)
```

`app/assets/config/manifest.js` already links all stylesheets, so `email.css` will be precompiled. I’d keep the shared spec that asserts the colours are inlined and that no `/assets/email.css` link appears in the final email.

Would you prefer that approach? Happy to update the PR if so.

gnclmorais:
That sounds great, @mroderick — thank you for providing an alternative 🙇🏻

mroderick:
Closing this in favour of #2747, which uses the strategy we discussed.

PR #2769, opened 2026-07-31
refactor: remove dead full_url_for helper and all call sites

mroderick:
## Summary

`full_url_for` was dead code. The method concatenated `#{@host}#{path}`, but `@host` was never set in any mailer. All callers already passed `*_url` helpers (`invitation_url`, `event_invitation_url`, `unsubscribe_url`, etc.) which return fully-qualified URLs via `config.action_mailer.default_url_options`.

## History

`full_url_for` was introduced in October 2013 (`6fb19ab5`) alongside the first invitation emails, with the apparent intent to prepend a host to relative paths. However, `@host` was never assigned in any mailer method across the entire project history — it has been dead code from day one.

Rails' `*_url` helpers (e.g., `invitation_url`, `unsubscribe_url`) have always been the actual mechanism for generating URLs in emails, and they produce fully-qualified URLs because `config.action_mailer.default_url_options` is configured in all environments (`localhost:3000` in dev/test, `codebar.io` in production).

## Changes

- **Delete** `app/helpers/email_helper.rb` — the module is now empty
- **Remove** `include EmailHelper` / `helper EmailHelper` from:
  - `VirtualWorkshopInvitationMailer`
  - `WorkshopInvitationMailer`
  - `ContactMailer`
- **Remove** inline `helper do` blocks defining `full_url_for` from:
  - `FeedbackRequestMailer`
  - `EventInvitationMailer`
  - `MeetingInvitationMailer`
- **Replace** `full_url_for(...)` with `...` in all 21 mailer view call sites across 14 template files

## Verification

All 75 mailer specs pass. Controller, feature, and policy specs also verified.

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -150,11 +150,6 @@ Rails/HasManyOrHasOneDependent:
     - 'app/models/workshop.rb'
     - 'app/models/workshop_invitation.rb'
 
-# Offense count: 1
-Rails/HelperInstanceVariable:
-  Exclude:
-    - 'app/helpers/email_helper.rb'
-
 # Offense count: 2
 # Configuration parameters: IgnoreScopes.
 Rails/InverseOf:
--- a/app/helpers/email_helper.rb
+++ b/app/helpers/email_helper.rb
@@ -1,7 +1,0 @@
-module EmailHelper
-  private
-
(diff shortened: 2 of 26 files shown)


olleolleolle (approved):
Good find! 🧹

PR #2777, opened 2026-08-03
chore(migrations): remove pre-2026 migrations and rely on schema.rb

mroderick:
## What

Removes all 163 pre-2026 migrations and relies on `db/schema.rb` + `db:prepare` for fresh databases. The 13 in-window 2026 migrations remain as the only files in `db/migrate/`.

## Why

Following the practice discussed in #2774: once migration files have been applied to all environments, `db/schema.rb` is the source of truth and the historical files can be removed. This keeps only the migrations in the strong_migrations window (2026 onwards).

## Changes

- Deleted 163 pre-2026 migration files.
- Heroku release step changed from `bundle exec rake db:migrate` to `bundle exec rake db:prepare` so a fresh dyno loads from `db/schema.rb` and an existing dyno runs pending migrations.
- Updated dev setup instructions in `AGENTS.md`, `README.md`, and `docs/development-setup.md` to use `db:prepare`.
- Added a note in `AGENTS.md` explaining the migration cleanup.

## Verification

- `bundle exec rake db:create db:prepare` from a scratch test database reproduces the committed `db/schema.rb` exactly.
- `db:migrate:status` on the scratch DB shows the 13 kept 2026 migrations all `up`.
- Full test suite: `bundle exec parallel_rspec spec/ -n 3` — 1172 examples, 0 failures.

## Dependencies

- Requires #2776 (merged), which removed unbacked columns from `db/schema.rb` so the schema can be loaded cleanly.


olleolleolle (approved):
Grand!

PR #2780, opened 2026-08-04
fix(auth): handle concurrent OAuth callback creating the same auth service

mroderick:
## Problem

A double-fired OAuth callback can pass the initial `AuthService.find_by(provider:, uid:)` lookup concurrently (both see nothing), then both enter the create branch and collide while `member.save!` saves the nested auth service. The unique `(uid, provider)` index (already live in prod) plus the auth_service uniqueness validation make one request error — `RecordNotUnique` on a true TOCTOU race, or `RecordInvalid` via the app validation when the duplicate is already visible. Result: an errored login request.

## What this does

In `auth_services_controller#create`, wrap `member.save!` in a rescue of both `RecordInvalid` and `RecordNotUnique`, and on collision reuse the already-created auth service:
- `AuthService.find_by!(provider:, uid:)` acts as the **guard** — if no winner exists this isn't a race, so it raises the real error instead of swallowing it.
- `can_log_in` is only enabled on **actual creation** (`created`) — `update_column(:can_log_in, true) if created` — so the losing callback doesn't touch the winner's flag (set, not toggle; see #2783).

## Verification
- New request spec (`spec/requests/auth_services_callback_spec.rb`) deterministically exercises the rescue and asserts the loser reuses the winner's member/service and does not touch `can_log_in`.
- Rebases cleanly onto current master (resolves the overlap with #2778's `can_log_in` change).
- Rubocop clean; request spec + signup feature specs (login / ToC / newsletter / joining): 20 examples, 0 failures.

## Why a separate PR from #2779
The `auth_services (uid, provider)` index already shipped to prod in the first partial run of the unique-index migration, so this fix is independently urgent and not gated on #2779 — it gets focused auth review on its own clock.

### DIFF
--- a/spec/requests/auth_services_callback_spec.rb
+++ b/spec/requests/auth_services_callback_spec.rb
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-RSpec.describe 'AuthServices callback', type: :request do
+RSpec.describe 'AuthServices callback' do
   it 'reuses the winner when a concurrent callback just created the same auth service' do
     winner = Fabricate(:member, email: 'winner@example.com')
     winner_service = Fabricate(:auth_service, member: winner,
--- a/spec/requests/auth_services_callback_spec.rb
+++ b/spec/requests/auth_services_callback_spec.rb
@@ -1,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe 'AuthServices callback', type: :request do
+  it 'reuses the winner when a concurrent callback just created the same auth service' do
+    winner = Fabricate(:member, email: 'winner@example.com')
+    winner_service = Fabricate(:auth_service, member: winner,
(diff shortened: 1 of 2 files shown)


spec/requests/auth_services_callback_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe 'AuthServices callback', type: :request do
```
olleolleolle: This type should be auto-recognized and added by RSpec configuration - not added explicitly.

```suggestion
RSpec.describe 'AuthServices callback' do
```

mroderick: Thank you ❤️

olleolleolle (approved):

PR #2785, opened 2026-08-05
feat(member): replace can_log_in with derived active? and drop the column

### LINKED ISSUE #2783: can_log_in is silently toggled off when a member links a second auth service — 115 members affected
## Summary

`can_log_in` doesn't mean "can this user log in". A `toggle!` in the auth callback makes it equivalent to the **parity of a member's auth-service count**: odd count → `true`, even count → `false`. Every time a member links a new login provider, the flag silently flips. This disables a member's login and breaks the `requires_additional_details?` gating that decides whether the profile "details" page is shown.

## Root cause

In `AuthServicesController#create`, the branch that handles a first-time login with a new provider does:

```ruby
member.save!
member.toggle!(:can_log_in)   # flips the flag on EVERY new provider link
...
redirect_to edit_member_details_path(member_type:)
```

`toggle!` inverts the boolean instead of setting it to `true`. So:
- A member's **first** auth service flips `false → true` (the intended "enable login").
- Linking a **second** service flips it back `true → false` (silently disabling login).
- A third would flip it on again, and so on.

Because `can_log_in` gates the `requires_additional_details?` checks used by `finish_registration` (and the member name/surname/email/about-you validations), the corrupted flag has two user-visible effects: an otherwise-complete member can be redirected to `/member/details/edit` for no reason on the first login with a new provider, and the flag flip otherwise silently disables their account.

### Reported repro

A test of the new codebar auth produced this sequence:

1. Login via GitHub option → redirected to `/member/details/edit` (member is already complete).
2. Log out, log in again via codebar → logged in with no details page.

The second login "self-correcting" is a symptom of the flag having been flipped to `false`, not an actual fix.

## Blast radius (from the production database)

`can_log_in` values across **29,366** members are exactly predicted by auth-service-count parity:

| Auth services | Members | `can_log_in = true` | `can_log_in = false` |
|---|---|---|---|
| Odd count (1 or 3) | 29,112 | 29,111 | 1 (non-parity anomaly) |
| Even count (2) | 115 | 0 | **115** |

- **115 members** currently have `can_log_in = false` purely because they hold 2 auth services (even count). Parity predicts all 115 exactly.
- **110 of those** got there from multiple GitHub services (pre-dates codebar).
- **5 of those** are the github + codebar cohort added during the new-auth rollout (member IDs 4473, 23289, 2195, 28910, 8668) — all otherwise complete, ToC-accepted members.
- The single odd-count/`false` member (ID 485) is a **non-parity anomaly** — it does not fit the `count.odd?` model, so something other than the toggle set it. It could be a deliberate disable, or a member who had a second service unlinked after a toggle (leaving the flag `false`). The database cannot distinguish these, so confirm intent before treating it as an exclusion.
- The 3 codebar-only accounts are `true` (new members: toggle went the intended off→on way).

So `can_log_in` is essentially `auth_services.count.odd?`, i.e. an accidental XOR of how many providers a member has accumulated, with zero relationship to whether they should be able to log in.

Because 110 of the 115 affected members got there by linking a **second GitHub** service through this same `toggle!` path long before codebar existed, this is **not a new-auth regression** — it is a long-standing parity bug that new auth only surfaced in a handful more accounts.

## How to detect

Run against the planner database. The first query finds every member whose `can_log_in` doesn't match the parity model (the corrupted set is the even-count rows); the second lists the actual affected members:

```sql
-- 1. Summary: does can_log_in track auth-service-count parity?
--    Expect: odd -> true, even -> false, modulo any deliberate lockouts.
WITH counts AS (
  SELECT m.id, count(*) AS svc, m.can_log_in
  FROM members m JOIN auth_services a ON a.member_id = m.id
  GROUP BY m.id, m.can_log_in
)
SELECT
  count(*) FILTER (WHERE svc % 2 = 1 AND can_log_in)     AS odd_svc_can_log_in,
  count(*) FILTER (WHERE svc % 2 = 1 AND NOT can_log_in) AS odd_svc_disabled,
  count(*) FILTER (WHERE svc % 2 = 0 AND can_log_in)     AS even_svc_enabled,
  count(*) FILTER (WHERE svc % 2 = 0 AND NOT can_log_in) AS even_svc_disabled
FROM counts;

-- 2. List affected members: even auth-service count with can_log_in = false,
--    excluding known deliberate lockouts (e.g. member_id 485).
WITH counts AS (
  SELECT m.id, m.name, m.surname, m.can_log_in, count(*) AS svc,
         string_agg(a.provider, ',') AS providers
  FROM members m JOIN auth_services a ON a.member_id = m.id
  GROUP BY m.id
)
SELECT id, can_log_in, svc, providers
FROM counts
WHERE svc % 2 = 0 AND NOT can_log_in
  AND id <> 485   -- non-parity anomaly; keep the exclusion list explicit
ORDER BY id;
```

The first query should return `even_svc_enabled = 0` (parity is fully predictive) and `odd_svc_disabled > 0` only for intentional lockouts. The second query is the list to review before repairing: confirm none of the rows are deliberate disables, then set `can_log_in = true` for the rest.

## Fix

Three-part:

1. **Do not toggle — set.** In the new-provider branch, replace `member.toggle!(:can_log_in)` with `member.update!(can_log_in: true)` so linking another provider can never disable the account.
2. **Only ask for details when they're actually needed.** The `redirect_to edit_member_details_path` immediately after is unconditional; gate it on the member actually being incomplete:

```ruby
redirect_to edit_member_details_path(member_type:) if member.requires_additional_details?
```

3. **Add regression coverage.** `AuthServicesController#create` currently has no spec coverage (the spec file only exercises `GET #new`), which is why this shipped and survived. Add a spec asserting that linking a *second* provider leaves `can_log_in` unchanged, and that a complete member is **not** redirected to `/member/details/edit`.

### Data repair for existing affected members

Set `can_log_in = true` for the 115 even-count members, **excluding deliberate lockouts**. Confirm with the team whether any existing disabled accounts (including member 485, whose `false` is a non-parity anomaly of unconfirmed intent) are intentional before flipping them back.

## Notes

- `PR #2778` fixes part 1 (`update!(can_log_in: true)`), but does **not** address part 2 — the unconditional details redirect remains, so a complete member still lands on `/member/details/edit` the first time they link a new provider.
- Note on PR #2778: its **body text** says the change is `toggle!` → `toggle(...).save!` (which would still flip the flag), but its **actual diff** is `member.update!(can_log_in: true)` — the correct non-flipping fix. Trust the diff, not the PR description. PR #2778 is still open at time of writing, so master still carries the bug.

### LINKED ISSUE #2784: can_log_in is stale, misnamed, and wrong — replace with a derived method
## Summary

`members.can_log_in` is a misnamed, denormalized boolean. It is **never** checked in the login path, its only real consumer (`requires_additional_details?` → `finish_registration`) runs only after a member has already logged in via an auth service, and #2783 showed its value has degraded to `auth_services.count.odd?` (parity garbage) across 116 members. The column is stale state that duplicates easily-derived truth. Proposal: replace it with a derived method and drop the column.

## What `can_log_in` actually does

Complete app-code reference set:

1. `app/models/member.rb:30,32` — presence validations for `name/surname/email/about_you` + strict email, gated `if: :can_log_in?`.
2. `app/models/member.rb:128` — `requires_additional_details? = can_log_in? && !valid?`, consumed by `finish_registration` (`app/controllers/application_controller.rb:95`).
3. `app/controllers/auth_services_controller.rb:49` — the `toggle!(:can_log_in)` that corrupts it (#2783).
4. `db/schema.rb` + a migration.
5. Specs: `spec/models/member_spec.rb`, `spec/features/member_joining_spec.rb`.

Login never consults it: `current_user = Member.find(session[:member_id])` (`app/controllers/application_controller.rb:31-36`), with the session set from a matched `AuthService`. There is no `can_log_in` check anywhere in the authentication path.

## Why the flag is redundant

`requires_additional_details?` is only ever evaluated after login, and login requires an auth service. So the `can_log_in? &&` prefix is always true at the point of use — the flag contributes nothing, it is dead weight that is also corrupted.

Live DB check (`codebar_dump`):

| stored `can_log_in` | has auth svc | members |
|---|---|---|
| false | no | 112 |
| true | no | 27 |
| false | yes | 116 |
| true | yes | 29,111 |

The 112 and 27 members have no auth service, so they cannot log in regardless — treating them as inactive matches reality. The 116 (including the 5 codebar + 110 github-github cohort from #2783, plus 485) all have an auth service and can log in; deriving "active" from `auth_services.exists?` makes them active without any data repair.

## Git history

`can_log_in` has been misnamed since its inception (2013-12-08, `54a7edbe` "Added github auth, no tests"): the name, `toggle!`, `requires_additional_details?`, and the gated validations all landed together, and login never checked it even then. Hours later `b3918f14` "Fix already created members" changed `Member.new` → `Member.find_or_create_by_email`, letting a new provider reuse an existing member — which turned `toggle!` into a flip and is the origin of #2783's parity corruption. No commit in `git log -S'can_log_in'` ever used it as a login/access gate. So both the misleading name and the bug predate codebar; the codebar rollout only added 5 victims and produced the report. (Note: `7a604b18`, the recent codebar sign-in switch, preserved the toggle unchanged.)

## Proposed change: derive, don't store

1. Replace the stored boolean with a derived method on `Member`:
   - `def active? = auth_services.exists?`
   - `def requires_additional_details? = active? && !valid?`
   (`active?` is equivalent to the old gate in the live post-login path, but is defensive if `requires_additional_details?` is ever called without a session.)
2. Update the consumers:
   - `app/models/member.rb:30,32` — `if: :active?` in place of `if: :can_log_in?`
   - `app/models/member.rb:128` — use `active?`
   - `app/controllers/auth_services_controller.rb:49` — remove the `toggle!`; a member is active by having an auth service.
   - `app/controllers/auth_services_controller.rb:52` — gate the otherwise-unconditional details redirect on `requires_additional_details?` so a complete member linking a second provider is not bounced to `/member/details/edit`:
     ```ruby
     redirect_to edit_member_details_path(member_type: member_type) if member.requires_additional_details?
     ```
3. Drop the `can_log_in` column via migration. **No data repair is needed** — the 116 self-correct to active because they have auth services.
4. If genuine account lockout is ever required, add a real gate (e.g. `disabled_at`) that is checked at login — `can_log_in` never locked anyone out, so nothing to preserve.

### Alternative (minimal): rename only

If the team prefers not to remove the column, rename it to something honest (e.g. `active`). Note this leaves a stale, re-corruptible column in place; it fixes the mislabel but not the redundant-stored-state root cause.

## Relationship to #2783

This change **fully resolves #2783**. It removes the `can_log_in` corruption at the root (part 1), gates the unconditional details redirect (part 2), and makes the 115-row data repair unnecessary (the 116 self-correct to active). PR #2778 (already merged) changed the write site to set the flag to `true` instead of toggling; this issue builds on that by dropping the column entirely — the remaining part 2 (redirect gating) is still open and is addressed here.

## Relationship to #2780 (not obsoleted)

PR #2780 (`fix/auth-callback-race`) fixes a **different** bug: a concurrent OAuth callback collides on the unique `(uid, provider)` index during `member.save!` and errors the request. It is needed **regardless** of this issue — #2784 does not touch the race handling. The only overlap is the `can_log_in` write in the same else-branch, which #2784 removes entirely.

**Ordering:** #2780 currently writes to the `can_log_in` column, so it should land first, then this issue drops the column on top (or #2780 is rebased on this issue to drop its now-dead `can_log_in` write, keeping only the race guard). Both must ship; the two fixes are orthogonal and complementary, not competing.

## Out of scope / dependency

- Whether the 116 rows should be treated as active → answered implicitly by this change (they are), so no separate repair step.
- Whether the team wants a genuine hard-lockout mechanism → a future, separate concern (`disabled_at`), not introduced here.

mroderick:
## Summary

Issue #2784. Replace the stored `members.can_log_in` boolean with a derived `Member#active?` and drop the column; stop writing `can_log_in` in the OAuth callback and gate the details redirect on `requires_additional_details?`.

`can_log_in` was never a login gate. Its only consumer, `requires_additional_details?` (`app/models/member.rb`), gates profile completion after login, and issue #2783 showed the value had degraded into a parity artifact across 116 members. A member is active by having an auth service, so the flag duplicates easily-derived truth — this removes the column and derives it.

Key behaviours:
- New members keep skipping the profile presence/email validations at signup: during `member.save!` in the callback the just-built `auth_service` is not yet persisted, so `active?` (`auth_services.exists?`) is false; the member becomes `active?` after save.
- A complete member who links a second provider is no longer bounced to `/member/details/edit` — they go to `referer_or_dashboard_path` instead. This closes the unconditional-redirect part of #2783.
- Migration follows the strong_migrations ignore-then-drop contract (`self.ignored_columns += ['can_log_in']` precedes the `safety_assured` drop). No data repair is needed: the 116 affected members self-correct to active because they have auth services.

Fixes #2784 · Resolves #2783

## Base

Rebased onto `master`. #2780 (OAuth race fix) is merged, so this PR contains only the #2784 work; no longer stacked on the race-fix branch.

## Verification

- `51 + 12` RSpec examples pass (member, callback, controller, joining, ToC specs); rubocop clean.
- ce-code-review (correctness, testing, data-migration): no blocking findings. Advisory P3 items applied (#3 active? false-branch model coverage, #4 referer redirect) and recorded; residual risks noted for the destructive-DDL rollback and the `session[:referer_path]` staleness (pre-existing).

- Deployment note: the migration drops `members.can_log_in`. Per the strong_migrations contract, ship the code containing `ignored_columns` in the same release as the migration.
- Rollback: `remove_column` is reversible (re-adds the column) but cannot restore dropped values; treat the column as obsolete (impact nil).
- After deploy, validation window ~1 week. Watch for `NoMethodError` / undefined-column errors on `can_log_in` in Rollbar (would indicate old code running against the dropped column — i.e. a missed two-phase deploy).
- Healthy signal: no errors; `Member#requires_additional_details?` behaves as before for new-signup and second-provider flows.
- No additional runtime monitoring required beyond the above; the change is entirely derive/drop with no new background or scheduled work.

### DIFF
--- a/app/controllers/auth_services_controller.rb
+++ b/app/controllers/auth_services_controller.rb
@@ -44,10 +44,8 @@ class AuthServicesController < ApplicationController
           uid: omnihash[:uid]
         )
 
-        created = false
         begin
           member.save!
-          created = true
         rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotUnique
           # Concurrent OAuth callback for the same account: another request just
           # created this member/auth_service (both email and (uid, provider) are
@@ -59,18 +57,16 @@ class AuthServicesController < ApplicationController
           member = member_service.member
         end
 
-        # Set, not Toggle: toggling would flip can_log_in off for a member who
-        # links a second auth service. Skip the conditional profile validations
-        # here on purpose — a brand-new member completes name/about_you on the
-        # next (details) page, and running them now aborts the signup callback.
-        member.update_column(:can_log_in, true) if created # rubocop:disable Rails/SkipsModelValidations
-
         session[:member_id]          = member.id
         session[:service_id]         = member_service.id
         session[:oauth_token]        = omnihash[:credentials][:token]
         session[:oauth_token_secret] = omnihash[:credentials][:secret]
 
-        redirect_to edit_member_details_path(member_type: member_type)
+        if member.requires_additional_details?
+          redirect_to edit_member_details_path(member_type: member_type)
+        else
+          redirect_to referer_or_dashboard_path
+        end
     end
   end
 
--- a/app/controllers/auth_services_controller.rb
+++ b/app/controllers/auth_services_controller.rb
@@ -63,7 +63,7 @@ class AuthServicesController < ApplicationController
         session[:oauth_token_secret] = omnihash[:credentials][:secret]
 
         if member.requires_additional_details?
-          redirect_to edit_member_details_path(member_type: member_type)
+          redirect_to edit_member_details_path(member_type:)
         else
           redirect_to referer_or_dashboard_path
         end
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -1,4 +1,6 @@
 class Member < ApplicationRecord
+  self.ignored_columns += ['can_log_in']
+
   include DigestHelper
 
   rolify role_cname: 'Permission', role_table_name: :permission, role_join_table_name: :members_permissions
@@ -27,9 +29,9 @@ class Member < ApplicationRecord
   has_many :member_email_deliveries
 
   validates :auth_services, presence: true
-  validates :name, :surname, :email, :about_you, presence: true, if: :can_log_in?
+  validates :name, :surname, :email, :about_you, presence: true, if: :active?
   validates :email, uniqueness: true
-  validates :email, email: { mode: :strict }, if: :can_log_in?
+  validates :email, email: { mode: :strict }, if: :active?
   validates :about_you, length: { maximum: 255 }
 
   DIETARY_RESTRICTIONS = %w[vegan vegetarian pescetarian halal gluten_free dairy_free other].freeze
@@ -125,7 +127,11 @@ class Member < ApplicationRecord
   end
 
   def requires_additional_details?
-    can_log_in? && !valid?
+    active? && !valid?
+  end
+
+  def active?
+    auth_services.exists?
   end
 
   def existing_rsvp_on?(date)
--- a/db/migrate/20260806000000_remove_can_log_in_from_members.rb
+++ b/db/migrate/20260806000000_remove_can_log_in_from_members.rb
@@ -1,0 +1,5 @@
+class RemoveCanLogInFromMembers < ActiveRecord::Migration[8.1]
+  def change
+    safety_assured { remove_column :members, :can_log_in, :boolean }
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_07_31_104506) do
+ActiveRecord::Schema[8.1].define(version: 2026_08_06_000000) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "pg_catalog.plpgsql"
 
@@ -441,7 +441,6 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_31_104506) do
   create_table "members", id: :serial, force: :cascade do |t|
     t.string "about_you"
     t.datetime "accepted_toc_at", precision: nil
-    t.boolean "can_log_in", default: false, null: false
     t.datetime "created_at", precision: nil
     t.enum "dietary_restrictions", default: [], array: true, enum_type: "dietary_restriction_enum"
     t.string "email"
--- a/spec/features/member_joining_spec.rb
+++ b/spec/features/member_joining_spec.rb
@@ -16,7 +16,6 @@ RSpec.feature 'A new student signs up', type: :feature do
 
   scenario 'A visitor must fill in all mandatory fields in order to sign up' do
     member = Fabricate(:member, name: nil, surname: nil, email: nil, about_you: nil, how_you_found_us: nil, how_you_found_us_other_reason: nil)
-    member.update_attribute(:can_log_in, true)
     login member
 
     visit edit_member_details_path
@@ -68,7 +67,6 @@ RSpec.feature 'A new student signs up', type: :feature do
     group = Fabricate(:group)
 
     login member
-    member.update(can_log_in: true)
 
     visit step2_member_path
     expect(page).to have_current_path(step2_member_path)
--- a/spec/models/member_spec.rb
+++ b/spec/models/member_spec.rb
@@ -7,8 +7,10 @@ RSpec.describe Member do
       it { is_expected.to validate_length_of(:about_you).is_at_most(255) }
       it { is_expected.to validate_uniqueness_of(:email) }
 
-      describe 'can_log_in' do
-        let(:member) { Fabricate.build(:member, can_log_in: true) }
+      describe 'active' do
+        # Fabricated members have a persisted auth service, so #active? is true
+        # and the profile presence/email validations run.
+        let(:member) { Fabricate(:member) }
 
         it { expect(member).to validate_presence_of(:name) }
         it { expect(member).to validate_presence_of(:surname) }
@@ -44,6 +46,32 @@ RSpec.describe Member do
         it 'accepts email with plus addressing' do
           member.email = 'user+tag@example.com'
           expect(member).to be_valid
+        end
+
+        context 'when the member has no auth service' do
+          let(:member) do
+            member = Fabricate(:member)
+            member.auth_services.delete_all
+            member.name = nil
+            member.surname = nil
+            member.about_you = nil
+            member
+          end
+
+          it 'is not active' do
+            expect(member.active?).to be(false)
+          end
+
+          it 'skips the gated profile presence validations' do
+            member.valid?
+            expect(member.errors[:name]).to be_empty
+            expect(member.errors[:surname]).to be_empty
+            expect(member.errors[:about_you]).to be_empty
+          end
+
+          it 'does not require additional details' do
+            expect(member.requires_additional_details?).to be(false)
+          end
         end
       end
     end
--- a/spec/requests/auth_services_callback_spec.rb
+++ b/spec/requests/auth_services_callback_spec.rb
@@ -17,10 +17,45 @@ RSpec.describe 'AuthServices callback' do
 
     expect { post '/auth/github/callback' }.not_to raise_error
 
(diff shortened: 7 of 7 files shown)


mroderick:
## Closes the account-duplication confusion from the 28 Jul `/auth/codebar` feedback

This PR resolves the reported issue where an existing member saw the
**"Thanks for signing up"** flash and the **"Almost there..."** details page
(with a pre-populated form) the first time they logged in via a new provider —
which read as if a duplicate account had been created.

**Why it happened:** in `AuthServicesController#create`'s `else` branch (member
found by email, new auth service built), we unconditionally
`redirect_to edit_member_details_path`. That page's controller flashes
`notifications.signing_up` ("Thanks for signing up...") and its view is titled
"Almost there..." (`member/details/edit.html.haml`). A complete, existing member
hit exactly this path, hence the signup messaging on a returning account.

**What this PR changes:** the redirect is now gated behind
`member.requires_additional_details?` (`active? && !valid?`), so a complete
member who links a second provider goes to `referer_or_dashboard_path` instead
of being bounced to profile completion — they never see the signup flash or the
"Almost there..." page. (The existing `sends a complete member...` specs cover
this.) It also drops the `can_log_in` write entirely, eliminating the
set-vs-toggle bug on that flag.

**Follow-up:** an *incomplete* existing member is still sent to the details
page and still gets the "Thanks for signing up" flash, which remains misleading
for a returning account. Tracked in #2788.

app/controllers/auth_services_controller.rb
```diff
@@ -59,18 +57,16 @@ def create
...
-        member.update_column(:can_log_in, true) if created # rubocop:disable Rails/SkipsModelValidations
-
         session[:member_id]          = member.id
         session[:service_id]         = member_service.id
         session[:oauth_token]        = omnihash[:credentials][:token]
         session[:oauth_token_secret] = omnihash[:credentials][:secret]
 
-        redirect_to edit_member_details_path(member_type: member_type)
+        if member.requires_additional_details?
+          redirect_to edit_member_details_path(member_type: member_type)
```
olleolleolle: Very minor: omission of hash values when they have the same name as the key is possible:


```suggestion
          redirect_to edit_member_details_path(member_type:)
```

olleolleolle (approved):
Nice.

app/controllers/auth_services_controller.rb
mroderick: Thank you! ❤️

PR #2789, opened 2026-08-06
fix: only show signup flash to newly created members

### LINKED ISSUE #2788: Do not show "Thanks for signing up" flash to existing members completing their profile
## Problem

The "Thanks for signing up" flash (`notifications.signing_up`, shown in
`Member::DetailsController#edit`) is misleading for returning members. It reads
as if a brand-new (duplicate) account was created, when in fact the member was
simply routed to the "Almost there..." details page because their profile is
incomplete.

This is a residual of the 28 Jul `/auth/codebar` feedback and remains after
#2785 (which gates the details redirect on
`member.requires_additional_details?`). A complete member no longer hits this
path, but an *incomplete* existing member (missing any of
name/surname/about_you) still does — and still gets told "Thanks for signing
up", which is wrong for a returning account.

## Proposed fix

Only flash `signing_up` when the member was actually created during the OAuth
callback, not when an existing member is sent to the details page.

In `AuthServicesController#create`, the `else` branch already tracks whether the
member was created (`member.save!` succeeded on a new record). Pass that signal
through so a find-existing redirect doesn't set the signup flash. The details
controller should show "Thanks for signing up" / "Almost there..." only for
genuinely new members; existing members should get a neutral message (e.g.
"Please complete your profile") or none.

## Acceptance criteria

- [ ] An existing, incomplete member logging in via a new provider is sent to
      the details page **without** the "Thanks for signing up" flash.
- [ ] A brand-new member still sees the "Thanks for signing up" flash.
- [ ] Specs cover both cases (extend `auth_services_callback_spec.rb` / the
      details feature spec).

mroderick:
## Summary

Returning members routed to the "Almost there..." details page (because their profile is incomplete) were still told "Thanks for signing up", even though no new account was created. The signup flash now shows only for members actually created during the OAuth callback; existing members land on the details page with no misleading message.

## Test plan

Extended `auth_services_callback_spec.rb` and the details controller spec to cover both cases. The browser feature test for a genuine new signup still shows the flash. 21 examples, 0 failures; RuboCop clean.

Fixes #2788

### DIFF
--- a/app/controllers/auth_services_controller.rb
+++ b/app/controllers/auth_services_controller.rb
@@ -35,6 +35,7 @@ class AuthServicesController < ApplicationController
 
         member = Member.find_by(email:)
         member ||= Member.new(email:)
+        new_member = member.new_record?
 
         member.name    ||= omnihash[:info][:name]&.split(' ')&.first || ''
         member.surname ||= omnihash[:info][:name]&.split(' ')&.drop(1)&.join(' ') || ''
@@ -55,6 +56,7 @@ class AuthServicesController < ApplicationController
           member_service = AuthService.find_by!(provider: omnihash[:provider],
                                                 uid: omnihash[:uid])
           member = member_service.member
+          new_member = false
         end
 
         session[:member_id]          = member.id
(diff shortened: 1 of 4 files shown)


KimberleyCook (approved):
🎉

PR #2791, opened 2026-08-07
Fix WorkshopsController#rsvp 500 when member already has a nil-attending invitation

### LINKED ISSUE #2790: WorkshopsController#rsvp crashes with 'No route matches ... id: nil' when member already has a workshop invitation
## Summary

A member who already has a WorkshopInvitation for a workshop + role (with `attending` left NULL) gets a 500 when they click **Attend as a coach/student** again on the workshop page.

Error:

```
ActionController::UrlGenerationError: No route matches {action: "show", controller: "workshop_invitation", id: nil}, missing required keys: [:id]
  app/controllers/workshops_controller.rb:29:in `rsvp'
    redirect_to invitation_path(@invitation)
```

Occurrences (from production error tracking) — database ids only, no user data:

- member 23754, workshop 3787, role Coach — 3 occurrences 2026-08-06
- member 24648, workshop 3761, role Student — 2 occurrences 2026-08-04

## Root cause

1. The member already has a `workshop_invitations` row for that workshop+role with `attending` NULL (e.g. auto-created when organisers invite chapter subscribers via `invitation_manager.rb:create_invitation`, or a prior RSVP that was never completed on the invitation page). Verified in the DB: id 1962613 (member 23754 / workshop 3787 / Coach) and id 1964312 (member 24648 / workshop 3761 / Student), both `attending` NULL.
2. The workshop page shows the **Attend as a coach/student** button because the guard renders it when `attendee?` is false (`app/views/workshops/_actions.html.haml`), and `attendee?` only matches invitations with `attending: true` (`app/models/workshop.rb:89`).
3. The `user_attending_or_waitlisted?` guard in `WorkshopsController#rsvp` passes (no accepted invitation, no waiting-list row), so it calls `find_or_create_invitation`.
4. `find_or_create_invitation` uses `WorkshopInvitation.create_or_find_by(workshop:, member:, role:)`. In Rails 8.1, `create_or_find_by` uses the **non-bang `create`**. The model's `validates :member_id, uniqueness: { scope: %i[workshop_id role] }` fails (row already exists), so `create` returns an **invalid, unsaved** record with `id = nil`. `create_or_find_by` only rescues `RecordNotUnique`, not the validation failure, so it returns that record instead of the pre-existing one.
5. `redirect_to invitation_path(@invitation)` then raises the UrlGenerationError because `@invitation.id` is nil.

## How to verify the problem exists

**Reproduction against a local `codebar_dump` database** (requires a user who already has a `workshop_invitations` row for the target workshop+role with `attending` NULL):

```ruby
# reproduce create_or_find_by returning an unsaved record
ActiveRecord::Base.establish_connection(adapter: 'postgresql', database: 'codebar_dump', host: 'localhost', port: 5432, username: '', password: '')
member = Member.find(23754)
workshop = Workshop.find(3787)
invitation = WorkshopInvitation.create_or_find_by(workshop: workshop, member: member, role: 'Coach')
invitation.id          # => nil  (unsaved!)
invitation.persisted?  # => false
invitation.valid?      # => false
invitation.errors.full_messages  # => ["Member has already been taken"]

Rails.application.routes.url_helpers.invitation_path(invitation)
# => ActionController::UrlGenerationError: No route matches {action: "show", controller: "workshop_invitation", id: nil}
```

**HTTP reproduction:** log in as a member who already has a `workshop_invitations` row for a workshop+role with `attending` NULL, visit the workshop page, and click **Attend as a coach/student**. The request `POST /workshops/:id/rsvp?role=Coach` returns a 500 with the error above.

## Proposed fix (for a separate implementation session)

In `find_or_create_invitation` (`app/controllers/workshops_controller.rb`), return the persisted existing row when creation hits the duplicate-validation path, e.g.:

```ruby
def find_or_create_invitation(workshop, user, role)
  invitation = WorkshopInvitation.create_or_find_by(workshop: workshop, member: user, role: role)
  invitation.persisted? ? invitation : WorkshopInvitation.find_by(workshop: workshop, member: user, role: role)
end
```

Keeping `create_or_find_by` preserves its concurrency safety for the create path; the fallback handles the already-exists case.

## How to verify the fix is resolved

1. Run the reproduction above against the same data and confirm `find_or_create_invitation` now returns the **existing persisted** row (id 1962613 / 1964312) rather than an unsaved record, and that `invitation_path(invitation)` produces a valid `/invitation/<token>` route (no raise).
2. Log in as a member who already has a `workshop_invitations` row with `attending` NULL, click **Attend as a coach/student**, and confirm the request now redirects (302) to their existing invitation page instead of 500ing.
3. Add a controller/request spec for `WorkshopsController#rsvp` covering this case — there is currently **no** controller/request spec for `rsvp`, which is why this went uncaught. The spec should assert that posting `rsvp` with a `role` for a member who already holds an invitation for that workshop+role redirects and does not raise.

## Notes

- The no-role branch of `rsvp` ("Manage your invitation") can hit the same nil-redirect if `find_attending_invitation` ever returns nil, but that branch's button only renders when `attendee?` is true, so it is not the trigger for the observed occurrences.

mroderick:
Closes #2790

## Summary

Fixes a 500 error when a member clicks **Attend as a coach/student** on a workshop where they already have a `WorkshopInvitation` row for that workshop + role with `attending` NULL.

## Root cause

`find_or_create_invitation` used `WorkshopInvitation.create_or_find_by`. In Rails 8.1, `create_or_find_by` calls the non-bang `create`. When the uniqueness validation fails (row already exists), it returns an unpersisted record with `id: nil`. The subsequent `redirect_to invitation_path(@invitation)` then raises `ActionController::UrlGenerationError`.

## Fix

After `create_or_find_by`, check `persisted?`. If the record is not persisted, fall back to `WorkshopInvitation.find_by(workshop:, member:, role:)` to return the existing row.

## Verification

- Added `spec/controllers/workshops_controller_spec.rb` covering:
  - Existing nil-attending invitation for the workshop + role redirects to the existing invitation page.
  - No new invitation is created in that case.
  - Member without an invitation still creates one.
  - Already-attending member is redirected back with the existing message.
- Ran the new spec and all controller/request specs:
  - `bundle exec rspec spec/controllers/workshops_controller_spec.rb` — 4 examples, 0 failures
  - `bundle exec rspec spec/controllers/ spec/requests/` — 136 examples, 0 failures
- `bundle exec rubocop app/controllers/workshops_controller.rb spec/controllers/workshops_controller_spec.rb` — no offenses

- Monitor error tracking for `ActionController::UrlGenerationError` with `controller: "workshop_invitation", id: nil` in `WorkshopsController#rsvp`.
- Expected signal: zero occurrences after deploy.
- If occurrences persist, verify the member/workshop/role combination and check for invitations that cannot be found by `find_by` (e.g. unique index vs validation mismatch).

### DIFF
--- a/app/controllers/workshops_controller.rb
+++ b/app/controllers/workshops_controller.rb
@@ -44,9 +44,10 @@ class WorkshopsController < ApplicationController
   end
 
   def find_or_create_invitation(workshop, user, role)
-    WorkshopInvitation.create_or_find_by(workshop: workshop,
-                                         member: user,
-                                         role: role)
+    invitation = WorkshopInvitation.create_or_find_by(workshop: workshop,
+                                                      member: user,
+                                                      role: role)
+    invitation.persisted? ? invitation : WorkshopInvitation.find_by(workshop: workshop, member: user, role: role)
   end
 
   def user_attending_or_waitlisted?(workshop, user)
--- a/app/controllers/workshops_controller.rb
+++ b/app/controllers/workshops_controller.rb
@@ -44,10 +44,10 @@ class WorkshopsController < ApplicationController
   end
 
   def find_or_create_invitation(workshop, user, role)
-    invitation = WorkshopInvitation.create_or_find_by(workshop: workshop,
+    invitation = WorkshopInvitation.create_or_find_by(workshop:,
                                                       member: user,
-                                                      role: role)
-    invitation.persisted? ? invitation : WorkshopInvitation.find_by(workshop: workshop, member: user, role: role)
+                                                      role:)
+    invitation.persisted? ? invitation : WorkshopInvitation.find_by(workshop:, member: user, role:)
   end
 
   def user_attending_or_waitlisted?(workshop, user)
--- a/spec/controllers/workshops_controller_spec.rb
+++ b/spec/controllers/workshops_controller_spec.rb
@@ -1,0 +1,50 @@
+RSpec.describe WorkshopsController, type: :controller do
+  let(:member) { Fabricate(:member) }
+  let(:workshop) { Fabricate(:workshop) }
+
+  before { login(member) }
+
+  describe 'POST #rsvp' do
+    context 'when the member already has an invitation for the workshop and role with attending nil' do
+      let!(:invitation) do
+        Fabricate(:workshop_invitation, workshop: workshop, member: member, role: 'Coach', attending: nil)
+      end
+
+      it 'redirects to the existing invitation page' do
+        post :rsvp, params: { id: workshop.id, role: 'Coach' }
+
+        expect(response).to redirect_to(invitation_path(invitation))
+      end
+
+      it 'does not create a new invitation' do
+        expect do
+          post :rsvp, params: { id: workshop.id, role: 'Coach' }
+        end.not_to change(WorkshopInvitation, :count)
+      end
+    end
+
+    context 'when the member does not have an invitation for the workshop and role' do
+      it 'creates a new invitation and redirects' do
+        expect do
+          post :rsvp, params: { id: workshop.id, role: 'Coach' }
+        end.to change(WorkshopInvitation, :count).by(1)
+
+        invitation = WorkshopInvitation.last
+        expect(response).to redirect_to(invitation_path(invitation))
+      end
+    end
+
+    context 'when the member is already attending' do
+      before do
+        Fabricate(:attending_workshop_invitation, workshop: workshop, member: member, role: 'Coach')
+      end
+
+      it 'redirects back with already wish to attend message' do
+        post :rsvp, params: { id: workshop.id, role: 'Coach' }
+
+        expect(response).to redirect_to(root_path)
+        expect(flash[:notice]).to eq(I18n.t('workshops.already_wish_to_attend'))
+      end
+    end
+  end
+end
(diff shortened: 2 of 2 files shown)


KimberleyCook (approved):

spec/controllers/workshops_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
+RSpec.describe WorkshopsController, type: :controller do
```
olleolleolle:
```suggestion
RSpec.describe WorkshopsController do
```

app/controllers/workshops_controller.rb
```diff
@@ -44,9 +44,10 @@ def find_attending_invitation(workshop, user)
   end
 
   def find_or_create_invitation(workshop, user, role)
-    WorkshopInvitation.create_or_find_by(workshop: workshop,
-                                         member: user,
-                                         role: role)
+    invitation = WorkshopInvitation.create_or_find_by(workshop: workshop,
+                                                      member: user,
+                                                      role: role)
+    invitation.persisted? ? invitation : WorkshopInvitation.find_by(workshop: workshop, member: user, role: role)
```
olleolleolle:
```suggestion
    invitation = WorkshopInvitation.create_or_find_by(workshop:,
                                                      member: user,
                                                      role:)
    invitation.persisted? ? invitation : WorkshopInvitation.find_by(workshop:, member: user, role:)
```

olleolleolle (approved):
Probably helps!

PR #2801, opened 2026-08-09
Add self check-in for events and workshops

mroderick:
It is vitally important for codebar to keep accurate attendance records, here are some arguments why.

- In case of trouble at or after an event, we need to have a record of who attended that we can share with the hosts or law enforcement 
- When we speak to sponsors and create our impact report, we need accurate data
- When we ask students for feedback, we need to know who attended as a coach

The PR allows organisers to take some of the workload of keeping attendance and let's members check themselves in.

Hopefully, this will lead to better attendance keeping and less work for organisers during events and workshops.

## What

Adds a self check-in flow for codebar events and workshops.

image

image

image

Admittedly, the PDF could be improved. I think this is OK for a first iteration and allows us to get it in the hands of organisers for a field test or two.

## Changes

- Stores a `check_in_code` on events/workshops and a `source` on invitations.
- Adds a `CheckInCode` concern for code generation and window/capacity helpers.
- Generates memorable check-in codes from the EFF long wordlist (7776 words); source documented in `lib/words/README.md`.
- Adds an admin check-in page with a QR-code PDF download.
- Adds a public check-in page where members sign in and select Student/Coach.
- Enforces event-time window, capacity, waiting list, and single-role rules.
- Records `source` for admin and check-in invitations.

## Manual verification

1. Make yourself an admin: `Member.find_by(email: "you@example.com").add_role(:admin)`.
2. Start the server and sign in.
3. Open an upcoming event or workshop admin page and click **Check in**.
4. Download the PDF and confirm it shows the event details, a QR code, and the check-in URL.
5. Open the check-in URL in an incognito/private window (or sign out) and sign in as a different member.
6. Select **Student** or **Coach** and submit.
7. Return to the admin page and confirm the member appears as attended.
8. Try checking in again with the same member and a different role — it should reject the role switch.
9. Try checking in outside the event window (before 1 hour before start or after `ends_at`) — it should be blocked.

## Testing

- Full suite passes: 1247 examples, 0 failures.
- RuboCop and haml-lint clean on changed files.


KimberleyCook:
This change is so so awesome! however in terms of reviewing the code I'm not sure I'm the best person 👀 

@matyikriszta if you have some time would you be able to take a look?

olleolleolle (approved):
A couple of things that I saw, nothing blocking,

- the name parent, perhaps we can figure one out that's more intent-revealing?
- the place where we picked between two routes (e/w for event and workshop), perhaps it could be just an if/else with concrete routes and not use public_send?
- where we picked up check_in_code, assign a local and use it, that would fit the Ruby "omit hash value if key matches the name" Hash thing.

<3

PR #2803, opened 2026-08-10
perf(sponsors): add indexes and fragment caching for sponsors page

mroderick:
Adds two PostgreSQL indexes and per-sponsor fragment caching to the sponsors page to address the slow `SponsorsController#index` reports.

**Database indexes**

`SponsorsController#index` loads active sponsors ordered by `updated_at`, and `SponsorsSearch` (admin sponsors list) orders by `lower(sponsors.name)`. Both queries previously used a sequential scan plus an explicit sort.

Local `EXPLAIN ANALYZE` with ~1,160 rows:

| Query | Before | After |
|---|---|---|
| Active sponsors by updated_at | Seq Scan + Sort, ~1.45 ms | Index Scan, ~0.31 ms |
| Admin order by lower(name) | Seq Scan + Sort, ~2.86 ms | Index Scan, ~0.05 ms |

**View allocations**

Scout APM showed some requests hitting ~5M allocations, 95% in view code. Profiling the sponsor partial showed CarrierWave creating uploader/version/sanitized-file objects for every sponsor on every render.

Splitting the sponsor partial into a single-sponsor partial and rendering the collection with `cached: true` drops warm-cache allocations for the partial by ~80% (from ~225k objects to ~44k objects locally). Cache keys are based on each sponsor's `updated_at`, so edits invalidate automatically.

The indexes are added concurrently with `if_not_exists: true`, so the migration is safe to re-run.


olleolleolle (approved):
Neat!

PR #2811, opened 2026-08-25
chore(deps): remove unused coffee-script gem

mroderick:
## What

Removes the `coffee-script` gem and its lockfile entries.

## Why

- No `.coffee` files exist in the repo.
- No bundled gem contains `.coffee` files.
- The Gemfile comment above the gem said it was only kept because Sprockets autoloads it when `.coffee` files are present, and that it could be removed when that no longer happens.

## Changes

- Remove `gem 'coffee-script'` and its explanatory comment from `Gemfile`
- Remove `coffee-script` and `coffee-script-source` from `Gemfile.lock`

## Verification

- `bundle exec rubocop Gemfile` — clean
- `bundle exec rails runner 'puts "boot OK"'` — boots
- `bundle exec rspec spec/models/chapter_spec.rb` — 15 examples, 0 failures
- `RAILS_ENV=test bundle exec rails assets:precompile` — succeeds without the CoffeeScript processor

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,10 +4,6 @@ ruby file: '.ruby-version'
 gem 'rails', '~> 8.1.3' # LOCKED: It is Rails.
 # The original asset pipeline for Rails [https://github.com/rails/sprockets-rails]
 gem 'sprockets-rails'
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
🌞

PR #2820, opened 2026-08-29
Replace Chosen with native datalist for sponsor chapter filter

mroderick:
## What

Replaces Chosen.js selects with a reusable `ChapterPickerComponent` that uses native HTML `datalist` + text input with `autocomplete="off"`.

### New: ChapterPickerComponent

A reusable ViewComponent that renders a text input backed by a `<datalist>` with chapter **names** as values:

```ruby
render ChapterPickerComponent.new(
  name: 'sponsors_search[chapter]',
  chapters: @chapters,
  placeholder: 'Filter by chapter'
)
```

The input shows the chapter name the user selects (not an ID), while the controller does a case-insensitive name lookup to resolve it. The `autocomplete="off"` prevents browser autofill from fighting the datalist suggestions.

### Applied to

1. **Sponsors filter** (`admin/sponsors/index`) — replaces `collection_select` with `chosen-select` class
2. **Workshop form** (`admin/workshops/_shared_form`) — replaces `f.association :chapter` select

### SponsorsSearch

Continues to do case-insensitive name lookup for chapter filtering (no change from the existing behaviour — the datalist now submits the name directly instead of an ID).

## Why

Chosen is abandoned. This removes a JavaScript dependency and uses built-in browser autocomplete instead. The component is reusable — any form that needs single-chapter selection can use it.

## Testing

- `bundle exec rspec spec/queries/sponsors_search_spec.rb`
- `bundle exec rspec spec/features/admin/filtering_sponsors_list_spec.rb`
- `bundle exec rspec spec/features/admin/workshops_spec.rb`
- `bundle exec rspec spec/components/chapter_picker_component_spec.rb`


app/components/chapter_picker_component.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ChapterPickerComponent < ViewComponent::Base
```
olleolleolle, line 3: Minor: We should have an ApplicationComponent base class, I think, for an easier control point. It won't change anything right now, but it's neat to have prepared.

spec/components/chapter_picker_component_spec.rb
```diff
@@ -0,0 +1,40 @@
+require 'rails_helper'
+require 'view_component/test_helpers'
+
+RSpec.describe ChapterPickerComponent do
+  include ViewComponent::TestHelpers
```
olleolleolle, line 5: Minor: we should make it so that each spec/components/ test get these by dint of having derived RSpec Metadata about their "type". 

Or, check whether that's already true, and these helpers are already included.

olleolleolle (approved):
I like a good datalist.

app/components/chapter_picker_component.rb
mroderick: I'm new to using ViewComponents, so this might be a silly question. What would this ApplicationComponent give us?

olleolleolle: Just like ApplicationController or ApplicationRecord, a single base class, where you can do shared things like "ah, always use this custom layout", to take an example from ApplicationController.

olleolleolle: But, let's install that thing whenever we need to.

Issue #2823, opened 2026-08-30
Auto-create Students and Coaches groups when a new chapter is created


mroderick:
## Background

Groups were introduced in 2014 (commit e681a268) as a way to organise chapter membership into two categories: `Students` and `Coaches`. Every chapter has conventionally required exactly these two groups. Issue #510 (2016) validated this by restricting `Group::NAMES` to only `%w[Coaches Students]`.

## Problem

Creating a chapter via `/admin/chapters/new` does **not** automatically create the required groups. An admin must then manually visit `/admin/groups/new` twice — once for Students, once for Coaches — and pick the new chapter from a dropdown. If this step is missed:

- Members visiting the chapter page see **no subscription buttons** (`chapter/_subscriptions.html.haml` iterates `@chapter.groups`)
- Workshop invitations **silently send to zero members** (`InvitationManager#chapter_students` / `#chapter_coaches` query empty relations)
- The admin chapter show page shows **no group stats**

This is a foot-gun. The `chapter_with_groups` fabricator already auto-creates groups in tests/seeds (`db/seeds.rb` uses it for every chapter), so the codebase already assumes this pattern.

## Evidence

Production DB (`codebar_production_dump`):
- 54 chapters
- 108 groups  
→ Exactly 2 groups per chapter, confirming the invariant holds in practice.

## Proposed solution

A **service object** in `app/services/chapter_creation_service.rb` that wraps chapter + groups in a single transaction:

```ruby
class ChapterCreationService
  Result = Struct.new(:chapter, :success, :errors, keyword_init: true)

  def self.call(params)
    chapter = Chapter.new(params)

    ActiveRecord::Base.transaction do
      chapter.save!
      chapter.groups.create!(name: 'Students')
      chapter.groups.create!(name: 'Coaches')
    end

    Result.new(chapter: chapter, success: true)
  rescue ActiveRecord::RecordInvalid => e
    Result.new(chapter: chapter, success: false, errors: e.message)
  end
end
```

**Controller change:** `Admin::ChaptersController#create` calls the service:

```ruby
def create
  result = ChapterCreationService.call(chapter_params)
  @chapter = result.chapter
  authorize(@chapter)

  if result.success
    flash[:notice] = "Chapter #{@chapter.name} has been successfully created"
    redirect_to [:admin, @chapter]
  else
    flash[:notice] = @chapter.errors.full_messages
    render 'new'
  end
end
```

**Why a service object:**
[...]

PR #2824, opened 2026-08-30
feat: create Students/Coaches groups on chapter creation

mroderick:
## Problem

Creating a chapter via `/admin/chapters/new` did not automatically create the required Students and Coaches groups. Admins had to manually visit `/admin/groups/new` twice to create them.

This caused:
- No subscription buttons on chapter pages (iterates `@chapter.groups`)
- Workshop invitations silently sent to zero members
- No group stats on admin chapter show page

## Solution

Added `ChapterCreationService` that creates chapter + groups in a single transaction:

- Creates chapter, then Students and Coaches groups
- If groups fail, chapter is rolled back
- `Admin::ChaptersController#create` authorizes the unsaved chapter before calling the service, so a denied request persists nothing
- Controller stays thin, just coordinates

## Changes

- **New:** `app/services/chapter_creation_service.rb` — service object with transaction
- **New:** `spec/services/chapter_creation_service_spec.rb` — unit tests, including a real group-failure rollback test
- **Updated:** `Admin::ChaptersController#create` — authorizes first, then delegates to the service
- **Updated:** feature spec — verifies groups are created; new scenario for the invalid-form failure branch
- **Updated:** controller spec — new denial-path regression spec (a chapter-scoped organiser persists nothing)

## Testing

All tests pass (51 examples, 0 failures). Code-reviewed with ce-code-review: the initial implementation persisted before authorizing (a chapter-scoped organiser would be denied after the commit, leaving phantom data) — fixed and covered by the denial regression spec.

## Follow-up

Once merged, can remove the manual group creation UI (`admin/groups/new`, `Admin::GroupsController#new/#create`, and the admin menu link) as a separate PR.

## Related

[#2823 — Auto-create Students and Coaches groups when a new chapter is created](https://github.com/codebar/planner/issues/2823) (linked only — this PR intentionally does not close the issue; the follow-up that removes the manual group creation UI will)


app/services/chapter_creation_service.rb
```diff
@@ -0,0 +1,17 @@
+class ChapterCreationService
+  Result = Struct.new(:chapter, :success, :errors, keyword_init: true)
```
olleolleolle: Not a blocker, but can use Data.define rather than Struct.new to make this class, I guess.

olleolleolle (approved):
Looks good!

PR #2449, opened 2026-01-29
Email users code

jonodrew:
This code should close #2383 , and form a decent basis for other tasks in that group. 

It collates the users who've not accepted a workshop invite in the last three months and who haven't already been chased, and sends them a chaser email.

Issue #2384, opened 2025-10-20
New automated email to people who have signed up via the website but have not subscribed to a chapter.


KimberleyCook:
## Description of the issue 📄

We would like to send a new automated email to people who have signed up via the website but have not subscribed to a chapter yet (give it a week after signing up).

If you'd like to implement this new feature please drop me a DM on [Slack](https://slack.codebar.io/) for the copy needed.

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

mroderick:
This has been implemented with #2835

Issue #2383, opened 2025-10-20
New automated email for students who have not attended a codebar workshop in 3 months


KimberleyCook:
## Description of the issue 📄

We would like to send an automated email to students who have not attended a codebar workshop in 3 months. 

If you'd like to implement this new feature please drop me a DM on [Slack](https://slack.codebar.io/) for the copy needed.

## To do 📋

* [ ] Claim this issue (comment below, or assign yourself if you are part of the codebar org)
* [ ] Fork and clone the repository
* [ ] Update the relevant files. Follow the steps to fix section in this issue.
* [ ] Commit your changes as one commit. Use the title of this issue as your commit message
* [ ] Submit a pull request
* [ ] Mention this issue in the PR description by including it's number
* [ ] Have your pull request reviewed & merged by a codebar team member

KimberleyCook:
@jonodrew does this now mean people will start receiving the email?? 🤞

jonodrew:
> [@jonodrew](https://github.com/jonodrew) does this now mean people will start receiving the email?? 🤞

We'll need to deploy into production first, but yes, once we've done that!

KimberleyCook:
Woooooop 🎉 this is going to help me SO MUCH!! Thank you everyone for working on it!

PR #2827, opened 2026-08-31
fix: persist delivered email body in MemberEmailDelivery

mroderick:
## Problem

`member_email_deliveries.body` has never captured any content — all 439 rows in production are empty, so the email audit log cannot show what was actually sent.

## Root cause

`premailer-rails` registers a Mail interceptor that runs during delivery: it converts the message to `multipart/alternative` (adding a plain-text part and inlining CSS), which empties the body container. `EmailDelivery#log_sent_email` then runs in an `after_deliver` callback and reads `mail.body.to_s` — the now-empty container — storing `""`.

The emails themselves are unaffected: recipients receive complete `multipart/alternative` messages with both plain-text and HTML parts. Only the audit log is broken. Chaser dedup checks row existence, not body, so it works as before.

## Fix

Read the html part instead of the body container:

```ruby
body: mail.html_part ? mail.html_part.body.to_s : mail.body.to_s,
```

Pure plain-text mailers (no html part) fall back to `mail.body` as before.

## Testing

Extended the `MemberMailer#chaser` logging spec to assert the persisted row has a body containing the template copy. Confirmed red→green: the assertion fails with `""` on master and passes with the fix. Full mailer specs and the three-month chaser service spec pass (85 examples), RuboCop clean.

## Context

Found while diagnosing chaser reliability for #2384 (chaser introduced in #2449 for #2383).

### DIFF
--- a/app/mailers/concerns/email_delivery.rb
+++ b/app/mailers/concerns/email_delivery.rb
@@ -11,7 +11,9 @@ module EmailDelivery
     MemberEmailDelivery.create!(
       member: member,
       subject: mail.subject,
-      body: mail.body.to_s,
+      # premailer-rails' interceptor empties the body container and moves the
+      # content into text/html parts during delivery, so read the html part
+      body: mail.html_part ? mail.html_part.body.to_s : mail.body.to_s,
       to: Array(mail.to),
       cc: Array(mail.cc),
       bcc: Array(mail.bcc)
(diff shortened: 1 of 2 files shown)


app/mailers/concerns/email_delivery.rb
```diff
@@ -11,7 +11,9 @@ def log_sent_email
     MemberEmailDelivery.create!(
       member: member,
       subject: mail.subject,
-      body: mail.body.to_s,
+      # premailer-rails' interceptor empties the body container and moves the
+      # content into text/html parts during delivery, so read the html part
```
olleolleolle: Perhaps we can ditch this comment? 


```suggestion
```

olleolleolle (approved):

PR #2828, opened 2026-08-31
test: wire ViewComponent RSpec helpers via type metadata

mroderick:
Follow-up to #2820, addressing [Olle's comment](https://github.com/codebar/planner/pull/2820#discussion_r3890376455): component specs should get `ViewComponent::TestHelpers` via RSpec type metadata instead of requiring and including it by hand.

While checking, we found that path-based inference for `type: :component` **never actually worked** in this repo. `infer_spec_type_from_file_location!` iterates `DIRECTORY_MAPPINGS` at call time, but `spec/rails_helper.rb` registered the `:component` mapping only on the following line. `event_card_component_spec` masked the problem by declaring `type: :component` explicitly (the view_component generator adds it); the other two component specs patched over the missing helpers with a manual `require` + `include`.

## Changes

- `spec/rails_helper.rb`: register `DIRECTORY_MAPPINGS[:component]` **before** calling `infer_spec_type_from_file_location!`, and include `ViewComponent::TestHelpers` once for `type: :component` specs. ViewComponent 4.x no longer ships its own RSpec auto-wiring (no `lib/view_component/rspec.rb`), so this belongs here.
- `spec/components/*`: drop the per-spec `require 'view_component/test_helpers'` and `include ViewComponent::TestHelpers` boilerplate.

Net −4 lines. All 16 component spec examples pass; rubocop clean.

### DIFF
--- a/spec/components/chapter_picker_component_spec.rb
+++ b/spec/components/chapter_picker_component_spec.rb
@@ -1,9 +1,6 @@
 require 'rails_helper'
-require 'view_component/test_helpers'
 
 RSpec.describe ChapterPickerComponent do
-  include ViewComponent::TestHelpers
-
   let(:chapters) { Fabricate.times(3, :chapter) }
 
   it 'renders a text input with datalist attributes' do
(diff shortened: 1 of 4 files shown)


olleolleolle (approved):
Nice! Thank you!

PR #2829, opened 2026-08-31
chore: remove unused member_type from member_email_deliveries

mroderick:
## Problem

`member_email_deliveries.member_type` is dead weight. It was added in #2449 (same day as the table) for polymorphism that was never built: the model is a plain `belongs_to :member`, the `EmailDelivery` concern never writes it, nothing reads it, and it is `NULL` in all 439 production rows.

## Change

- Drop the column (`safety_assured` — same single-phase pattern as the `can_log_in` removal in `20260806000000`; safe because no code references the column, so old dynos during a rolling deploy cannot break)
- Keep `ignored_columns` on the model to guard stale schema caches, matching the `can_log_in` precedent

If non-`Member` recipients ever need email logging, re-add the column together with a real polymorphic association.

## Testing

Migration applied to dev and test databases; `member_mailer` and `three_month_email_service` specs pass (32 examples, including the chaser logging spec that writes rows through the concern). RuboCop clean.

## Context

Found during the #2384 brainstorm while investigating the table's history.

### DIFF
--- a/app/models/member_email_delivery.rb
+++ b/app/models/member_email_delivery.rb
@@ -1,3 +1,5 @@
 class MemberEmailDelivery < ApplicationRecord
+  self.ignored_columns += ['member_type']
+
   belongs_to :member
 end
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
We thank it and let it go.

PR #2832, opened 2026-08-31
refactor: make member_email_deliveries a typed log

mroderick:
## Problem

`member_email_deliveries` records which member received an email but not *which* email it was, and the three-month chaser's "already emailed" exclusion only works because it is the sole writer of the table — any other mailer action that starts logging would silently change chaser behavior.

## Change

- **Type each row:** new `email_type` column, populated with the mailer action that sent the email (`email_type: action_name` in the `EmailDelivery` concern). The migration backfills existing rows with `'chaser'` (all 439 are chasers) and then drops the column default, so every writer must set the type explicitly
- **Integrity in the database:** unique index on `(member_id, email_type)` — a member can only ever have one chaser row, enforced by Postgres rather than by query convention. Built `:concurrently` in its own migration so the column DDL stays atomic and reversible
- **Scoped exclusion:** the chaser now excludes members via `email_type: 'chaser'` rows only — identical behavior today (it is the only writer), and immune to whatever else gets logged later
- **Replay-tolerant writes:** `find_or_create_by!` so a retried delivery job can't raise on the unique index *after* the email has already gone out (a `create!` would fail the job and cause a re-delivery loop); content columns are written on first creation only

No behavior change for anything that exists today: the only logging mailer action is the chaser, and its eligibility outcome is identical under scoped vs blanket exclusion.

## Testing

- Chaser logging spec now asserts the logged row's `email_type`, and a new spec proves a replayed delivery logs only one row
- Chaser service spec proves a member whose only logged delivery is of another type remains chaser-eligible, while chaser-typed rows still exclude

Built on #2829 + #2830 (both merged).

### DIFF
--- a/app/mailers/concerns/email_delivery.rb
+++ b/app/mailers/concerns/email_delivery.rb
@@ -11,7 +11,7 @@ module EmailDelivery
     return unless member
     return unless @_mail_was_called
 
-    MemberEmailDelivery.find_or_create_by!(member: member, email_type: action_name) do |delivery|
+    MemberEmailDelivery.find_or_create_by!(member:, email_type: action_name) do |delivery|
       delivery.subject = mail.subject
       delivery.body = mail.html_part ? mail.html_part.body.to_s : mail.body.to_s
       delivery.to = Array(mail.to)
--- a/app/mailers/concerns/email_delivery.rb
+++ b/app/mailers/concerns/email_delivery.rb
@@ -3,18 +3,20 @@ module EmailDelivery
 
   private
 
+  # The mailer action name distinguishes which email a logged row is for.
+  # find_or_create_by keeps a replayed delivery (e.g. a retried job) from
+  # raising on the unique index after the email has already gone out.
   def log_sent_email
     member = params[:member]
     return unless member
     return unless @_mail_was_called
 
-    MemberEmailDelivery.create!(
-      member: member,
-      subject: mail.subject,
-      body: mail.html_part ? mail.html_part.body.to_s : mail.body.to_s,
-      to: Array(mail.to),
-      cc: Array(mail.cc),
-      bcc: Array(mail.bcc)
-    )
+    MemberEmailDelivery.find_or_create_by!(member: member, email_type: action_name) do |delivery|
+      delivery.subject = mail.subject
+      delivery.body = mail.html_part ? mail.html_part.body.to_s : mail.body.to_s
+      delivery.to = Array(mail.to)
+      delivery.cc = Array(mail.cc)
+      delivery.bcc = Array(mail.bcc)
+    end
   end
 end
--- a/spec/mailers/member_mailer_spec.rb
+++ b/spec/mailers/member_mailer_spec.rb
@@ -192,7 +192,7 @@ RSpec.describe MemberMailer do
     it 'logs one row per member even if the delivery is performed twice' do
       expect do
         described_class.with(member: member).chaser.deliver_now
-        described_class.with(member: member).chaser.deliver_now
+        described_class.with(member:).chaser.deliver_now
       end.to change(MemberEmailDelivery, :count).by(1)
     end
   end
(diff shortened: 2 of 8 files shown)


KimberleyCook (approved):

app/mailers/concerns/email_delivery.rb
```diff
@@ -3,18 +3,20 @@ module EmailDelivery
...
 
-    MemberEmailDelivery.create!(
-      member: member,
-      subject: mail.subject,
-      body: mail.html_part ? mail.html_part.body.to_s : mail.body.to_s,
-      to: Array(mail.to),
-      cc: Array(mail.cc),
-      bcc: Array(mail.bcc)
-    )
+    MemberEmailDelivery.find_or_create_by!(member: member, email_type: action_name) do |delivery|
```
olleolleolle:
```suggestion
    MemberEmailDelivery.find_or_create_by!(member:, email_type: action_name) do |delivery|
```

spec/mailers/member_mailer_spec.rb
```diff
@@ -180,12 +180,20 @@
...
       expect(log.to).to eq([member.email])
       # premailer-rails converts the message to multipart/alternative during
       # delivery, so the logged body must come from the html part
       expect(log.body).to be_present
       expect(log.body).to include('codebar workshop')
     end
+
+    it 'logs one row per member even if the delivery is performed twice' do
+      expect do
+        described_class.with(member: member).chaser.deliver_now
```
olleolleolle, line 194:
```suggestion
        described_class.with(member:).chaser.deliver_now
```

spec/mailers/member_mailer_spec.rb
```diff
@@ -180,12 +180,20 @@
...
+        described_class.with(member: member).chaser.deliver_now
```
olleolleolle:
```suggestion
        described_class.with(member:).chaser.deliver_now
```

spec/services/three_month_email_service_spec.rb
```diff
@@ -34,6 +34,20 @@
       member
     end
 
+    let!(:student_with_other_email_logged) do
+      member = Fabricate(:member)
+      Fabricate(:subscription, member: member, group: students_group)
```
olleolleolle:
```suggestion
      Fabricate(:subscription, member:, group: students_group)
```

spec/services/three_month_email_service_spec.rb
```diff
@@ -34,6 +34,20 @@
...
+      Fabricate(
+        :workshop_invitation,
+        member: member,
+        workshop: Fabricate(:workshop, chapter: chapter, date_and_time: 6.months.ago),
```
olleolleolle:
```suggestion
        workshop: Fabricate(:workshop, chapter:, date_and_time: 6.months.ago),
```

spec/services/three_month_email_service_spec.rb
```diff
@@ -34,6 +34,20 @@
...
+        role: 'Student',
+        attended: true
+      )
+      Fabricate(:member_email_delivery, member: member, email_type: 'welcome_email')
```
olleolleolle:
```suggestion
      Fabricate(:member_email_delivery, member:, email_type: 'welcome_email')
```

spec/services/three_month_email_service_spec.rb
```diff
@@ -185,6 +206,13 @@
         role: 'Student',
         attended: true
       )
+      Fabricate(
+        :workshop_invitation,
+        member: student_with_other_email_logged,
+        workshop: Fabricate(:workshop, chapter: chapter, date_and_time: 1.month.ago),
```
olleolleolle:
```suggestion
        workshop: Fabricate(:workshop, chapter:, date_and_time: 1.month.ago),
```

olleolleolle (approved):
Nice!

mroderick:
Thank you for your review @olleolleolle. I've created https://github.com/codebar/planner/issues/2833 to enforce the use of the shorthand style, so you don't need to suggest it on every review going forwards.

PR #2835, opened 2026-09-01
feat: signup nudge email sequence for members without a chapter

mroderick:
## Description

Implements #2384: automated emails to members who signed up via the website but haven't subscribed to a chapter.

A bounded two-stage sequence, recorded in `member_email_deliveries` (typed log from #2832):

1. **Signup nudge** — members created 7–30 days ago with no chapter subscription, not banned
2. **Follow-up** — members whose nudge row is at least a month old, still no subscription, not banned

Stage eligibility is derived from the typed log rows; delivery-confirmed writes (via the `EmailDelivery` concern, `email_type: signup_nudge` / `signup_nudge_followup`) mean a lost send retries within the stage-1 window and no member receives the same stage twice — enforced by the `(member_id, email_type)` unique index. Members who subscribe at any point exit the sequence.

**Email copy:** both emails currently use the copy from Kimberley's first-email draft (subject "Let's get you connected with codebar!"). Dedicated follow-up copy is pending from Kimberley (issue #2384) and will land as a small follow-up commit.

## What happens when (plain language)

Every day, one scheduled job looks at recent sign-ups and sends at most two emails — then never bothers anyone again:

- **A week after someone signs up**, they get a short nudge: pick a codebar chapter. Sent exactly once — the send is recorded in a log whose database index makes a second copy impossible, even if the job re-runs.
- **One month after that nudge**, if they still haven't subscribed to any chapter, they get one follow-up.
- **After that, silence.** Even a member who never subscribes gets no third email, ever.

People who are exempt at every stage:

- anyone who **subscribes to a chapter** (they leave the sequence immediately and get the normal chapter welcome email instead),
- anyone **banned** from codebar,
- anyone who signed up **more than a month before this ships** (the sequence only looks forward — nobody gets a backfilled backlog of old nudges).

Members who signed up but never finished the signup form are included — the nudge is for everyone who created an account and walked away.

Driving it: one daily task (`rake chaser:signup_nudges`) that must be added to the Heroku Scheduler after merge — see the ops note above.

## Ops note ⚠️

After merge, add a Heroku Scheduler entry: `rake chaser:signup_nudges`, daily — same mechanism as `rake chaser:three_months`. Without it the feature ships dark. Post-merge verification (first scheduled run creates `signup_nudge` rows for the current cohort): Morgan.

## Steps to verify

- `make test` — service spec covers the full stage matrix (window edges, subscribed/banned, both rows = terminal, follow-up anchored to the nudge's send time); mailer specs cover headers, body, and typed logging; rake spec covers the entry point.

### DIFF
--- a/app/services/signup_nudge_email_service.rb
+++ b/app/services/signup_nudge_email_service.rb
@@ -1,0 +1,42 @@
+class SignupNudgeEmailService
+  NUDGE = 'signup_nudge'.freeze
+  FOLLOWUP = 'signup_nudge_followup'.freeze
+
+  def self.send_nudges
+    send_signup_nudge
+    send_signup_nudge_followup
+  end
+
+  def self.send_signup_nudge
+    Member.not_banned
+          .where(created_at: nudge_window)
+          .merge(unemailed(NUDGE))
+          .merge(never_subscribed)
+          .find_each { |member| MemberMailer.with(member:).signup_nudge.deliver_later }
+  end
+
+  def self.send_signup_nudge_followup
+    Member.not_banned
+          .joins(:member_email_deliveries)
+          .where(member_email_deliveries: { email_type: NUDGE, created_at: ..1.month.ago })
+          .merge(unemailed(FOLLOWUP))
+          .merge(never_subscribed)
+          .distinct
+          .find_each { |member| MemberMailer.with(member:).signup_nudge_followup.deliver_later }
+  end
+
+  def self.unemailed(email_type)
+    Member.where.not(id: MemberEmailDelivery.where(email_type:).select(:member_id))
+  end
+
+  def self.never_subscribed
+    Member.where.not(id: Subscription.select(:member_id))
+  end
+
+  def self.nudge_window
+    30.days.ago.beginning_of_day..7.days.ago.end_of_day
+  end
+
+  private_class_method :send_signup_nudge, :send_signup_nudge_followup, :unemailed, :never_subscribed,
+                       :nudge_window
+end
--- a/app/services/signup_nudge_email_service.rb
+++ b/app/services/signup_nudge_email_service.rb
@@ -25,12 +25,13 @@ class SignupNudgeEmailService
           .find_each { |member| MemberMailer.with(member:).signup_nudge_followup.deliver_later }
   end
 
+  # NULL-safe NOT IN: a NULL member_id in the subquery would exclude every member
   def self.unemailed(email_type)
-    Member.where.not(id: MemberEmailDelivery.where(email_type:).select(:member_id))
+    Member.where.not(id: MemberEmailDelivery.where(email_type:).where.not(member_id: nil).select(:member_id))
   end
 
   def self.never_subscribed
-    Member.where.not(id: Subscription.select(:member_id))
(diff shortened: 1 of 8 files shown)


KimberleyCook:
This looks awesome!

Could we tweak the backdate to one month, rather than 2 weeks.

Here is the copy for that first reminder email https://docs.google.com/document/d/1R2DjZiSoBVMnfp0mBS4nz8rMKywlyhhmb9exuDYAUR0/edit?usp=sharing, I'll add to the document today for the second email.

Thaaaank you :)

mroderick:
@KimberleyCook, the date changes have been applied. I've used the same content for both e-mails for now, so we can ship this today. It's a 5 minute job to replace the content of the second email, once it's ready.

app/services/signup_nudge_email_service.rb
```diff
@@ -0,0 +1,42 @@
...
+          .distinct
+          .find_each { |member| MemberMailer.with(member:).signup_nudge_followup.deliver_later }
+  end
+
+  def self.unemailed(email_type)
+    Member.where.not(id: MemberEmailDelivery.where(email_type:).select(:member_id))
+  end
+
+  def self.never_subscribed
+    Member.where.not(id: Subscription.select(:member_id))
```
olleolleolle: If the Subscription table grows very long, hm, could this be a very long SQL query? Perhaps there are other ways to tell the Member model to exclude all subscribers? A subquery or something.

spec/lib/tasks/chaser_rake_spec.rb
```diff
@@ -0,0 +1,13 @@
+RSpec.describe 'rake chaser:signup_nudges', type: :task do
```
olleolleolle, line 3: require the rails_helper - at the top of the file.

spec/services/signup_nudge_email_service_spec.rb
```diff
@@ -0,0 +1,100 @@
+RSpec.describe SignupNudgeEmailService, type: :service do
```
olleolleolle, line 3: Require.

olleolleolle (approved):
Cool!

spec/lib/tasks/chaser_rake_spec.rb
mroderick: Done — added `require 'rails_helper'` at the top.

app/services/signup_nudge_email_service.rb
mroderick: The query was already a subquery — `where.not(id: Subscription.select(:member_id))` generates `NOT IN (SELECT member_id FROM subscriptions)`, an uncorrelated subquery that Postgres runs as a hash anti-join, so it stays linear even as the table grows.

Your comment did surface a real bug though: both `member_id` columns are nullable in the database, and a single NULL row would make `NOT IN` exclude every member — the daily job would send nothing. Fixed by filtering NULLs inside the subqueries, with a regression test.

spec/services/signup_nudge_email_service_spec.rb
mroderick: Done — added `require 'rails_helper'` at the top.

PR #2838, opened 2026-09-02
Skip CSRF protection for token-keyed invitation RSVP actions

mroderick:
Members who open invitation links from mail apps cannot RSVP: their browser withholds the session cookie, so CSRF verification fails with `ActionController::InvalidAuthenticityToken` (Rollbar [535](https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/535#detail)). The invitation token in the URL already authenticates these actions, so this change removes the redundant CSRF check from them.

Key changes:

- `skip_forgery_protection` for the six token-keyed RSVP actions: `InvitationsController#attend/#reject`, `WorkshopInvitationController#update/#accept`, and `WaitingListsController#create/#destroy` (scoped `only:`).
- Each action looks up the invitation by a 128-bit `SecureRandom` token carried in the URL. The request carries no session authority, which is the only thing CSRF protects — an attacker who cannot guess the token cannot forge the request. PR #2641 applied the same reasoning to the feedback form.
- Regression specs for all six actions: each posts without a CSRF token and asserts the mutation still happens.

<details>
<summary>Background and scope</summary>

Web browsers withhold cookies when they classify a navigation as cross-site. WebKit's Intelligent Tracking Prevention does this for every iOS browser, including Chrome, so members arriving from Mail, Gmail, or Slack reach the invitation page with an ephemeral session. The RSVP POST then arrives without a matching session and fails CSRF verification.

The failure first surfaced on the feedback form and was fixed for that form alone in PR #2641. An audit of all mutating routes found six more token-keyed actions with the same structure; everything else (self check-in, account settings, admin, OAuth) is session-gated and keeps CSRF protection.

Related, not addressed here: the workshop invitation "reject" link mutates via a GET request, which email prefetchers can trigger. That needs its own change.
</details>

### DIFF
--- a/app/controllers/invitations_controller.rb
+++ b/app/controllers/invitations_controller.rb
@@ -1,4 +1,10 @@
 class InvitationsController < ApplicationController
+  # The invitation token in the URL is the authenticator for these actions;
+  # CSRF is redundant and fails when browsers withhold the session cookie
+  # (e.g. Safari/WebKit ITP on cross-site navigation). Same rationale as
+  # FeedbackController#submit (PR #2641, Rollbar #535).
+  skip_forgery_protection only: %i[attend reject]
+
   before_action :require_login, only: [:index]
   before_action :set_invitation, only: %i[show attend reject]
 
--- a/app/controllers/waiting_lists_controller.rb
+++ b/app/controllers/waiting_lists_controller.rb
@@ -1,6 +1,12 @@
 class WaitingListsController < ApplicationController
   include WorkshopInvitationConcerns
 
+  # The invitation token in the URL is the authenticator for these actions;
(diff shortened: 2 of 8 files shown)


olleolleolle (approved):
Perhaps some of the test preconditions could become a nested context.

mroderick:
Applied: the CSRF-protection toggle now lives in a shared context (`spec/support/shared_contexts/forgery_protection.rb`) used by all seven specs, and each precondition moved into nested `context` blocks with `let`/`before`. The `#update` and `#accept` no-CSRF specs are now nested inside their existing action describes, reusing the file's top-level `let(:invitation)`. The feedback spec from #2641 uses the shared context too.

PR #2843, opened 2026-09-03
Upgrade image_processing to 2.0.3 with explicit mini_magick backend

mroderick:
Upgrading image_processing to 2.0.3 breaks this app on its own: 2.0 made the mini_magick and ruby-vips backends soft dependencies, so dependabot's re-resolution (#2840) dropped both from the lockfile, Active Storage's boot path failed without ruby-vips, and the first avatar or image upload would fail without mini_magick. This PR takes the upgrade with the two changes image_processing 2.0's migration requires, scoped so the vips backend and its native libvips system dependency are never needed.

Key changes:

- `gem "mini_magick"` added as a direct dependency — `CarrierWave::MiniMagick` (`avatar_uploader.rb`, `image_uploader.rb`) processes every avatar and event image upload through it. Depending on it transitively is what let the lockfile drop it.
- `config.active_storage.variant_processor = :mini_magick` in `config/application.rb` — Active Storage variants are unused by this app, so the transformer keeps the same backend CarrierWave already uses and the vips backend (which needs a native libvips install in CI and production) is never loaded.
- `ruby-vips` is therefore not needed anywhere and stays out of the lockfile.

<details>
<summary>Why this upgrade is worth taking</summary>

image_processing 2.0.1-2.0.3 fix several remote-code-execution paths (shell execution through loader/saver options, operation names from user input). The app never calls ImageProcessing directly and uses no Active Storage variants, so the vulnerable surface is not reachable today; the upgrade keeps the gem current and closes those paths if variant processing is ever adopted. Replacing dependabot PR #2840, whose lockfile dropped the backends and failed to boot.
</details>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -29,7 +29,12 @@ gem 'jquery-rails'
 gem 'jwt'
 
 # Use Active Storage variant
+# image_processing 2.0 made its backends soft dependencies; mini_magick is
+# required by CarrierWave::MiniMagick (avatar/image uploads). Active Storage's
+# variant processor is set to :mini_magick in config/application.rb so the
+# vips backend is never loaded.
 gem 'image_processing'
+gem 'mini_magick'
 gem 'mutex_m' # LOCKED: Added because of activesupport 7.0
 gem 'nokogiri'
 gem 'omniauth'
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Very clear, thanks!

PR #2844, opened 2026-09-05
Chapter health card on the admin chapter page

mroderick:
I'm trying to think of ways to make it easier for admins and organisers alike to have better insights into the health of chapters. I think that workshop frequency and number of active organisers are some good indicators, so I've tried to capture this in a health card that is displayed at the top of the chapter admin page.

---

Adds a Chapter Health card to the admin chapter page so admins and chapter organisers can see a chapter's health at a glance, without visiting the separate Chapter Status page.

- Status badge (active / dormant / inactive) using the same classification rules as the Chapter Status page
- Previous workshop (days ago) and next workshop (days away), with exact dates on hover
- Median workshop cadence over the past 180 days
- Eligible members (subscribed, not banned, TOC accepted) and organiser count
- Workshop and events timeline: SVG, 180 days past / 90 days future, today marked; workshops as blue circles, single-chapter events as amber squares; same-day events stack with an xN count. Pure SVG plus native tooltips, no JavaScript

image

<details>
<summary>Design and review notes</summary>

- The card is visible to admins and the chapter's organisers. Chapter show was already organiser-accessible, and the data shown is the chapter's own, so no new access is granted.
- Single-chapter events only: multi-chapter events list 33-44 chapters each, so they carry no per-chapter signal (production data: 13 events in the last 180 days cover 36 chapters). New `Event.single_chapter` scope.
- Classification logic mirrors `Admin::ChaptersController#status` inline rules; the two implementations coexist by design (status page stays unchanged).
- Timeline date basis is midnight-aligned so tick labels and markers cannot drift at DST boundaries.
- Reviews applied: layered-rails (scope extraction, view arithmetic moved to the value object, spec coverage restored), ponytail-review (dead code, inlined single-caller module), ce-code-review (test gap on the new scope).
- Residual cosmetic edge: more than three same-day events stack only three markers; the xN label still shows the true count.

</details>

PR #2846, opened 2026-09-08
Enable RSpecRails/InferredSpecType and drop redundant spec types

mroderick:
Enables the RuboCop cop that flags spec type metadata RSpec already infers from the file path, so redundant `type: :model`-style tags never need manual review again.

- Enable RSpecRails/InferredSpecType (it ships pending) and register `components: component` in its Inferences, since ViewComponent specs are outside the rspec-rails defaults
- Autocorrect the 56 offenses it found across 56 spec files (removal of the redundant `type:` pair from `RSpec.describe`)

Details:

<details><summary>Why this is safe here</summary>

The cop assumes `config.infer_spec_type_from_file_location!` is enabled, which this repo does in `spec/rails_helper.rb` - including the custom `spec/components` mapping registered before inference runs, and `config.include ViewComponent::TestHelpers, type: :component`, which keys off the inferred type. Removing the explicit tag therefore changes nothing at runtime. The full suite passes (1308 examples, 0 failures) and RuboCop is clean with the cop enabled.
</details>

Motivation: reviewer feedback on #2844 pointed out redundant `type: :component` metadata; this makes the linter catch it permanently.

### DIFF
--- a/spec/controllers/admin/chapters_controller_spec.rb
+++ b/spec/controllers/admin/chapters_controller_spec.rb
@@ -1,4 +1,4 @@
-RSpec.describe Admin::ChaptersController, type: :controller do
+RSpec.describe Admin::ChaptersController do
   let(:admin) { Fabricate(:member) }
 
   before do
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -21,6 +21,14 @@ plugins:
   - rubocop-rspec
   - rubocop-rspec_rails
 
+# spec/components is registered as the :component type in rails_helper.rb, so
+# an explicit `type: :component` is redundant. Inferences covers the standard
+# rspec-rails directories; ViewComponent's directory is not one of them.
+RSpecRails/InferredSpecType:
+  Enabled: true
+  Inferences:
+    components: component
+
 Layout/LineLength:
   Max: 120
   Exclude:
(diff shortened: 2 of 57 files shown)


spec/controllers/admin/chapters_controller_spec.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.describe Admin::ChaptersController, type: :controller do
+RSpec.describe Admin::ChaptersController do
```
olleolleolle, line 1: Perhaps all files should include the require line?

mroderick: I'm not sure I follow. Could you elaborate?

olleolleolle: `require ‘rails_helper’` this is the line I referred to.

Why? When executing 1 single spec, we want to be sure that the helper code has been loaded, when running the rest of the test.

mroderick: Got it, thanks!

I've created https://github.com/codebar/planner/issues/2858 to track that, as it is orthogonal to the work in this PR :)

olleolleolle (approved):
🚀

PR #2850, opened 2026-09-08
Record member session activity in the activities log

mroderick:
## Summary

First of a stacked series that records member, organiser, and admin activity in the existing public_activity `activities` table. This PR adds the recorder — the single funnel every later PR calls — and instruments the session lifecycle: `member.login` via a shared `sign_in!` funnel in `AuthServicesController#create` (both OAuth branches), `member.logout` in `ApplicationController#logout!`, and `auth_service.removed` on provider unlink. Design: `docs/organiser-activity` side branch (spec: 2026-09-07-organiser-activity-strip-design).

## Key changes

- `MemberActivityRecorder.record(actor:, key:, trackable:, recipient:)` — one row per action, owner = acting member; persistence failures are logged, never raised, so recording cannot break a user flow.
- `sign_in!` refactor: the two session-setting branches in `AuthServicesController#create` funnel through one private method; session behaviour unchanged (existing OAuth callback specs pass).
- The provider-unlink action is instrumented but stays unreachable — its route addition was deliberately reverted; enabling it is a separate product decision.

## Review notes

Focus first on the `sign_in!` refactor (session behaviour must be identical) and the recorder's `trackable ||= actor` fallback (public_activity validates trackable presence; self-referential rows are intentional). Deliberately not done: no backfill — history starts at deploy.

Stack: PR 2 adds subscription/profile activity -> PR 3 RSVPs -> PR 4 admin actions -> PR 5 workshops/batches -> PR 6 the admin activity strip.

---

Context and motivation: #2857

### DIFF
--- a/spec/requests/member_login_activity_spec.rb
+++ b/spec/requests/member_login_activity_spec.rb
@@ -1,4 +1,3 @@
-# spec/requests/member_login_activity_spec.rb
 require 'rails_helper'
 
 RSpec.describe 'Member login activity' do
--- a/spec/requests/member_login_activity_spec.rb
+++ b/spec/requests/member_login_activity_spec.rb
@@ -1,0 +1,21 @@
+# spec/requests/member_login_activity_spec.rb
+require 'rails_helper'
+
+RSpec.describe 'Member login activity' do
+  it 'records member.login on the existing auth service path' do
+    member = Fabricate(:member, email: 'existing-login@example.com')
+    Fabricate(:auth_service, member:, provider: 'github', uid: 'login-uid-1')
+
+    mock_auth_hash(provider: 'github', uid: 'login-uid-1', email: member.email)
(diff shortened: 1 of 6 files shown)


spec/requests/member_login_activity_spec.rb
```diff
@@ -0,0 +1,21 @@
+# spec/requests/member_login_activity_spec.rb
```
olleolleolle:
```suggestion
```

spec/requests/member_logout_activity_spec.rb
```diff
@@ -0,0 +1,15 @@
+# spec/requests/member_logout_activity_spec.rb
```
olleolleolle:
```suggestion
```

olleolleolle (approved):
Good to have.

PR #2866, opened 2026-09-10
feat(admin): organisation-wide monthly stats page for admins

### LINKED ISSUE #2860: Stats page for super admins
In the admin area under the "Chapter Status" button I'd like a new page where I can check out a bunch of codebar stats. Including : 

- Attendance (both students and coaches)
- New signs up (both students and coaches).
- Workshop number (breakdown by year like the below link)

The main thing with these stats is I would like to be able to filter between a data range. Super useful for me when reporting grant data.

v1 is this page https://stats.codebar.io/

mroderick:
Super admins can now pull codebar-wide monthly numbers for any month range on an admin page and export the same figures as CSV. The page exists for grant reporting, where the current stats surface can only break numbers down per year.

image

Fixes #2860

- One captioned table per metric group: attendance (check-ins and RSVPs, students and coaches), new sign-ups by role with a distinct-member total, and workshop counts — monthly columns plus totals for the selected range.
- 3/6/12-month presets in complete calendar months (default: the most recent 3) or a custom start/end month range with inline validation; Download CSV always mirrors the displayed table.
- Visible only to global admins: the admin area also admits chapter organisers, so the stats controller checks admin status itself for both the page and the CSV.
- Figures are computed on request from existing tables; no new storage, jobs, or caching.

## Review notes

- The invalid-range split in `Admin::StatsController#index`: the page shows an inline error and keeps the last valid range; the CSV endpoint instead falls back to the default range.
- Sign-up role assignment in `Admin::Stats::Monthly`: a member in both a Students and a Coaches group appears in both role rows but once in the total, so role rows may sum above it.
- Deliberately not done: caching or precomputation (not needed at current scale), per-chapter breakdowns, and events/meetings — the page covers workshops only.

<details>
<summary>Background and scope</summary>

- The page complements stats.codebar.io, which stays unchanged; feedback ratings remain there.
- Sign-up rows count every member ever created, with no ban or terms filter — sign-ups are historical facts, so counts stay stable when a member is later banned or accepts the terms. Role rows use current group membership.
- Check-in figures exist only where organisers recorded check-ins (the flow launched in March 2025), so older periods under-report them; the RSVP rows give a complete basis for the same period.
- Custom ranges are capped server-side at 240 months; wider requests reuse the invalid-range path. Very wide tables scroll horizontally.
- The CSV writes raw integers (no thousands separators) so it pastes cleanly into spreadsheets.

</details>

## Post-deploy monitoring

- Watch `/admin/stats` response times in Scout APM; the page runs three grouped aggregate queries per render with no caching.
- Rollbar errors mentioning `Admin::Stats` or `CookieOverflow` are the rollback signal; revert the branch.
- Validation window: the first week after deploy.

### DIFF
--- a/app/components/admin/stats/table_component.html.erb
+++ b/app/components/admin/stats/table_component.html.erb
@@ -1,0 +1,25 @@
+<div class="table-responsive mb-4">
+  <table class="table table-sm table-hover align-middle">
+    <caption class="caption-top text-start"><%= caption %></caption>
+    <thead>
+      <tr>
+        <th scope="col">Metric</th>
+        <% months.each do |month| %>
+          <th scope="col" class="text-end"><%= month_header(month) %></th>
+        <% end %>
+        <th scope="col" class="text-end">Totals</th>
+      </tr>
+    </thead>
+    <tbody>
+      <% rows.each do |row| %>
+        <tr>
+          <th scope="row"><%= row.label %></th>
+          <% months.each do |month| %>
+            <td class="text-end"><%= cell_value(row, month) %></td>
+          <% end %>
+          <td class="text-end"><%= total_value(row) %></td>
+        </tr>
+      <% end %>
+    </tbody>
+  </table>
+</div>
--- a/app/components/admin/stats/table_component.rb
+++ b/app/components/admin/stats/table_component.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+module Admin
+  module Stats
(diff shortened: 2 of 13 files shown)


KimberleyCook:
This is very very exciting 👀

KimberleyCook (approved):
I'm probably not the best person to review the code itself but from what I see it looks good!

And the UI and feature itself is so awesome! This is going to help me so so much!! **THANK YOU!!!**

PR #2855, opened 2026-09-08
Add activity strip and activity log to the admin member profile

mroderick:
## Summary

Last of the activity-log stack (base: #2854). Adds a GitHub-status-style weekly activity strip (52 ISO weeks, three states) plus a recent activity log on the admin member profile, rendered in the wide column. Both render for every member; not member-facing.

Fake data from local development shows what this could look like for someone very active.

image

## Key changes

- `Admin::Members::ActivityStrip` service: one indexed query over `activities` owned by the member, bucketed into 52 ISO weeks; states `:empty` / `:login_only` (logins and logouts only) / `:active` (any other key); per-week key counts for tooltips.
- `Admin::Members::ActivityStripComponent`: 52 tall SVG bars, codebar blue for active weeks, with SVG `<title>` tooltips (ISO week + counts, e.g. "2026-W37 (week of 7 Sep 2026): 2 member login") and `aria-label` per bar; no JavaScript. Uses SVG titles rather than HTML `title` attributes so tooltips work in Firefox.
- Activity log: latest 30 activities for the viewed member, grouped under ISO week headings matching the strip's format, timestamps in `Sunday, 2026-09-13 11:58` form. Each row links to the related entity where one exists — e.g. "Event invitation rsvp for <event>" — with chapter-qualified group names ("Peterborough Coaches"); member-subject activities (logins, profile updates) render without links.
- Data fidelity: the log starts at instrumentation deploy (PR 1 of this stack) — earlier weeks read as empty by design.

## Review notes

Week math and state classification live solely in the service (component renders only). Verified: full suite 1365/0, RuboCop clean.

---

Context and motivation: #2857

PR #2879, opened 2026-09-15
Add RuboCop cop enforcing Data.define over Struct.new

### LINKED ISSUE #2878: Add RuboCop cop enforcing Data.define over Struct.new and convert existing usages
Olle's review on #2855 asked for `Data.define` instead of `Struct.new` in `ActivityStrip`. RuboCop has no built-in cop for this preference, so we should add one and clean up existing usages.

## Work

1. Add a custom cop `Style/PreferDataDefine` (in `lib/rubocop/cop/style/prefer_data_define.rb`, required from `.rubocop.yml`). It flags `Struct.new` sends. No autocorrection: the rewrite isn't safely mechanical (mutability and keyword semantics differ), so each hit gets a human decision.

```ruby
module RuboCop
  module Cop
    module Style
      class PreferDataDefine < Base
        MSG = 'Prefer `Data.define` over `Struct.new` for immutable value objects.'

        def on_send(node)
          return unless node.receiver&.const_type? && node.receiver.value == :Struct
          return unless node.method?(:new)

          add_offense(node)
        end
      end
    end
  end
end
```

2. Convert the existing offenses: the 7 `Struct.new(:id, :email).new(...)` test doubles in `spec/helpers/email_header_helper_spec.rb`. They are the only usages in the repo, and `Data.define(:id, :email).new(1, '...')` works with the same positional arguments, so the conversion is mechanical there.

3. Add cop tests (`spec/rubocop/...` or inline) covering the flag case and a pass case for `Data.define`.

## Notes

- After adding the cop, regenerate `.rubocop_todo.yml` if needed and confirm CI passes.
- Alternative rejected: a grep-based CI check — less precise and bypasses the existing RuboCop todo workflow.

Context: https://github.com/codebar/planner/pull/2855#discussion_r4013378571

mroderick:
Closes #2878.

Adds a custom RuboCop cop, `Style/PreferDataDefine`, that flags `Struct.new` sends with no autocorrection. The seven `Struct.new` test doubles in `spec/helpers/email_header_helper_spec.rb` are now RSpec verifying doubles (`instance_double(Member, ...)`), per review feedback — the repo's only `Struct.new` usages are gone. Follows Olle's review request on #2855.

The cop is required from `.rubocop.yml`, so `bundle exec rubocop` and CI enforce the preference from now on. A review round also fixed Zeitwerk eager-loading of `lib/rubocop` under CI. No autocorrection on purpose: the `Struct.new` -> `Data.define` rewrite isn't safely mechanical (mutability and keyword semantics differ), so each future hit gets a human decision.

## Review notes

- One deviation from the issue's snippet: `ConstNode#value` no longer exists in rubocop-ast 1.50, the cop uses `const_name == 'Struct'` instead (verified for both `Struct` and `::Struct`; namespaced receivers like `Other::Struct.new` are correctly ignored).
- Known blind spots, accepted per the issue's intent: `AllCops.Exclude` (db/config/bin) is not covered; a future `--regenerate-todo` while offenses exist would add a todo entry and suppress the cop; name-based indirection (`s = Struct; s.new`) bypasses it.


spec/helpers/email_header_helper_spec.rb
```diff
@@ -6,7 +6,7 @@
   end
 
   describe '#mail_to_member' do
-    let(:member) { Struct.new(:id, :email).new(1, 'test@example.com') }
+    let(:member) { Data.define(:id, :email).new(1, 'test@example.com') }
```
olleolleolle, line 9: In this test and others, I suggest `double(:member, id: 1, email: 'test@example.com')`, that is, RSpec doubles.

olleolleolle (approved):

spec/helpers/email_header_helper_spec.rb
mroderick: Done — switched all seven to `instance_double(Member, ...)` (a verifying double: checks against the real Member class and satisfies RSpec/VerifiedDoubles). The helper only reads `member.id` and `member.email`, so the double covers it. > In this test and others, I suggest `double(:member, id: 1, email: 'test@example.com')`, that is, RSpec doubles.

PR #2881, opened 2026-09-16
Stop MIME scanner probes from creating Rollbar noise

### LINKED ISSUE #2873: Malformed Accept headers from bots cause 500 errors (Rollbar #549)
## Description of the issue 📄

Rollbar shows recurring `ActionDispatch::Http::MimeNegotiation::InvalidType` errors, e.g.:
`"../../../../../../../../../../etc/services{{" is not a valid MIME type`
([Rollbar #549](https://app.rollbar.com/a/codebar-production/fix/item/codebar-production/549))

These are automated scanner probes that send path-traversal strings in the `Accept` header. Rails 8.1 raises `InvalidType` during MIME negotiation, and the class is not in the default `rescue_responses` map (verified in actionpack 8.1.3.1's `exception_wrapper.rb`), so each probe becomes a 500 and a Rollbar notification. There is nothing for the application to do with these requests — they should never reach it, or at worst return 4xx.

## Options

**1. Nginx edge block** — `config/nginx.conf.erb` gets a `map` on the Accept/Content-Type headers and returns 406 when either contains `../`:

```nginx
map "$http_accept $content_type" $bad_mime {
    default 0;
    ~"\.\./" 1;
}

if ($bad_mime) { return 406; }
```

Pros: stops these requests before Rails, no app code. Cons: nginx config is baked into the dyno, so it's awkward to test locally; the rule only covers `../` patterns in those two headers.

**2. App-level rescue** — map the exception to 406 in `config/application.rb`:

```ruby
config.action_dispatch.rescue_responses[
  "ActionDispatch::Http::MimeNegotiation::InvalidType"
] = :not_acceptable
```

Pros: one line, testable, uses the same mechanism Rails already uses for `ActionController::UnknownFormat`. Cons: the request still reaches the app (and its access logs), though Rollbar goes quiet.

**3. Both** — nginx rule as the primary block, `rescue_responses` as the safety net for anything nginx misses (other probe shapes hitting the same code path).

## Recommendation

Option 3. The nginx rule blocks the current probe pattern cheaply, and the rescue mapping guarantees no 500 for this class of malformed request regardless of what pattern the bots use next.

Comments on the approach welcome — happy to open a PR once agreed.

mroderick:
Fixes #2873

## Corrected diagnosis

The issue description assumed these probes cause 500s and that `rescue_responses` needs the mapping added. Neither holds on Rails 8.1.3.1:

- `ActionDispatch::Http::MimeNegotiation::InvalidType` has been mapped to `:not_acceptable` in the default `rescue_responses` since Rails 6.1 (rails/rails commit 2bad7eb1, 2020). Malformed Accept/Content-Type headers already get 406 responses.
- Rollbar still reported the exception because its `RollbarMiddleware` reports every exception passing through it (`:use_exception_level_filters => true`), even ones Rails has already rescued and rendered as 406.

<details>
<summary>Verification of the middleware chain</summary>

`config/initializers/rollbar.rb` already ignores `ActionController::RoutingError` via `config.exception_level_filters` for the same reason. Rollbar's `ExceptionReporter#exception_data` passes `use_exception_level_filters: true`, and `'ignore'` values suppress reporting entirely (rollbar 3.8.0, `lib/rollbar/notifier.rb`).
</details>

## Changes

1. **Rollbar ignore filter** (`config/initializers/rollbar.rb`) — add `ActionDispatch::Http::MimeNegotiation::InvalidType` to `exception_level_filters` as `'ignore'`. This stops the Rollbar noise regardless of which header shape the bots use.
2. **Nginx edge block** (`config/nginx.conf.erb`) — a `map` on the combined Accept/Content-Type headers returns 406 when either contains `../`, keeping these probes out of the application's access logs.

## Testing

- New `spec/config/rollbar_config_spec.rb` asserts the filter is registered (written first, failed with `nil` before the initializer change).
- New `spec/requests/malformed_mime_headers_spec.rb` characterises the existing 406 behaviour for both headers.
- Full suite: `bundle exec parallel_rspec spec/ -n 3` — 1446 examples, 0 failures. RuboCop clean on changed files.
- The nginx config cannot be exercised in CI (nginx is baked into the dyno image). After deploy, verify on staging with:
  `curl -s -o /dev/null -w '%{http_code}' -H 'Accept: ../../etc/services{{' https://<staging-host>/` should return 406.


KimberleyCook (approved):

olleolleolle (approved):
Neat!

Issue #2883, opened 2026-09-16
SponsorsController#index still slow in production after #2803 caching (median 1.5s, render-bound)


mroderick:
## Summary

Requests to `/sponsors` are chronically slow despite the fragment caching and table indexes merged in #2803. Measured from the canonical Rails logs (`rails_semantic_logger` `process_action` payload, collected via the production log drain) over a 3-hour window on 2026-09-16/17.

## Measured (3h window, 123 requests)

| Metric | Value |
|---|---|
| median duration | 1,474ms |
| p90 duration | 2,891ms |
| max duration | 21.3s |
| median view_runtime | 1,337ms |
| median db_runtime | 41ms |
| queries per request | 5 |
| avg allocations | 2.5M per request (17.7M on the slowest) |

The page is render-bound: 41ms of DB work across 5 queries, but 1.3s of view time. Fragment caching from #2803 does not appear to be effective in production.

## Context

#2803 (merged 2026-08-31) split the sponsors partial for collection caching and added indexes. Local profiling at the time showed warm-cache partial allocations dropping ~80% (225k → 44k objects). Production medians are unchanged a month later.

## Hypotheses

1. Fragment cache misses in production — cache store may be unset or per-dyno memory store, flushed by restarts
2. Cache key churn (sponsor `updated_at` touches) defeating reuse
3. Render cost outside the cached fragments (page shell, other partials)

## Next steps

- Confirm the production cache store and measure the fragment hit rate
- Reproduce with a production-sized sponsor list and profile allocations again
- Decide on a caching strategy (longer-lived fragments, low-level caching, or both)

mroderick:
Verified against the production canonical log archive (2026-09-16 → 09-23): the numbers in this issue no longer exist in production.

The slow period ended at 09-17 10:10 CEST with deploys v762/v763 (Rollbar locals-capture disable #2896, nginx scanner-junk edge block #2884) — p50 dropped from ~1.6s to ~60–90ms, and fragment cache hits returned (allocations 1.8M → 77k), consistent with scanner-junk traffic thrashing Solid Cache rather than a wiped store. PR #2897 (deployed as v766 on 09-22) then added the whole-page body cache: current p50 is 6–9ms, p90 12ms, view_runtime ~0ms, 2 queries, ~660 allocations, no 5xx since.

Closing as resolved by #2897 (which also closed the follow-up diagnosis in #2885). If the page ever regresses, the v1 cache key in `SponsorsController#index` is the invalidation lever.

PR #2897, opened 2026-09-17
Cache the rendered /sponsors page body

### LINKED ISSUE #2885: Cache the rendered /sponsors page body — recurring 1.3s render stalls dyno threads
## Summary

`/sponsors` is the public sponsors page: `SponsorsController#index` loads every active sponsor, groups them by level, and the view renders a logo grid per level from `shared/sponsors` → a per-sponsor partial with `cached: true`. It re-renders 721 sponsor fragments on **every** request, even when those fragments are cache hits. That costs ~1.3s of CPU-bound Ruby per render (~123 requests in a 3h window). On production dynos (puma, 1 worker × 3 threads) a CPU-bound render holds the GVL and stalls unrelated requests on the same dyno. The slowest renders reach **17–21s**.

Proposal: cache the rendered page body in Solid Cache, keyed on the latest `sponsors.updated_at`, so a warm request costs one cache read.

## Measured

From production canonical logs (3h window, 123 requests) and a local reproduction against a production dump:

| Metric | Value |
|---|---|
| median view_runtime (warm, fragments all hits) | 1,337ms |
| allocations per render | ~2.5M (17.7M worst) |
| db_runtime / queries | 41ms / 5 queries |
| cold render after dyno restart | 17.8s local repro, 21.3s in production |
| page size | ~210KB HTML, 721 logos |

Local reproduction: cold 17.8s, warm ~1.1s — matching production's max and median.

## Why the per-sponsor fragments don't help

#2803 fragment-cached each sponsor partial (`cached: true`) and removed the CarrierWave allocation hot spot. The fragments do live in Solid Cache in production (`config.cache_store = :solid_cache_store`, which Rails also uses for collection caching — not the default per-process MemoryStore as hypothesised in #2883), so restarts don't wipe them. But the per-request cost is still dominated by the collection machinery: building 721 fragment keys, reading 721 entries back from the DB-backed cache store, and assembling a 210KB page — all per render, scaling with sponsor count.

The 17–21s cold renders are not explained by a wiped in-process store. The likely mechanism is Solid Cache eviction or expiry forcing a full re-render plus 721 fragment writes on the first request after a deploy, but that needs verification before we lean on it.

## Proposed fix

Cache the whole rendered body, keyed on the most recent sponsor change:

```ruby
def index
  key = "sponsors/index/#{Sponsor.active.maximum(:updated_at)&.to_fs(:usec)}"
  body = Rails.cache.fetch(key) do
    @sponsor_levels = Sponsor.active.group_by(&:level)
    render_to_string
  end
  render html: body.html_safe
end
```

Properties:

- Warm requests cost one `maximum` query plus one Solid Cache read — the 721-fragment read and the sponsor load only run on a miss
- Solid Cache is DB-backed, so the body survives dyno restarts and deploys
- Any sponsor edit bumps `updated_at` → new key → one re-render, then warm again
- Stale window: none (key changes the moment a sponsor is saved)

## Expected impact

Median `/sponsors` view drops from ~1,337ms to ~10–20ms (one cache read; a full re-render only after a sponsor edit), and the recurring GVL stall disappears from both dynos.

## Related

- #2883 — the original diagnosis; its "per-dyno memory store" hypothesis is corrected above
- #2803 — the merged fragment caching + indexes this builds on

mroderick:
## Summary

/sponsors re-rendered all 721 sponsor fragments on every request (~1.3s CPU-bound median, ~2.5M allocations), even when every fragment was a cache hit. The page body is now cached in the shared cache store, keyed on the latest active sponsor's updated_at. A warm request costs one `maximum` query plus one cache read; the sponsor load and render only run on a miss.

Correction to the original diagnosis (#2885): the per-sponsor fragments already lived in Solid Cache and survived deploys and restarts, so a wiped store was never what caused the 17-21s cold renders. The likely mechanism is eviction or expiry in Solid Cache forcing a full rebuild. This change doesn't make those rebuilds cheaper — a miss still costs the full ~17-21s render — it makes them rare: one long-lived body entry instead of 721 fragments under constant eviction pressure.

| Metric | Before | After |
|---|---|---|
| Median view_runtime (warm) | ~1,337ms | one cache read (~10-20ms) |
| Cache miss (sponsor edit or eviction) | full 721-fragment build | same, but rare — one entry instead of 721 |
| Invalidation | none | key changes the moment any sponsor is saved |

## Design decisions

- The key is `sponsors/index/v1/{Sponsor.active.maximum(:updated_at)&.to_fs(:usec)}`. The manually bumped v1 segment lets a reviewer invalidate the body when the view or partials change; without it, a deploy that changes the template keeps serving the cached body until the next sponsor save.
- The key is computed before the records load, so a save committing between the two queries caches a body newer than its key, which self-heals on the next request. The reverse order could pin a stale body under the new key permanently.
- The sponsor load sits inside the cache block, so warm requests instantiate no sponsor records.
- The layout still renders per request (`render_to_string(layout: false)`, then `render html:`), so CSRF meta tags and navigation stay fresh.
- Writes that bypass `updated_at` (`update_columns`, `update_all`) leave the key unchanged and the page stale. No in-app caller does this today; a request spec documents the accepted gap.

## Test plan

- New request spec (`spec/requests/sponsors_spec.rb`) covers rendering, the cache hit (body unchanged when no sponsor's updated_at changed) and the re-render after a sponsor save.
- Full suite: 1,449 examples, 0 failures. RuboCop clean.

- Scout APM: the `/sponsors` endpoint's mean and 95th-percentile view runtime should drop from ~1.3s median to tens of milliseconds once the key is warm. Warm requests issue 1 query.
- Across releases, `/sponsors` view_runtime should stay low. An eviction or expiry of the body entry still costs a full build on the next request; expect that only after a sponsor edit.
- Failure signal: `/sponsors` serving stale content for more than a few minutes after a sponsor edit would mean the key is not changing; investigate `Sponsor.active.maximum(:updated_at)` and Solid Cache health. Rollback is a revert; the previous uncached behaviour is restored.
- Owner: whoever reviews this PR.

Fixes #2885
Related: #2803, #2883


mroderick:
The numbers are no longer accurate, since we merged https://github.com/codebar/planner/pull/2896 ... but the idea is still the same.

olleolleolle (approved):

PR #2905, opened 2026-09-18
feat: close RSVPs by default 3.5 hours before start

### LINKED ISSUE #2901: Automatic RSVP close time: default workshops to close 3.5 hours before start
Organisers expect RSVPs to close automatically around 3pm on the day of a workshop. The planner has never done this — the expectation goes unmet, and late RSVPs are routine. Discussed in [chapter-organiser-meetings#111 (comment)](https://github.com/codebar/chapter-organiser-meetings/issues/111#issuecomment-5705005942).

## Background

- Since October 2014, members cannot *cancel* later than 3.5 hours before a workshop starts ([391089f](https://github.com/codebar/planner/commit/391089fd8a3b92d2491d42a9a4328b3cf4780005), message corrected in [fc605b4](https://github.com/codebar/planner/commit/fc605b48709e697bb0af9b40183de12f3f7d9b89)). 3.5 hours before an 18:30 workshop is 15:00, which is likely where the 3pm expectation comes from.
- A `rsvp_close_time` field has existed since February 2015 ([fd059c6](https://github.com/codebar/planner/commit/fd059c6098582581643afa11a7f508eaaa203691)) but was never enforced by default. #238 and #1237 were closed without implementation.
- #2687 was fixed by [#2691](https://github.com/codebar/planner/pull/2691) with a manual, opt-in close-time field on the workshop form. It works, but it is per-workshop and easy to skip.

## Data (production dump, 16 Sep 2026)

- Of the 206 workshops held in the past 12 months, 5 had a close time set (3 London, 2 Nottingham). The manual habit is not sticking.
- 58 student RSVPs were made on the workshop day after 15:00 (74 including waiting-list promotions). All but 2 were on workshops with no close time.

## Proposal

Give every workshop, multi-chapter event, and meeting an effective close time of **3.5 hours before its start time** when no explicit close time is set. This blocks new RSVPs only — waiting-list promotions continue as today. Organisers can still set a per-workshop value through the existing field, which overrides the default.

## Form behaviour

When an organiser sets the start time on the event form, the RSVP close time field should populate with the default (start time minus 3.5 hours) so it's visible before save. This is client-side, vanilla JavaScript — no jQuery.

Changing the start time must not overwrite an RSVP close time the organiser has already set or that was saved with the event; only an empty field gets the default.

mroderick:
## Summary

Workshops, multi-chapter events, and meetings now close to new RSVPs 3.5 hours before their start time unless an organiser sets an explicit close time. The workshop form pre-fills the close date and time fields with that default so organisers see it before saving.

Closes #2901

### Behaviour

- New RSVPs (workshop accept, event attend, meeting RSVP) are rejected once the effective close time passes: the controller redirects with a notice, and the event and meeting pages show a closed badge.
- An explicit `rsvp_closes_at` on a workshop overrides the default.
- Waiting-list promotions and cancellations are unchanged. Cancellation keeps its existing deadline (the earlier of start minus 3.5 hours and any explicit close time).
- Retroactive on deploy: events and meetings have no stored close time, so upcoming ones start closing 3.5 hours before their start as soon as this ships. Members holding RSVP links inside that window will see the closed notice.

### Form pre-fill

`rsvp-close-time.js` (vanilla JS, loaded via the Sprockets manifest): changing the start date or time on the workshop form fills empty close fields with start minus 3.5 hours, and refreshes values the script itself wrote earlier. Values the organiser typed or that were saved with the workshop are never overwritten.

## Unapplied review findings

From the focused code review (run `20260918-122423-c88d7a19`, verdict: Ready to merge):

- [ ] P2 — app/controllers/workshop_invitation_controller.rb:74 — cancellation deadline re-implements the 3.5-hour rule with `min()` semantics instead of using `RsvpClosable::DEFAULT_RSVPS_CLOSE_OFFSET`. Left for maintainers: cancellation timing is a separate, deliberate attendance policy (cancellations were already capped at 3.5 hours before start since 2014); switching it to the override semantics would extend the cancellation window whenever an explicit close time is later, which is a product decision. Findings about the stale JS prefill and the DST caveat were applied in bf969384.

- Expected behaviour change: RSVP attempts after the effective close time return the "RSVPs have now closed" notice. Production data from the issue (58 student RSVPs on workshop days after 15:00 in 12 months) should drop to near zero.
- Watch: Rollbar for unexpected exceptions (as opposed to the notice) in `invitations#attend` and `invitations#rsvp_meeting`; organiser reports of surprise closures; waiting-list promotion volume, which should be unchanged.
- Validation window: the first two weeks of workshops after deploy.
- Rollback: reverting restores the previous open-until-start behaviour for workshops without an explicit close time.

## Testing

- `make test` (parallel RSpec): 1465 examples, 0 failures.
- `rubocop` clean on all changed files.
- New coverage: `RsvpClosable` shared example (default close time, boundary offsets, post-start) included by the event, meeting, and workshop specs; workshop explicit-close-time override; controller specs for the closed paths of `InvitationsController#attend` and `#rsvp_meeting`.


olleolleolle (approved):

olleolleolle:
Curious: the JS used `var` rather than let and const. Is that universal in this app?

mroderick:
> Curious: the JS used `var` rather than let and const. Is that universal in this app?

Yes. The majority of JS in the app is jQuery era.

PR #2910, opened 2026-09-22
fix(auth): stop error callbacks crashing with NoMethodError in Rack middleware

mroderick:
Provider error redirects from the auth app (invalid scope, PKCE required, state mismatch) reached the planner as a 500 (`NoMethodError: undefined method '[]=' for nil`, Rollbar item 714) instead of the intended redirect to `/auth/failure`.

**Change**

- Return the failure response from `fail!` in `callback_phase`'s error branch instead of discarding it with a bare `return`. The discarded response left the OmniAuth middleware returning nil, which Rack's `TempfileReaper` then crashed on.
- In the same rescue path, re-raise exceptions raised by the app after successful auth (they are not auth failures) and return `on_failure`'s response for auth failures.
- Add regression specs covering both levels: the strategy must return a Rack response for provider error callbacks, and the full middleware stack must redirect to `/auth/failure` rather than raise.

**Review notes**

- The behaviour change to note: exceptions raised by the controller during a successful callback previously got masked as "authentication error" redirects; they now propagate as normal 500s.
- The redirect target and message stay unchanged (`/auth/failure` → flash + root); only the crash path is fixed.

**CI failure follow-up**

The first CI run failed one of the new specs. Cause: Capybara's Puma test server sets `ENV['RACK_ENV'] ||= 'development'` when a feature spec boots it mid-process, and omniauth's `FailureEndpoint` raises (instead of redirecting) whenever `RACK_ENV` is `development`. Specs running after a feature spec therefore saw raise-out behaviour. Fixed by pinning `OmniAuth.config.failure_raise_out_environments = []` in the test environment, so failure handling is deterministic regardless of spec order.


olleolleolle (approved):
Looks like a good idea! The specs read well.

